PR #151, opened 2016-12-20
[quantizer] replace the regex with an iterative grouping

palazzem:
### What it does

Groups similar escaped patterns like ``( ?, ?, ?, ? )`` and ``( ? ), ( ? ), ( ? )`` in a single ``( ? )`` without using regex. This improves the performance and the reliability of the quantizer, when SQL queries are incomplete and a regex cannot handle the case.

#### Benchmarks
```
BenchmarkTokenizer/Escaping-4              	  100000	     13835 ns/op	    4400 B/op	      87 allocs/op
BenchmarkTokenizer/Grouping-4              	  200000	      7440 ns/op	    2720 B/op	      68 allocs/op

= Previous optimization =

BenchmarkTokenizer/Escaping-4              	   50000	     34285 ns/op	    4768 B/op	      90 allocs/op
BenchmarkTokenizer/Grouping-4              	  100000	     19095 ns/op	    2944 B/op	      71 allocs/op
```


quantizer/consumer.go
```diff
@@ -3,16 +3,16 @@ package quantizer
 import (
```
LotharSee: Let's put some extra comments at the top of `consumer.go` and `tokenizer.go` to explain what they are for (seeing them at the `quantizer` level, it isn't clear that they are in fact related to sql-only processing).

palazzem: Sure! Will do in this PR!

quantizer/consumer.go
```diff
@@ -96,20 +152,19 @@ func (t *TokenConsumer) Process(in string) (string, error) {
...
-	//
-	//   [...] VALUES ( ?, ?, ? ) -> VALUES ?
-	//
-	// This part should be replaced later to improve a lot the performance of this TokenConsumer
-	// but for the moment is a good compromise.
-	result := r.ReplaceAllString(out.String(), " ? ")
-	result = strings.TrimSpace(result)
-
+	// remove whitespaces at the begin / end of the string
+	result := strings.TrimSpace(out.String())
```
clutchski: nitpicky: but can't be done in the parsing loop?

quantizer/consumer.go
clutchski: we might want to move this to a `quantizer/sql.go` package too.

palazzem: +1 moving right now

quantizer/consumer.go
palazzem: nice catch! that was related to the previous regex but with the new ``GroupingFilter`` this is not required.

quantizer/consumer.go
palazzem: merged ``consumer.go`` with ``sql.go`` and added comments on top of the ``tokenizer.go`` file

quantizer/consumer.go
```diff
@@ -1,119 +0,0 @@
-package quantizer
```
palazzem: merged with ``sql.go``

quantizer/consumer_test.go
```diff
@@ -1,72 +0,0 @@
-package quantizer
```
palazzem: merged with ``sql_test.go``

quantizer/sql.go
```diff
@@ -1,24 +1,194 @@
...
+				out.WriteRune(' ')
+			}
+
+			out.Write(buff)
+		}
+
+		t.lastToken = token
+	}
+
+	return out.String(), nil
```
palazzem: no extra steps are required

PR #155, opened 2016-12-26
yank defers in hot paths

LeoCavaille:


LeoCavaille:
Slightly faster, and in general we just need to get in the habit of keeping defers away for high throughput stuff. Follows @clutchski ' s remark of perf boost by removing those.

PR #156, opened 2016-12-26
introduce a new resource syntax for pipelined redis commands

galdor:
When there are more than one command, the syntax is now:

        PIPELINE [ <commands> ]

Where <commands> is a deduplicated and sorted list of commands.

This new format will significantly reduce the cardinality of the
resource set for application which use a lot of pipelined commands.

### DIFF
--- a/quantizer/redis.go
+++ b/quantizer/redis.go
@@ -16,11 +16,21 @@ var redisCompoundCommandSet = map[string]bool{
 func QuantizeRedis(span model.Span) model.Span {
 	query := compactWhitespaces(span.Resource)
 
-	rawLines := strings.Split(query, "\n")
-	lines := make([]string, 0, len(rawLines))
-	for _, rawLine := range rawLines {
+	lines := []string{}
+	for len(query) > 0 {
+		var rawLine string
+
+		idx := strings.IndexByte(query, '\n')
+		if idx == -1 {
+			rawLine = query
+			query = ""
+		} else {
+			rawLine = query[:idx]
+			query = query[idx+1:]
+		}
+
 		if line := strings.Trim(rawLine, " "); len(line) > 0 {
-			lines = append(lines, line)
+			lines = append(lines, string(line))
 		}
 	}
 
--- a/quantizer/redis.go
+++ b/quantizer/redis.go
@@ -1,6 +1,8 @@
 package quantizer
 
 import (
+	"bytes"
+	"sort"
 	"strings"
 
 	"github.com/DataDog/datadog-trace-agent/model"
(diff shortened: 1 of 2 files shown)


quantizer/redis.go
```diff
@@ -12,38 +13,59 @@ var redisCompoundCommandSet = map[string]bool{
...
+	case 0:
+		span.Resource = ""
+
+	case 1:
+		// Single command
+		span.Resource = readLine(lines[0])
+
+	default:
+		// Pipeline
+		commandMap := make(map[string]bool)
```
LeoCavaille: You can save a few bytes by using a `map[string]struct{}` to implement a set.

quantizer/redis.go
```diff
@@ -12,38 +13,59 @@ var redisCompoundCommandSet = map[string]bool{
...
+			commands[i] = command
+			i++
+		}
+		sort.Strings(commands)
+
+		span.Resource = "PIPELINE ["
+		for _, command := range commands {
+			span.Resource += " " + command
 		}
+		span.Resource += " ]"
```
LeoCavaille: Rather than joining strings by using the `+` operator, you should use a `bytes.Buffer` to save allocations (strings are immutable in go) and use `.String()` only once.

quantizer/redis.go
```diff
@@ -12,38 +13,59 @@ var redisCompoundCommandSet = map[string]bool{
...
+		for _, line := range lines {
+			commandMap[readLine(line)] = true
+		}
+
+		commands := make([]string, len(commandMap))
+		i := 0
+		for command := range commandMap {
+			commands[i] = command
+			i++
+		}
```
LeoCavaille: nit but slightly more idiomatic:
```go
cmds := make([]string, 0, len(commandMap))
for cmd := range commandMap {
    cmds = append(cmds, cmd)
}

File: quantizer/redis.go [hunk 4]
```diff
@@ -12,38 +13,59 @@ var redisCompoundCommandSet = map[string]bool{
...
+	readLine := func(line string) string {
+		args := strings.SplitN(line, " ", 3)
+		command := strings.ToUpper(args[0])
+
+		if redisCompoundCommandSet[command] {
+			command += " " + strings.ToUpper(args[1])
+		}
+
+		return command
+	}
```
LeoCavaille: I know this was like that before, but looking at that code, I'm slightly worried about the perfs with all the string manipulations.
Can you add a go benchmark to the test suite and try to verify that this doesn't introduce a perf regression.

quantizer/redis_test.go
```diff
@@ -37,13 +37,13 @@ func TestRedisQuantizer(t *testing.T) {
...
 		{"SET toto tata \n \n  EXPIRE toto 15  ",
-			"SET EXPIRE"},
+			"PIPELINE [ EXPIRE SET ]"},
 
 		{"MSET toto tata toto tata toto tata \n ",
 			"MSET"},
 
 		{"MULTI\nSET k1 v1\nSET k2 v2\nSET k3 v3\nSET k4 v4\nDEL to_del\nEXEC",
-			"MULTI SET* DEL EXEC"},
+			"PIPELINE [ DEL EXEC MULTI SET ]"},
```
LeoCavaille: Can you add a few more examples of things that are currently badly quantized in prod and get some more coverage in this test case?

galdor:
I applied the modifications you asked for. Test coverage is 100%. There is now a benchmark function based on a long command pipeline.

quantizer/redis.go
```diff
@@ -12,38 +14,58 @@ var redisCompoundCommandSet = map[string]bool{
...
+
+	resource := bytes.NewBuffer([]byte{})
```
LeoCavaille: nitpick but you can just do
```go
var resources bytes.Buffer
```

quantizer/redis.go
```diff
@@ -12,38 +14,58 @@ var redisCompoundCommandSet = map[string]bool{
...
+		}
+		sort.Strings(commands)
+
+		resource.WriteString("PIPELINE [")
+		for _, command := range commands {
+			resource.WriteByte(' ')
+			resource.WriteString(command)
 		}
+		resource.WriteString(" ]")
+		span.Resource += " ]"
```
LeoCavaille: I think you want to remove this last line?

galdor: Indeed

quantizer/redis.go
galdor: done

PR #159, opened 2016-12-27
[benchmark] add cpuprofiler flag to save a profiling session

palazzem:
### What it does

* add `-cpuprofile=cpu.pprof` flag
* add `-memprofile=mem.pprof` flag

It's useful to profile the full flow overhead when the agent is constantly hit by a huge number of traces per second.

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -7,6 +7,7 @@ import (
 	"math/rand"
 	"os"
 	"os/signal"
+	"runtime"
 	"runtime/pprof"
 	"syscall"
 	"time"
@@ -83,6 +84,7 @@ func main() {
 
 	// profiling arguments
 	cpuprofile := flag.String("cpuprofile", "", "write cpu profile to file")
+	memprofile := flag.String("memprofile", "", "write memory profile to `file`")
 	flag.Parse()
 
 	// start CPU profiling
@@ -141,4 +143,19 @@ func main() {
 	go handleSignal(agent.exit)
 
 	agent.Run()
+
+	// collect memory profile
+	if *memprofile != "" {
+		f, err := os.Create(*memprofile)
+		if err != nil {
+			log.Critical("could not create memory profile: ", err)
+		}
+
+		// get up-to-date statistics
+		runtime.GC()
+		if err := pprof.WriteHeapProfile(f); err != nil {
+			log.Critical("could not write memory profile: ", err)
+		}
+		f.Close()
+	}
 }
--- a/agent/main.go
+++ b/agent/main.go
@@ -5,15 +5,17 @@ import (
 	"flag"
 	"fmt"
 	"math/rand"
-	_ "net/http/pprof" // register debugger
 	"os"
 	"os/signal"
+	"runtime/pprof"
 	"syscall"
 	"time"
 
 	"github.com/DataDog/datadog-trace-agent/config"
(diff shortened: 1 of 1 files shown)


agent/main.go
```diff
@@ -72,13 +75,29 @@ func versionString() string {
...
+	// command-line arguments
 	flag.StringVar(&opts.ddConfigFile, "ddconfig", "/etc/dd-agent/datadog.conf", "Classic agent config file location")
 	// FIXME: merge all APM configuration into dd-agent/datadog.conf and deprecate the below flag
 	flag.StringVar(&opts.configFile, "config", "/etc/datadog/trace-agent.ini", "Trace agent ini config file.")
 	flag.BoolVar(&opts.debug, "debug", false, "Turn on debug mode")
 	flag.BoolVar(&opts.version, "version", false, "Show version information and exit")
+
+	// profiling arguments
+	cpuprofile := flag.String("cpuprofile", "", "write cpu profile to file")
+	memprofile := flag.String("memprofile", "", "write memory profile to `file`")
```
LeoCavaille: Not sure why these are handled differently, they are command-line arguments too?
Also I wonder if we should go this way or prefer signaling to profile a live process. Like you send SIGUSR1 to start and stop a profile (we have that for some of our python stuff).

agent/main.go
```diff
@@ -5,15 +5,18 @@ import (
...
+	"runtime"
+	"runtime/pprof"
 	"syscall"
 	"time"
 
 	"github.com/DataDog/datadog-trace-agent/config"
 	"github.com/DataDog/datadog-trace-agent/statsd"
 	log "github.com/cihub/seelog"
+
+	_ "net/http/pprof"
```
LeoCavaille: Do we need this? I think there is already a pprof server running in the receiver at localhost:7777/debug/pprof/ ?

agent/main.go
LeoCavaille: Actually scratch that, we can already get that via pprof. Can't we do long running profiles with it already?

agent/main.go
palazzem: yes that one can be removed

agent/main.go
palazzem: I've used these flags only to have a single entrypoint to:
* start the agent with ``-cpuprofile``
* start the ``trace-flood`` command that hits the agent a huge number of times
* wait for some seconds (profiling is activated)
* stop everything
* read the profile file

I can remove everything, but in this way (I think) it's easy to script and to run locally. To profile a running execution we can just use the pprof HTTP.

agent/main.go
palazzem: Wait.... we don't have a ``_ "net/http/pprof"`` in the receiver. If I remove that line:
```
$ curl localhost:7777/debug/pprof/
404 page not found
```

agent/main.go
palazzem: I'm currently using ``forego`` with the following ``Procfile``:
```
agent: ./trace-agent -cpuprofile cpu.prof
flood: ./trace-flood
interrupt: sleep 60; exit 1
```

In this way I could run the profiler for X seconds and see profiling results.

palazzem:
This PR has been merged with #160 in a new branch: https://github.com/DataDog/datadog-trace-agent/compare/master...palazzem-ufoot/profiling-branch

Closing since it will not land on ``master`` for the moment.

PR #160, opened 2016-12-27
[tools] add cmd package for external profiling tools

palazzem:
### What it does

* add the ``cmd`` package for external profiling tools
* ``rake build_tools`` builds the ``trace-flood`` binary
* ``trace-flood`` will start flooding a real ``trace-agent`` executed in ``localhost``

#### Notes

``trace-flood`` uses the ``fixtures`` package. Actually it doesn't create a good random combination of traces but it's a separated task that should only impact how we build fake traces.

### DIFF
--- a/cmd/traceflood/traceflood.go
+++ b/cmd/traceflood/traceflood.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"bytes"
+	"flag"
+	"log"
+	"math/rand"
+	"net/http"
+	"time"
+
+	"github.com/DataDog/datadog-trace-agent/fixtures"
+	"github.com/DataDog/datadog-trace-agent/model"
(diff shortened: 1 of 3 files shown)


cmd/traceflood/traceflood.go
```diff
@@ -0,0 +1,61 @@
+package main
```
palazzem: Not using ``dd-trace-go`` because in this way the tool doesn't depend on external projects. Furthermore, in this way it's easier to reuse the ``fixtures`` package.

palazzem:
This PR has been merged with #159 in a new branch: https://github.com/DataDog/datadog-trace-agent/compare/master...palazzem-ufoot/profiling-branch

Closing since it will not land on ``master`` for the moment.

PR #161, opened 2016-12-27
ignore truncated redis commands during quantization

galdor:


galdor:
There is still something wrong, and I need this patch in another branch anyway.

PR #163, opened 2016-12-28
[profiling] minor speed-up on assembleGrain

ufoot:
Slightly faster assembleGrain implementation, less generic, hardcodes env, resource, service

### DIFF
--- a/model/stats.go
+++ b/model/stats.go
@@ -3,6 +3,7 @@ package model
 import (
 	"bytes"
 	"fmt"
+	"sort"
 
 	"github.com/DataDog/datadog-trace-agent/quantile"
 )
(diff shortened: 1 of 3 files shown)


model/stats.go
```diff
@@ -133,25 +134,45 @@ func NewStatsBucket(ts, d int64) StatsBucket {
...
 }
 
-func assembleGrain(b *bytes.Buffer, keys, vals []string) (string, TagSet) {
-	if len(keys) != len(vals) {
-		panic("assembleGrain diff lengths!")
+func assembleGrain(b *bytes.Buffer, env, resource, service string, m map[string]string) (string, TagSet) {
+	b.Reset()
+
+	b.WriteString("env")
+	b.WriteRune(':')
```
palazzem: we may compact this right?

PR #165, opened 2016-12-28
[profile] dump traces and services input

ufoot:
Added `-dumptraces` and `-dumpservices` to dump all incoming input.

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -34,15 +34,13 @@ func handleSignal(exit chan struct{}) {
 
 // opts are the command-line options
 var opts struct {
-	ddConfigFile   string
-	configFile     string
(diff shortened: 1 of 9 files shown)


agent/main.go
```diff
@@ -34,11 +34,15 @@ func handleSignal(exit chan struct{}) {
...
-	logLevel     string
-	version      bool
+	ddConfigFile   string
+	configFile     string
+	debug          bool
+	logLevel       string
+	version        bool
+	dumpTraces     string
+	dumpServices   string
+	replayTraces   string
```
palazzem: we can use a separated command `trace-replay` to replay flushes

PR #167, opened 2016-12-29
[decoder] use typed models for msgpack decoding

palazzem:
### What it does

Uses ``codecgen`` to generate ``typed_msgpack.go`` file that contains all implementations for named types. Involved models are:
* ``Span``
* ``Trace``
* ``ServicesMetadata``

It adds a new command to re-generate the ``typed_msgpack.go`` file:
```
rake generator:update
```

#### Performance impact

```
benchmark                     old ns/op     new ns/op     delta
BenchmarkDecoderMsgpack-4     36573120      26245336      -28.24%

benchmark                     old allocs     new allocs     delta
BenchmarkDecoderMsgpack-4     129007         128857         -0.12%

benchmark                     old bytes     new bytes     delta
BenchmarkDecoderMsgpack-4     9832704       9827904       -0.05%
```


palazzem:
OK @LeoCavaille, we can delay this PR until we've tried it for a lot of time with a successful fuzzy, real data testing.

clutchski:
worth a try.

clutchski:
though it feels like protobuf should handle this case just as well no?

palazzem:
Trying another PR since we can get a better performance with the same approach.

PR #168, opened 2016-12-29
handle errors when sending payloads to endpoints

galdor:
The goal of these commits is to avoid losing payloads when the API is unavailable. A longer description is available in the last commit.

### DIFF
--- a/agent/endpoint.go
+++ b/agent/endpoint.go
@@ -102,10 +102,10 @@ func (a APIEndpoint) Write(p model.AgentPayload) (int, error) {
 		url := a.urls[i] + model.AgentPayloadAPIPath()
 		req, err := http.NewRequest("POST", url, bytes.NewBuffer(data))
 		if err != nil {
-			log.Errorf("could not create request for endpoint %s: %v", url, err)
 			// If the request cannot be created, there is no point
 			// in trying again later, it will always yield the
 			// same result.
+			log.Errorf("could not create request for endpoint %s: %v", url, err)
 			continue
 		}
 
@@ -132,20 +132,14 @@ func (a APIEndpoint) Write(p model.AgentPayload) (int, error) {
 			if resp.StatusCode/100 == 5 {
 				endpointErr.Append(a.urls[i], a.apiKeys[i], err)
 			}
+
 			continue
 		}
 
 		flushTime := time.Since(startFlush)
 		log.Infof("flushed payload to the API, time:%s, size:%d", flushTime, len(data))
-		truncKey := a.apiKeys[i]
-		if len(truncKey) > 5 {
-			truncKey = truncKey[0:5]
-		}
-		tags := []string{
-			fmt.Sprintf("url:%s", a.urls[i]),
-			fmt.Sprintf("apikey:%s", truncKey),
-		}
-		statsd.Client.Gauge("trace_agent.writer.flush_duration", flushTime.Seconds(), tags, 1)
+		statsd.Client.Gauge("trace_agent.writer.flush_duration",
+			flushTime.Seconds(), nil, 1)
 	}
 
 	if endpointErr.IsEmpty() {
--- a/agent/endpoint.go
+++ b/agent/endpoint.go
@@ -11,12 +11,48 @@ import (
 	log "github.com/cihub/seelog"
 )
 
+// apiError stores a list of errors triggered when sending data to a
+// list of endpoints. The endpoints member contains an api key and url for
+// each error.
+type apiError struct {
+	errs     []error // the errors, one for each endpoint
+	endpoint APIEndpoint
+}
+
+func newAPIError() *apiError {
+	return &apiError{}
+}
+
+func (err *apiError) IsEmpty() bool {
+	return len(err.errs) == 0
+}
+
+func (err *apiError) Append(url, apiKey string, e error) {
+	err.errs = append(err.errs, e)
+	err.endpoint.urls = append(err.endpoint.urls, url)
+	err.endpoint.apiKeys = append(err.endpoint.apiKeys, apiKey)
+}
+
+func (err *apiError) Error() string {
+	var buf bytes.Buffer
+
+	for i, e := range err.errs {
+		if i > 0 {
+			buf.WriteString(", ")
+		}
+
+		fmt.Fprintf(&buf, "%s: %v", err.endpoint.urls[i], e)
+	}
+
+	return buf.String()
+}
+
 // AgentEndpoint is an interface where we write the data
 // that comes out of the agent
 type AgentEndpoint interface {
 	// Write sends an agent payload which carries all the
 	// pre-processed stats/traces
-	Write(b model.AgentPayload)
+	Write(b model.AgentPayload) (int, error)
 	// WriteServices sends updates about the services metadata
 	WriteServices(s model.ServicesMetadata)
 }
@@ -49,17 +85,16 @@ func NewAPIEndpoint(urls, apiKeys []string) APIEndpoint {
 }
 
 // Write writes the bucket to the API collector endpoint.
-// Currently, the errors are just logged and we fail silently, keeping
-// writing till we have tried everything. This is because currently the
-// choice is to just drop the data we cannot write on the floor.
-// FIXME?
-func (a APIEndpoint) Write(p model.AgentPayload) {
+func (a APIEndpoint) Write(p model.AgentPayload) (int, error) {
 	data, err := model.EncodeAgentPayload(p)
 	if err != nil {
 		log.Errorf("encoding issue: %v", err)
-		return
+		return 0, err
 	}
-	statsd.Client.Count("trace_agent.writer.payload_bytes", int64(len(data)), nil, 1)
+	payloadSize := len(data)
+	statsd.Client.Count("trace_agent.writer.payload_bytes", int64(payloadSize), nil, 1)
+
+	endpointErr := newAPIError()
 
 	for i := range a.urls {
 		startFlush := time.Now()
@@ -68,6 +103,9 @@ func (a APIEndpoint) Write(p model.AgentPayload) {
 		req, err := http.NewRequest("POST", url, bytes.NewBuffer(data))
 		if err != nil {
 			log.Errorf("could not create request for endpoint %s: %v", url, err)
+			// If the request cannot be created, there is no point
+			// in trying again later, it will always yield the
+			// same result.
 			continue
 		}
 
@@ -79,12 +117,21 @@ func (a APIEndpoint) Write(p model.AgentPayload) {
 		resp, err := http.DefaultClient.Do(req)
 		if err != nil {
 			log.Errorf("error when requesting to endpoint %s: %v", url, err)
+			endpointErr.Append(a.urls[i], a.apiKeys[i], err)
 			continue
 		}
 		defer resp.Body.Close()
 
 		if resp.StatusCode/100 != 2 {
-			log.Errorf("request to %s responded with %s", url, resp.Status)
+			err := fmt.Errorf("request to %s responded with %s", url, resp.Status)
+			log.Error(err)
+
+			// Only retry for 5xx (server) errors; for 4xx errors,
+			// something is wrong with the request and there is
+			// usually no point in trying again.
+			if resp.StatusCode/100 == 5 {
+				endpointErr.Append(a.urls[i], a.apiKeys[i], err)
+			}
 			continue
 		}
 
@@ -100,6 +147,13 @@ func (a APIEndpoint) Write(p model.AgentPayload) {
 		}
 		statsd.Client.Gauge("trace_agent.writer.flush_duration", flushTime.Seconds(), tags, 1)
 	}
+
+	if endpointErr.IsEmpty() {
+		// The payload was sent to all endpoints without any error
+		return payloadSize, nil
+	}
+
+	return payloadSize, endpointErr
 }
 
 // WriteServices writes services to the services endpoint
@@ -146,8 +200,9 @@ func (a APIEndpoint) WriteServices(s model.ServicesMetadata) {
 type NullEndpoint struct{}
 
 // Write just logs and bails
-func (ne NullEndpoint) Write(p model.AgentPayload) {
+func (ne NullEndpoint) Write(p model.AgentPayload) (int, error) {
 	log.Debug("null endpoint: dropping payload, %d traces, %d stats buckets", p.Traces, p.Stats)
+	return 0, nil
 }
 
 // WriteServices just logs and stops
--- a/agent/trace-agent.ini
+++ b/agent/trace-agent.ini
@@ -27,6 +27,9 @@ api_key=apikey_2
 # default to true, disable if you want dry-run mode
 # enabled=false
 
+# the maximum size of the payload buffer in bytes
+# buffering is disabled if this setting is set to 0
+payload_buffer_max_size=16777216
 
 ###################################################
 # Agent concentrator - stats aggregation
--- a/agent/writer.go
+++ b/agent/writer.go
@@ -93,12 +93,17 @@ func (w *Writer) isPayloadBufferingEnabled() bool {
 	return w.conf.APIPayloadBufferMaxSize > 0
 }
 
-// Run starts the writer and starts writing what comes through the
-// input channel.
-// NOTE: this currently happens sequentially, but it would not be too
-// hard to mutex and parallelize. Not sure it's needed though.
+// Run starts the writer.
 func (w *Writer) Run() {
 	w.exitWG.Add(1)
+	go w.main()
+}
+
+// main is the main loop of the writer goroutine. If buffers payloads and
+// services read from input chans and flushes them when necessary.
+// NOTE: this currently happens sequentially, but it would not be too
+// hard to mutex and parallelize. Not sure it's needed though.
+func (w *Writer) main() {
 	defer w.exitWG.Done()
 
 	flushTicker := time.NewTicker(time.Second)
--- a/agent/writer.go
+++ b/agent/writer.go
@@ -2,12 +2,43 @@ package main
 
 import (
 	"sync"
+	"time"
 
 	log "github.com/cihub/seelog"
 
 	"github.com/DataDog/datadog-trace-agent/config"
 	"github.com/DataDog/datadog-trace-agent/model"
 )
+
+// the amount of time in seconds to wait before resending a payload
+const payloadResendDelay = 5 * time.Second
+
+// the amount of time in seconds a payload can stay buffered before being dropped
+const payloadMaxAge = 10 * time.Minute
+
+// writerPayload wraps a model.AgentPayload and keeps track of a list of
+// endpoints the payload must be sent to.
+type writerPayload struct {
+	payload      model.AgentPayload // the payload itself
+	size         int                // the size of the serialized payload or 0 if it has not been serialized yet
+	endpoint     AgentEndpoint      // the endpoints the payload must be sent to
+	creationDate time.Time          // the creation date of the payload
+	nextFlush    time.Time          // The earliest moment we can flush
+}
+
+func newWriterPayload(p model.AgentPayload, endpoint AgentEndpoint) *writerPayload {
+	return &writerPayload{
+		payload:      p,
+		endpoint:     endpoint,
+		creationDate: time.Now(),
+	}
+}
+
+func (p *writerPayload) write() error {
+	size, err := p.endpoint.Write(p.payload)
+	p.size = size
+	return err
+}
 
 // Writer is the last chain of trace-agent which takes the
 // pre-processed data from channels and tentatively output them
@@ -19,11 +50,13 @@ type Writer struct {
 	inPayloads chan model.AgentPayload     // main payloads for processed traces/stats
 	inServices chan model.ServicesMetadata // secondary services metadata
 
-	payloadBuffer []model.AgentPayload   // buffered of payloads ready to send
+	payloadBuffer []*writerPayload       // buffer of payloads ready to send
 	serviceBuffer model.ServicesMetadata // services are merged into this map continuously
 
 	exit   chan struct{}
 	exitWG *sync.WaitGroup
+
+	conf *config.AgentConfig
 }
 
 // NewWriter returns a new Writer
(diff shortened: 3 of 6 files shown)


agent/writer.go
```diff
@@ -86,14 +124,51 @@ func (w *Writer) FlushServices() {
...
+		if w.conf.APIResendPayloadsOnFailure {
+			if p.nextFlush.After(now) {
+				// We already tried to flush recently, so
+				// there's no point in trying again right now.
+				payloads = append(payloads, p)
+				continue
+			}
+		}
+
+		if err := p.endpoint.Write(p.payload); err != nil {
```
ufoot: AFAIK Write already counts `trace_agent.writer.payload_bytes` but I think we'd be interested by the number of:

- successful writes
- failed writes (and possibly distinguish with a tag the cases `dropped`, `retry_later`, `too_late`, `buffer_full`, ...)

This could be useful to make control dashboards.

agent/writer_test.go
```diff
@@ -87,29 +106,11 @@ func TestWriterPayload(t *testing.T) {
...
-			body:      string(body),
-		}
-		w.WriteHeader(200)
-	}))
+	testAPI := newTestServer(t, data)
+	defer testAPI.Close()
 
 	// buggy server
-	testAPI2 := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.WriteHeader(400)
```
ufoot: I think the old test was returning 400 but the new one is returning 500. And if I read the code well, the client checks that error code is 5xx. I'm not sure how we should handle 400, maybe not retry them (after all, if auth is bad or your client lib is not working, retrying later is pointless). In any case, I think it's worth testing both of them, a 500 internal and a 401 `StatusUnauthorized` for instance. This way there's no doubt, reading the test, we know the behavior.

agent/writer_test.go
```diff
@@ -143,8 +138,75 @@ receivingLoop:
...
+	w.Stop()
+
+	// Since the writer was created with a buffer of size 2, the first and
+	// oldest payload (p0) was discarded.
+	assert.Equal(2, len(w.payloadBuffer))
+	assert.Equal("p1", w.payloadBuffer[0].payload.Env)
+	assert.Equal("p2", w.payloadBuffer[1].payload.Env)
+}
+
+func TestWriterDisabledBufferization(t *testing.T) {
```
ufoot: 👍  thanks for testing both modes

GLOCKFILE
```diff
@@ -5,4 +5,4 @@ golang.org/x/tools 344080534465c5613c3563c30fd38c031073841d
 github.com/golang/lint 32a87160691b3c96046c0c678fe57c5bef761456
 github.com/stretchr/testify 8ce79b9f0b77745113f82c17d0756771456ccbd3
 github.com/ugorji/go faddd6128c66c4708f45fdc007f575f75e592a3c
-gopkg.in/ini.v1 12f418cc7edc5a618a51407b7ac1f1f512139df3
+gopkg.in/ini.v1 v1.11.0
```
galdor: Yep, as indicated in the commit message, it is an update to get support for boolean values. I prefer tags for readability, but if SHA1 are the rules here, no problem.

agent/writer.go
galdor: The previous code did not udpate these stats, but I could make a new patch to do it.

agent/writer_test.go
galdor: As the comment in the code says, we do not retry 4xx errors because they indicate that something is wrong with the request. I could add a test for it of course.

agent/writer.go
```diff
@@ -86,14 +124,51 @@ func (w *Writer) FlushServices() {
...
+				// again later, but only with the endpoints
+				// that failed.
+				p.endpoint = terr.endpoint
+
+				payloads = append(payloads, p)
+			}
+		}
+	}
+
+	if len(payloads) > w.conf.APIPayloadBufferMaxLength {
```
galdor: Actually we drop at the end because we want to always drop the oldest payloads (i.e. the beginning of the slice). If we drop in the loop when the slice is full, we will always drop the newest payloads.

GLOCKFILE
LeoCavaille: @galdor let's keep it as a SHA and be consistent for now.
At some point I think we just want to yank glock from the equation anyway. Likely use another tool like glide or just vendor everything to make things much easier.

agent/endpoint.go
```diff
@@ -11,12 +11,48 @@ import (
...
+}
+
+func (err *apiEndpointError) Append(url, apiKey string, e error) {
+	err.errs = append(err.errs, e)
+	err.endpoint.urls = append(err.endpoint.urls, url)
+	err.endpoint.apiKeys = append(err.endpoint.apiKeys, apiKey)
+}
+
+func (err *apiEndpointError) Error() string {
+	buf := bytes.NewBuffer([]byte{})
```
LeoCavaille: nit, default constructor with `var buf bytes.Buffer`

agent/writer.go
```diff
@@ -56,13 +81,21 @@ func NewWriter(conf *config.AgentConfig) *Writer {
 // NOTE: this currently happens sequentially, but it would not be too
 // hard to mutex and parallelize. Not sure it's needed though.
 func (w *Writer) Run() {
+	w.exitWG.Add(1)
+	defer w.exitWG.Done()
+
+	flushTicker := time.NewTicker(time.Second)
```
LeoCavaille: nit but you can defer `Stop` on the ticker to deallocate

agent/trace-agent.ini
```diff
@@ -27,6 +27,18 @@ api_key=apikey_2
...
+
+# the maximum number of payloads that can be bufferized
+payload_buffer_max_length=180
+
+# the minimum amount of time in seconds to wait before resending a payload
+# this delay is doubled after every failure until it reaches the maximum delay
+payload_resend_delay_min=5
+
+# the maximum amount of time in seconds to wait before resending a payload
+payload_resend_delay_max=160
```
LeoCavaille: I think there are a lot of parameters in here and it's maybe slightly overwhelming to a user.
Also buffer is a verb so ~~bufferize~~ :)

Maybe the boolean is not necessary as you'd just put the buffer size to 0 if you want too. For the delays I don't think any user would like to change that so let's just hardcode them in code. Likely 5s is good for min and max 5 to 15 minutes maybe?

agent/writer.go
LeoCavaille: Yes! Please modify the instrumentation so we can track this new behavior properly :+1:

LeoCavaille: Also nit but we can test `if err == nil { continue }` and skip this big indented block.

agent/writer.go
```diff
@@ -86,14 +124,51 @@ func (w *Writer) FlushServices() {
...
+
+			if terr, ok := err.(*apiEndpointError); ok {
+				// We could not send the payload and this is
+				// an API endpoint error, so we can try again
+				// next time we flush.
+
+				p.nextFlush = now.Add(p.resendDelay)
+				if p.resendDelay < w.conf.APIPayloadResendDelayMax {
+					p.resendDelay *= 2
+				}
```
LeoCavaille: When we reached the max I think we might want to drop the payload, it's going to be really cold data anyway? Also, not sure the exponential backoff is the best here, do you have a reasoning behind this?

agent/writer.go
LeoCavaille: Or we can reverse the slice to always flush the oldest data first :upside_down_face:

agent/trace-agent.ini
LeoCavaille: Also for the max buffer length, I don't think we actually care about the number of payloads buffered, I'd rather care about the allocated size of that buffer. It would be great to say allow 50MB and then drop.

agent/writer_test.go
```diff
@@ -143,8 +174,76 @@ receivingLoop:
...
+	// have been kept since we do not retry on 4xx errors.
+	assert.Equal(1, len(w.payloadBuffer))
+
+	p0 := w.payloadBuffer[0]
+	endpoint := p0.endpoint.(APIEndpoint)
+	assert.Equal(1, len(endpoint.apiKeys))
+	assert.Equal("key500", endpoint.apiKeys[0])
+}
+
+func TestWriterBufferization(t *testing.T) {
```
LeoCavaille: Buffering :)

GLOCKFILE
galdor: Got it. +1 for glide, which handles vendoring correctling. I've been using it for some time now.

agent/writer.go
galdor: But reversing means writing extra boilerplate or implementing `sort.Interface`. I can do either if you wish, but the current code is simpler.

agent/trace-agent.ini
galdor: I would also prefer a size in bytes, but it would means serialize each payload just to get its size. We can always cache the serialization to use it when sending though. If you don't think it's a problem, I'll do it.

agent/writer.go
galdor: Cold data are still data, if we have a chance to send them, why not ? As you wish anyway. Regarding the exponential backoff, it's a pretty common technique to avoid spamming when something goes wrong. Do you have something else in mind ?

agent/endpoint.go
galdor: No, because I then use `fmt.Fprintf`. Using a `bytes.Buffer` will yield:

```
bytes.Buffer does not implement io.Writer (Write method has pointer receiver)
```

LeoCavaille: Easily fixable https://play.golang.org/p/3NaV2nucAA but that was a nitpick anyway :)

agent/trace-agent.ini
LeoCavaille: When you've tried to flush the payload once and got an error you already serialized it so as you said shouldn't be too hard to make something relatively smart.

agent/writer.go
LeoCavaille: We should probably always prioritize flushing recent data because colder data (not a few minutes) might break the backend (especially the metrics). The flushes happen on a tick in the agent, let's say at t=0 the API breaks and the agent start buffering stuff, it fails a few times making it a resend delay of 2 minutes.
If the next flush succeeds when the API is restored, the rest is going to take 2 minutes to be flushed where it could be flushed right now. And, if it's a site-wide outage all agents will have to do the same so you're just more or less delaying the spam (if you don't have any jitter in your backoff). I don't even think the metrics agent has a backoff mechanism, could be wrong.

agent/writer.go
```diff
@@ -86,14 +142,74 @@ func (w *Writer) FlushServices() {
...
+		}
+
+		err := p.write()
+		if err == nil {
+			continue
+		}
+
+		if !w.IsPayloadBufferingEnabled() {
+			continue
+		}
```
LeoCavaille: nit but `if err == nil || !w.IsPayloadBufferingEnabled()`

agent/writer.go
```diff
@@ -86,14 +142,74 @@ func (w *Writer) FlushServices() {
...
+
+		if terr, ok := err.(*apiEndpointError); ok {
+			// We could not send the payload and this is an API
+			// endpoint error, so we can try again later.
+
+			if now.Sub(p.creationDate) > payloadMaxAge {
+				// The payload is too old, let's drop it
+				statsd.Client.Count("trace_agent.writer.dropped_payloads",
```
LeoCavaille: nit: i think we use singular usually for metrics `dropped_payload`

agent/writer.go
```diff
@@ -86,14 +142,74 @@ func (w *Writer) FlushServices() {
...
+		// We are keeping too many payloads, drop the oldest ones
+		n := 0
+		for _, p := range payloads {
+			bufSize -= p.size
+			n++
+
+			if bufSize <= w.conf.APIPayloadBufferMaxSize {
+				break
+			}
+		}
```
LeoCavaille: I think you can replace that by
```go
n := 0
for ; n < len(payloads) && bufSize > w.conf.APIPayloadBufferMaxSize; n++ {
    bufSize -= payload[n].size
}

payloads = payloads[n+1:]
```

Also I think you need to handle the edge case (unlikely but still) of having a single payload in the buffer but which is larger than the buffer max size.

galdor: Actually you need a small modification for your code to work; in your example, you will always drop at least one payload even if the payload buffer is empty. But I get the idea.

agent/endpoint.go
```diff
@@ -11,12 +11,48 @@ import (
 	log "github.com/cihub/seelog"
 )
 
+// apiEndpointError stores a list of errors triggered when sending data to a
+// list of endpoints. The endpoints member contains an api key and url for
+// each error.
+type apiEndpointError struct {
```
clutchski: i'd call this `apiErrors`

agent/endpoint.go
```diff
@@ -49,25 +85,42 @@ func NewAPIEndpoint(urls, apiKeys []string) APIEndpoint {
...
 	for i := range a.urls {
 		startFlush := time.Now()
 
+		truncKey := a.apiKeys[i]
+		if len(truncKey) > 5 {
+			truncKey = truncKey[0:5]
+		}
+		statsTags := []string{
+			fmt.Sprintf("url:%s", a.urls[i]),
+			fmt.Sprintf("apikey:%s", truncKey),
```
clutchski: let's drop these tags. api keys are secrets and it's not worth showing even some of them.

agent/writer.go
```diff
@@ -44,25 +77,43 @@ func NewWriter(conf *config.AgentConfig) *Writer {
...
+func (w *Writer) IsPayloadBufferingEnabled() bool {
+	return w.conf.APIPayloadBufferMaxSize > 0
+}
+
 // Run starts the writer and starts writing what comes through the
 // input channel.
 // NOTE: this currently happens sequentially, but it would not be too
 // hard to mutex and parallelize. Not sure it's needed though.
 func (w *Writer) Run() {
+	w.exitWG.Add(1)
```
clutchski: this already blocks until it's done no?

agent/endpoint.go
```diff
@@ -49,25 +85,42 @@ func NewAPIEndpoint(urls, apiKeys []string) APIEndpoint {
...
 
 // Write writes the bucket to the API collector endpoint.
-// Currently, the errors are just logged and we fail silently, keeping
-// writing till we have tried everything. This is because currently the
-// choice is to just drop the data we cannot write on the floor.
-// FIXME?
-func (a APIEndpoint) Write(p model.AgentPayload) {
+func (a APIEndpoint) Write(p model.AgentPayload) (int, error) {
+	countFlushResult := func(result string, tags []string) {
+		statsd.Client.Count("trace_agent.writer.flush_"+result, 1, tags, 1)
```
clutchski: in general statuses are much better to graph with tags (especially when there are a handful of them). `count by {status}` is much easier to graph than `status.error, status.success, ..`

agent/endpoint.go
```diff
@@ -79,27 +132,41 @@ func (a APIEndpoint) Write(p model.AgentPayload) {
...
+			// usually no point in trying again.
+			if resp.StatusCode/100 == 5 {
+				endpointErr.Append(a.urls[i], a.apiKeys[i], err)
+			}
+
+			countFlushResult("error", statsTags)
 			continue
 		}
 
+		countFlushResult("success", statsTags)
```
clutchski: instead of sending a statsd payload, keep a count by status and submit together at the end. no need for inline functions, etc.

agent/writer.go
```diff
@@ -44,25 +77,43 @@ func NewWriter(conf *config.AgentConfig) *Writer {
...
+		exit:   make(chan struct{}),
+		exitWG: &sync.WaitGroup{},
+
+		conf: conf,
 	}
 }
 
+// IsPayloadBufferingEnabled returns true if payload buffering is enabled or
+// false if it is not.
+func (w *Writer) IsPayloadBufferingEnabled() bool {
```
clutchski: likely should be private

agent/writer.go
```diff
@@ -86,14 +142,65 @@ func (w *Writer) FlushServices() {
...
+			// We already tried to flush recently, so there's no
+			// point in trying again right now.
+			bufferPayload(p)
+			continue
+		}
+
+		err := p.write()
+
+		if err == nil || !w.IsPayloadBufferingEnabled() {
+			continue
```
clutchski: wouldn't we want to log this?

agent/endpoint.go
```diff
@@ -11,12 +11,48 @@ import (
...
+}
+
+func (err *apiEndpointError) IsEmpty() bool {
+	return len(err.errs) == 0
+}
+
+func (err *apiEndpointError) Append(url, apiKey string, e error) {
+	err.errs = append(err.errs, e)
+	err.endpoint.urls = append(err.endpoint.urls, url)
+	err.endpoint.apiKeys = append(err.endpoint.apiKeys, apiKey)
```
clutchski: we definitely don't want to keep these anywhere.

agent/endpoint.go
clutchski: after reading the code, i would drop this struct in favour of a slice of errors and do `fmt.Errorf("error hiting %s: %s", url, err)`.

not sure it gets you anything.

agent/writer.go
```diff
@@ -1,14 +1,46 @@
...
+	size         int                // the size of the serialized payload or 0 if it has not been serialized yet
+	endpoint     AgentEndpoint      // the endpoints the payload must be sent to
+	creationDate time.Time          // the creation date of the payload
+	nextFlush    time.Time          // The earliest moment we can flush
+}
+
+func newWriterPayload(w *Writer, p model.AgentPayload) *WriterPayload {
+	return &WriterPayload{
+		payload:      p,
+		endpoint:     w.endpoint,
```
clutchski: if you only need the writer to get an endpoint, just pass in the endpoint directly. it's more loosely coupled and easier to understand.

agent/writer.go
```diff
@@ -19,12 +51,13 @@ type Writer struct {
 	inPayloads chan model.AgentPayload     // main payloads for processed traces/stats
 	inServices chan model.ServicesMetadata // secondary services metadata
 
-	payloadBuffer []model.AgentPayload   // buffered of payloads ready to send
+	payloadBuffer []*WriterPayload       // buffer of payloads ready to send
```
clutchski: why do these need to be pointers?

agent/writer.go
```diff
@@ -1,14 +1,46 @@
...
 
+// the amount of time in seconds to wait before resending a payload
+const payloadResendDelay = 5 * time.Second
+
+// the amount of time in seconds a payload can stay buffered before being dropped
+const payloadMaxAge = 10 * time.Minute
+
+// WriterPayload wraps a model.AgentPayload and keeps track of a list of
+// endpoints the payload must be sent to.
+type WriterPayload struct {
```
clutchski: probably shouldn't be public as well.

agent/writer.go
```diff
@@ -44,25 +77,43 @@ func NewWriter(conf *config.AgentConfig) *Writer {
...
+	defer flushTicker.Stop()
+
 	for {
 		select {
 		case p := <-w.inPayloads:
 			if p.IsEmpty() {
 				continue
 			}
-			w.payloadBuffer = append(w.payloadBuffer, p)
+			np := newWriterPayload(w, p)
```
clutchski: nitpick: i'd shortner `writerPayload` to `wp` not `np`.

agent/writer.go
```diff
@@ -44,25 +77,43 @@ func NewWriter(conf *config.AgentConfig) *Writer {
...
+			w.payloadBuffer = append(w.payloadBuffer, np)
+			w.Flush()
```
clutchski: we flush to the API on every payload now? are we sure about this?

agent/writer.go
```diff
@@ -86,14 +142,65 @@ func (w *Writer) FlushServices() {
...
-func (w *Writer) getPayloadBufferLen() int32 {
-	return atomic.LoadInt32(&w.payloadBufferLen)
+	// Drop payloads to respect the buffer size limit if necessary.
+	nbDrops := 0
+	for n := 0; n < len(payloads) && bufSize > w.conf.APIPayloadBufferMaxSize; n++ {
+		bufSize -= payloads[n].size
+		nbDrops++
+	}
+
+	log.Infof("dropping %d payloads (payload buffer full)", nbDrops)
```
clutchski: add in `if nbDrops > 0>` here

agent/writer.go
```diff
@@ -86,14 +142,65 @@ func (w *Writer) FlushServices() {
...
 	}
-	// regardless if the http post was was success or not. We don't want to buffer
-	// data in case of api outage. See also endpoint.Write() comment.
-	w.payloadBuffer = nil
-}
 
-func (w *Writer) getPayloadBufferLen() int32 {
-	return atomic.LoadInt32(&w.payloadBufferLen)
+	// Drop payloads to respect the buffer size limit if necessary.
+	nbDrops := 0
```
clutchski: nit: i'd call this `ndrops` or `drops`.

agent/endpoint.go
galdor: Encapsuling in a type implementing `error` gives the ability to return it directly. And I need the list of endpoints for whom the request failed to retry with them only, instead of all original endpoints, hence the `endpoints` member.

agent/writer.go
LeoCavaille: agreed let's make it private

agent/endpoint.go
galdor: It was here before I wrote these patches :) I can remove them if you wish.

agent/writer.go
galdor: When `Stop` is called, the goroutine will read on `w.exit`, and flush one last time before exiting. We want `Stop` to block until the goroutine actuall exits, i.e. after the flush. The wait group gives us this property.

agent/writer.go
LeoCavaille: I think because the buffer is recreated when we decide to drop or not the data it's probably better to have pointers, not the biggest deal in the world but we can benchmark it.

agent/writer.go
galdor: `APIEndpoint.Write` already logs errors. Therefore if `p.write()` fails and buffering is disabled, these errors will already be logged.

agent/writer.go
LeoCavaille: This was already the case, here the ticker was added.

agent/endpoint.go
galdor: We keep them in buffered payloads to be able to resend them to the endpoints that could not be contacted.

agent/writer.go
galdor: Good point; there was other values before, but this was modified :)

agent/writer.go
LeoCavaille: The err is already logged in the `write` method, and if buffering is disabled by the user, he probably does not want to see it in the logs either.

agent/endpoint.go
LeoCavaille: You need to keep track of these to resubmit the data because payloads can be sent to multiple endpoints and fail only for some of them.

agent/endpoint.go
LeoCavaille: We might also drop URL actually now that I think about it, using multiple endpoints is really rare so 99.99% of the cases should be just fine without tags.

agent/writer.go
galdor: `WriterPayload` has of size of 152 bytes. Beyond 8 ou 16 bytes, copying could be expensive. That could be benchmarked of course.

agent/writer.go
galdor: It was `np` for "new payload" since there is `p` for payload right before. I can inline it anyway.

agent/writer.go
galdor: This was the case before my commits.

PR #171, opened 2016-12-31
[optimization] improving pipeline entrypoint performance

palazzem:
### What it does

* improves the ``ClientDecoder`` interface to make decoders implementations reusable
* implements the ``jsonDecoder`` and ``msgpackDecoder`` structs to keep references to the internal buffers and decoders; in this way decoders are reusable multiple times
* ``readAll`` re-implementation that overwrites the given buffer without any kind of allocations (the built-in version allocates a new resizable ``Buffer`` and returns a slice descriptor for each request)
* when a decoder is created with the internal ``Buffer``, the buffer has ``minBufferSize`` size. This improves a lot performances because during each request the buffer is resized many times to contain the ``req.Body``
* the request handler hot path, doesn't make useless allocations when a request is handled without errors.

#### Performance impact
The benchmark includes the trace handling for a single trace with a single span, so the performance gain is linear with the number of traces. Furthermore, because we're not allocating anything (obviously with the exception of the ``[]Trace`` and the built-in server objects), we're reducing the number of GC.

```
benchmark                   old ns/op     new ns/op     delta
BenchmarkHandleTraces-4     13198         9723          -26.33%

benchmark                   old allocs     new allocs     delta
BenchmarkHandleTraces-4     33             23             -30.30%

benchmark                   old bytes     new bytes     delta
BenchmarkHandleTraces-4     6530          1408          -78.44%
```

### DIFF
--- a/agent/receiver.go
+++ b/agent/receiver.go
@@ -17,11 +17,14 @@ import (
 // APIVersion is a dumb way to version our collector handlers
 type APIVersion int
 
-// Max size of decoders pool
-const decoderSize = 10
+const (
+	decoderSize       = 10 // Max size of decoders pool
+	tagTraceHandler   = "handler:traces"
+	tagServiceHandler = "handler:services"
+)
 
 // Pool of decoders to prevent continuous allocations
-// TODO[manu]: don't make it global
+// TODO[manu]: don't make it global even if it's thread-safe (uses channels)
 var decoderPool *model.DecoderPool = model.NewDecoderPool(decoderSize)
 
 const (
@@ -111,7 +114,6 @@ func (r *HTTPReceiver) Run() {
 
 // handleTraces knows how to handle a bunch of traces
 func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *http.Request) {
-	handlerTags := []string{"handler:traces", fmt.Sprintf("v:%d", v)}
 	// we need an io.ReadSeeker if we want to be able to display
 	// error feedback to the user, otherwise r.Body is trash
 	// once it's been decoded
@@ -127,7 +129,7 @@ func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *ht
 	case v01:
 		if contentType != "application/json" && contentType != "text/json" && contentType != "" {
 			r.logger.Errorf("rejecting client request, unsupported media type: '%s'", contentType)
-			HTTPFormatError(handlerTags, w)
+			HTTPFormatError([]string{tagTraceHandler, fmt.Sprintf("v:%d", v)}, w)
 			return
 		}
 
@@ -138,7 +140,7 @@ func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *ht
 		if err != nil {
 			r.logger.Errorf(model.HumanReadableJSONError(dec.BufferReader(), err))
 			decoderPool.Release(dec)
-			HTTPDecodingError(handlerTags, w)
+			HTTPDecodingError([]string{tagTraceHandler, fmt.Sprintf("v:%d", v)}, w)
 			return
 		}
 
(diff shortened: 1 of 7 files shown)


agent/receiver.go
```diff
@@ -21,6 +18,16 @@ import (
...
 
 const (
+	decoderSize       = 10 // Max size of decoders pool
+	tagTraceHandler   = "handler:traces"
+	tagServiceHandler = "handler:services"
+)
+
+// Pool of decoders to prevent continuous allocations
+// TODO[manu]: don't make it global even if it's thread-safe (uses channels)
+var decoderPool = model.NewDecoderPool(decoderSize)
```
palazzem: I don't like globals. We can put it on an ``init()`` method (that is still a global) or just change the signatures of ``httpHandleWithVersion`` so that it accepts a decoder pool that is initialized in the ``Run()`` method.

Suggestions about what is more correct/idiomatic?

agent/receiver_responses.go
```diff
@@ -30,5 +31,5 @@ func HTTPEndpointNotSupported(tags []string, w http.ResponseWriter) {
 // HTTPOK is a dumb response for when things are a OK
 func HTTPOK(w http.ResponseWriter) {
 	w.WriteHeader(http.StatusOK)
-	w.Write([]byte("OK\n"))
+	io.WriteString(w, "OK\n")
```
palazzem: using a ``WriteString`` removes the slice allocation for each request

agent/receiver_test.go
```diff
@@ -7,46 +7,14 @@ import (
...
-	"time"
 
 	"github.com/DataDog/datadog-trace-agent/config"
+	"github.com/DataDog/datadog-trace-agent/fixtures"
 	"github.com/DataDog/datadog-trace-agent/model"
 	"github.com/stretchr/testify/assert"
 	"github.com/ugorji/go/codec"
 )
 
-// getTestSpan returns a Span with different fields set
```
palazzem: just housekeeping to avoid circular imports in testing

agent/receiver_test.go
```diff
@@ -378,9 +346,44 @@ func TestReceiverServiceMsgpackDecoder(t *testing.T) {
...
+	receiver := NewHTTPReceiver(config)
+
+	// response recorder
+	handler := http.HandlerFunc(httpHandleWithVersion(v03, receiver.handleTraces))
+
+	// benchmark
+	b.ResetTimer()
+	b.ReportAllocs()
+	for n := 0; n < b.N; n++ {
+		b.StopTimer()
```
palazzem: looked at the ``StopTimer()`` code and when it's used, it stops the timer and doesn't count allocations

agent/receiver_test.go
```diff
@@ -378,9 +346,44 @@ func TestReceiverServiceMsgpackDecoder(t *testing.T) {
...
+		// consume the traces channel without doing anything
+		select {
+		case <-receiver.traces:
```
palazzem: since we don't have a background daemon, we should consume that channel otherwise we get stuck when the channel buffer is full

model/client_api.go
```diff
@@ -1,26 +1,190 @@
...
 )
 
+// average size of a buffer; because usually the payloads are huge,
+// this value ensures that initialized buffers are big enough so that
+// the resize operation is not usually called while reading the response.
+const minBufferSize = 512
+
+// readAll reads from source until an error or EOF and writes to dest;
+// if the dest buffer contains data, it is truncated
+func readAll(source io.Reader, dest *bytes.Buffer) (err error) {
```
palazzem: when reading we don't allocate anything since the ``dest`` buffer is ``Truncated(0)``. Buffer resizes are preserved.

model/client_api.go
```diff
@@ -1,26 +1,190 @@
...
 	"encoding/json"
 	"io"
 
 	"github.com/ugorji/go/codec"
 )
 
+// average size of a buffer; because usually the payloads are huge,
+// this value ensures that initialized buffers are big enough so that
+// the resize operation is not usually called while reading the response.
+const minBufferSize = 512
```
palazzem: this is the default value for the stdlib, we may want to update it with a bigger value

agent/receiver.go
palazzem: +1

PR #172, opened 2017-01-02
[profile] more statsd log

ufoot:
Reporting more informations using statsd, more precisely:
- tags now report version, go version, and git commit
- report go memory stats
- report cpu usage (parses /proc/pid/stat so not cross-platform)

### DIFF
--- a/agent/receiver.go
+++ b/agent/receiver.go
@@ -4,12 +4,14 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
 
 	"github.com/DataDog/datadog-trace-agent/config"
 	"github.com/DataDog/datadog-trace-agent/model"
+	"github.com/DataDog/datadog-trace-agent/profile"
 	"github.com/DataDog/datadog-trace-agent/statsd"
 	log "github.com/cihub/seelog"
 )
@@ -37,6 +39,20 @@ const (
 	// Services: msgpack/JSON, map[string]map[string][string]
 	v03
 )
+
+var receiverTags []string
+
+func init() {
+	if Version != "" {
+		receiverTags = append(receiverTags, "version:"+Version)
+	}
+	if GoVersion != "" {
+		receiverTags = append(receiverTags, "go_version:"+GoVersion)
+	}
+	if GitCommit != "" {
+		receiverTags = append(receiverTags, "git_commit:"+GitCommit)
+	}
+}
 
 func httpHandleWithVersion(v APIVersion, f func(APIVersion, http.ResponseWriter, *http.Request)) func(http.ResponseWriter, *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
@@ -122,6 +138,7 @@ func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *ht
 
 	var traces []model.Trace
 	contentType := req.Header.Get("Content-Type")
+	contentLength := req.Header.Get("Content-Length")
 
 	switch v {
 	case v01:
@@ -183,6 +200,12 @@ func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *ht
 	}
 
 	HTTPOK(w)
+
+	atomic.AddInt64(&r.stats.PayloadsReceived, 1)
+	l, err := strconv.Atoi(contentLength)
+	if err == nil {
+		atomic.AddInt64(&r.stats.BytesReceived, int64(l))
+	}
 
 	// normalize data
 	for i := range traces {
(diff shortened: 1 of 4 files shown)


clutchski:
lets discuss this one before shipping.

agent/receiver.go
```diff
@@ -272,6 +295,12 @@ func (r *HTTPReceiver) handleServices(v APIVersion, w http.ResponseWriter, req *
 func (r *HTTPReceiver) logStats() {
 	for range time.Tick(60 * time.Second) {
 		// Load counters and reset them for the next flush
+		payloads := atomic.LoadInt64(&r.stats.PayloadsReceived)
+		r.stats.PayloadsReceived = 0
+
+		bytes := atomic.LoadInt64(&r.stats.BytesReceived)
+		r.stats.BytesReceived = 0
+
```
galdor: Note that writing `r.stats.BytesReceived` is unsafe. Generally speaking, I would suggest adding a mutex to `receiverStats` and simply lock it to access all stats.

profile/cpu.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
+		return nil, err
+	}
+	ci.Cstime += int64(cstime)
+	if err != nil {
+		return &ci, nil
+	}
+
+	return &ci, nil
+}
```
galdor: `man proc(5)` provides the format definitions for all fields in `/proc/[pid]/stat`, it might be simpler to use `fmt.Scanf`.

profile/cpu.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// There are functions in https://golang.org/pkg/os/#ProcessState to get
+	// process info but they all refer to children as here, we want to get
+	// info about this process.
+	path := fmt.Sprintf("/proc/%d/stat", os.Getpid())
+	content, err := ioutil.ReadFile(path)
+	if err != nil && len(content) > 0 {
+		return nil, fmt.Errorf("unable to read CPU info from '%s': %v", path, err)
+	}
+	fields := strings.Split(string(content), " ")
```
galdor: `man proc(5)` provides the format definitions for all fields in `/proc/[pid]/stat`, it might be simpler to use `fmt.Scanf`.

agent/receiver.go
LeoCavaille: or use https://golang.org/pkg/sync/atomic/#StoreInt64

galdor: If you are using multiple atomic load/store to access the same structure at the same time, using a mutex would both make the code shorter and guarantee a coherent view of the members of the structure. And using a single mutex could even be faster than multiple locking instructions (but the difference is barely worth mentioning).

Just my 2 cents :)

agent/receiver.go
```diff
@@ -284,19 +313,31 @@ func (r *HTTPReceiver) logStats() {
...
+		statsd.Client.Count("trace_agent.receiver.span_dropped", sdropped, receiverTags, 1)
+		statsd.Client.Count("trace_agent.receiver.trace_dropped", tdropped, receiverTags, 1)
+
+		profile.CPUStatsd(profile.IntakeStats{
+			Payloads: payloads,
+			Traces:   traces - tdropped,
+			Spans:    spans - tdropped,
+			Bytes:    bytes,
+		}, receiverTags)
+		profile.MemStatsd(receiverTags)
```
LeoCavaille: I don't think memstats this has to do much with the receiver? Maybe we can just have a profiling routine in the main.go controlled by a flag that users can disable at will (or that we can tell user to enable for debugging).
Also have you looked at http://docs.datadoghq.com/integrations/goexpvar/ maybe it could be faster to reach the goal of getting the metrics and see which one we'd really use/need.

profile/cpu.go
```diff
@@ -0,0 +1,155 @@
+package profile
```
LeoCavaille: I think I'd much rather use something already packaged like https://godoc.org/github.com/shirou/gopsutil/process if we want to go this way. Gives us directly compatibility across kernels and OS, `psutil` is the lib that's used in the process check of the agent.

agent/receiver.go
```diff
@@ -38,6 +40,20 @@ const (
...
+func init() {
+	if Version != "" {
+		receiverTags = append(receiverTags, "version:"+Version)
+	}
+	if GoVersion != "" {
+		receiverTags = append(receiverTags, "go_version:"+GoVersion)
+	}
+	if GitCommit != "" {
+		receiverTags = append(receiverTags, "git_commit:"+GitCommit)
+	}
```
LeoCavaille: Do we need this? Version should be enough no?

agent/receiver.go
LeoCavaille: Agreed for "faster" but here I think the reasoning is cleaner/leaner business code. IMO, it's always better to make instrumentation light & inconspicuous.
Example in this block of code https://github.com/DataDog/datadog-trace-agent/blob/a6d6149f532e7c391fa9af92dc0c4c0c7b3a51f3/agent/receiver.go#L206-L234 or with that simple example, 1 line vs 3 lines.

```go
var bytes int64
func do() {
    // some code doing stuff
    atomic.AddInt64(&bytes, 1)
    // some more code doing more stuff
}

func logStats {
    b := atomic.LoadInt64(&bytes)
    fmt.Println(b)
    atomic.StoreInt64(&bytes, 0)
}
```

versus


```go
var struct {
    bytes int64
    sync.Mutex
} stats

func do() {
    // some code doing stuff
    stats.Lock()
    stats.bytes += 1
    stats.Unlock()
    // some more code doing more stuff
}

func logStats {
    stats.Lock()
    fmt.Println(b.bytes)
    b.bytes = 0
    stats.Unlock()
}

clutchski:
so in general not 100% sure this is a problem we need to solve today. i wouldn't ship this until we've hit the benchmarks.

agent/receiver.go
ufoot: Thanks for taking the time to spot this.

ufoot:
Superseded by https://github.com/DataDog/datadog-trace-agent/pull/182

PR #174, opened 2017-01-03
[receiver-logger] dumb mistake in skip error count

LeoCavaille:
This would output negative values.

### DIFF
--- a/agent/receiver_logger.go
+++ b/agent/receiver_logger.go
@@ -30,8 +30,8 @@ func (l *errorLogger) Errorf(format string, params ...interface{}) {
 
 func (l *errorLogger) Reset() {
 	l.Lock()
-	if l.errors > 0 {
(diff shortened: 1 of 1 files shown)


galdor:
As far as I understand the way the interval work, it seems good to me. At least mathematically speaking :)

agent/receiver_logger.go
```diff
@@ -30,7 +30,7 @@ func (l *errorLogger) Errorf(format string, params ...interface{}) {
 
 func (l *errorLogger) Reset() {
 	l.Lock()
-	if l.errors > 0 {
+	if l.errors > maxPerInterval {
```
ufoot: I think it's `>=` (this one will only display stuff when at least 2 lines are skipped)

galdor: If there are as many errors as `maxPerInterval` then nothing was skipped, right ?

PR #177, opened 2017-01-03
[profile] unpacking "_resource" as resource in tags

ufoot:
In many implementations, resource is sent in the resource field and
as a tag, since agent can decide to treat them differently. Typically,
resource would be quantized, but not the value in the tagset. With
this patch, a client implementation can just send "_resource" instead
of the complete resource value, to save bandwith.

### DIFF
--- a/model/meta.go
+++ b/model/meta.go
@@ -1,0 +1,25 @@
+package model
+
+const (
+	refResource = "_resource"
+)
+
+// MetaCompact compacts tags so that if they refer to an existing field, they are not duped
+func (s *Span) MetaCompact() {
+	for k, v := range s.Meta {
+		switch v {
+		case s.Resource:
+			s.Meta[k] = refResource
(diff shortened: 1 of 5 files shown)


clutchski:
@ufoot totally agreed they should only be sent once. but i'm not sure about the `_resource` thing. why not just send the `resource` only?

LeoCavaille:
Agreed with @clutchski , I think the meta key should just be set by the quantizer code itself like we do here https://github.com/DataDog/datadog-trace-agent/blob/master/quantizer/sql.go#L228-L232

Should be true for all types handled, the clients should just care about the resource and that's all.

ufoot:
So I'm going to close this one, because the change should indeed be different -> when we quantize, quantize resource only and store the original value in some normalized field. Then second part of the fix: stop sending that extra tag in the client implementations.

PR #180, opened 2017-01-06
[decoder] switching decoding library with safe casting

palazzem:
### What it does

* changes the decoding library from ``github.com/ugorji/go`` to ``github.com/tinylib/msgp``
* uses types during the struct decoding, improving a lot the decoding speed and allocations
* safe casting if other msgpack libraries send ``int64`` fields in ``uint64`` over the wire; this ensures that the struct decoding takes care about safe casting (safe in the sense of prevent overflows) 

#### uint64 -> int64

We should take care about this problem because ``github.com/tinylib/msgp`` doesn't allow to read ``uint64`` values in ``int64`` structs fields even if the original value was an ``int64``. The reason of why this happens is explained below.

According [to this discussion][1], msgpack-c implementation ensures:
> If an object can be represented in multiple possible output formats, serializers SHOULD use the format which represents the data in the smallest number of bytes.

This means that if an ``int64`` is positive, it will be sent using the ``uint64`` bytes representation to reduce the number of used bytes. To handle that case, the decoding phase does:
* try to parse the ``int64`` for ``int64`` struct fields (``Start`` and ``Duration``)
* if it fails, fallback to ``uint64`` and
* if the ``uint64`` value is ``> math.MaxInt64``, drop the value and return an error
* if the ``uint64`` is ``<= math.MaxInt64``, cast the value

This cast makes sense **only because** msgpack specification ensures that this is the way how ``int64`` fields are sent over the wire.

#### Reusing payloads

This approach allows us to reuse payloads so that the number of allocations are completely dropped to a reasonable value (``-45.97%``). In the current PR, payloads are not reused.

#### Performance improvements

```
benchmark                     old ns/op     new ns/op      delta
BenchmarkDecoderMsgpack-4     37520388      15106440       -59.74%

benchmark                     old allocs    new allocs     delta
BenchmarkDecoderMsgpack-4     129007        109053         -15.47%

benchmark                     old bytes     new bytes      delta
BenchmarkDecoderMsgpack-4     9832704       8875122        -9.74%
```

[1]: https://github.com/msgpack/msgpack/issues/164#issuecomment-86368081


model/trace.go
```diff
@@ -4,7 +4,7 @@ import (
 	log "github.com/cihub/seelog"
 )
 
-//go:generate msgp
+//go:generate msgp -marshal=false
```
ufoot: This might be the right place to have a small comment mentioning the possible side effects when running go generate (eg: should some files be manually modified afterwards)

model/trace_gen.go
```diff
@@ -42,75 +42,40 @@ func (z Trace) EncodeMsg(en *msgp.Writer) (err error) {
...
-	o = bts
-	return
-}
-
 // Msgsize returns an upper bound estimate of the number of bytes occupied by the serialized message
 func (z Trace) Msgsize() (s int) {
 	s = msgp.ArrayHeaderSize
-	for zhct := range z {
-		s += z[zhct].Msgsize()
+	for zcmr := range z {
```
ufoot: Joy of generated code, looks like those var names are sort of randomly picked?

palazzem: Yes, they do

PR #182, opened 2017-01-09
[watchdog] basic funcs returning CPU and Mem usage

ufoot:
Basic funcs returning:
* CPU usage (value between 0 and number of CPU, 0.3 means 30% of one CPU)
* Mem usage (currently allocated bytes from the point of view of the Go memory manager)
* Byte allocation rate (bytes allocated per second)
* Number of TCP Connections

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -42,6 +41,8 @@ type Agent struct {
 
 	// Used to synchronize on a clean exit
 	exit chan struct{}
+
+	die func(format string, args ...interface{})
 }
 
 // NewAgent returns a new Agent object, ready to be started
@@ -65,6 +66,7 @@ func NewAgent(conf *config.AgentConfig) *Agent {
 		Writer:       w,
 		conf:         conf,
 		exit:         exit,
+		die:          die,
 	}
 }
 
@@ -163,8 +165,6 @@ func (a *Agent) watchdog() {
 	statsd.Client.Gauge("trace_agent.process.mem.alloc_per_sec", m.AllocPerSec, tags, 1)
 
 	if float64(m.Alloc) > a.conf.MaxMemory {
-		msg := fmt.Sprintf("exceeded max memory (current=%d, max=%d)", m.Alloc, int64(a.conf.MaxMemory))
-		log.Error(msg)
-		panic(msg)
+		a.die("exceeded max memory (current=%d, max=%d)", m.Alloc, int64(a.conf.MaxMemory))
 	}
 }
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -8,8 +8,12 @@ import (
 	"github.com/DataDog/datadog-trace-agent/config"
 	"github.com/DataDog/datadog-trace-agent/model"
 	"github.com/DataDog/datadog-trace-agent/quantizer"
+	"github.com/DataDog/datadog-trace-agent/statsd"
+	"github.com/DataDog/datadog-trace-agent/watchdog"
 	log "github.com/cihub/seelog"
 )
+
+const processStatsInterval = time.Minute
 
 type processedTrace struct {
 	Trace     model.Trace
@@ -67,6 +71,8 @@ func NewAgent(conf *config.AgentConfig) *Agent {
 func (a *Agent) Run() {
 	flushTicker := time.NewTicker(a.conf.BucketInterval)
 	defer flushTicker.Stop()
+	statsTicker := time.NewTicker(processStatsInterval)
+	defer statsTicker.Stop()
 
 	a.Receiver.Run()
 	a.Writer.Run()
@@ -95,6 +101,8 @@ func (a *Agent) Run() {
 			wg.Wait()
 
 			a.Writer.inPayloads <- p
+		case <-statsTicker.C:
+			processStats()
 		case <-a.exit:
 			log.Info("exiting")
 			close(a.Receiver.exit)
@@ -143,4 +151,13 @@ func (a *Agent) Process(t model.Trace) {
 	weight := pt.weight() // need to do this now because sampler edits .Metrics map
 	go a.Concentrator.Add(pt, weight)
 	go a.Sampler.Add(pt)
+}
+
+func processStats() {
+	tags := []string{"version:" + Version, "go_version:" + GoVersion}
+	c := watchdog.CPU()
+	statsd.Client.Gauge("trace_agent.process.cpu.pct", c.UserAvg*100.0, tags, 1)
+	m := watchdog.Mem()
+	statsd.Client.Gauge("trace_agent.process.mem.alloc", float64(m.Alloc), tags, 1)
+	statsd.Client.Gauge("trace_agent.process.mem.alloc_per_sec", m.AllocPerSec, tags, 1)
 }
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -71,8 +72,8 @@ func NewAgent(conf *config.AgentConfig) *Agent {
 func (a *Agent) Run() {
 	flushTicker := time.NewTicker(a.conf.BucketInterval)
 	defer flushTicker.Stop()
-	statsTicker := time.NewTicker(processStatsInterval)
-	defer statsTicker.Stop()
+	watchdogTicker := time.NewTicker(watchdogInterval)
+	defer watchdogTicker.Stop()
 
 	a.Receiver.Run()
 	a.Writer.Run()
@@ -101,8 +102,8 @@ func (a *Agent) Run() {
 			wg.Wait()
 
 			a.Writer.inPayloads <- p
-		case <-statsTicker.C:
-			processStats()
+		case <-watchdogTicker.C:
+			a.watchdog()
 		case <-a.exit:
 			log.Info("exiting")
 			close(a.Receiver.exit)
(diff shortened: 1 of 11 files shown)


agent/agent.go
```diff
@@ -131,3 +138,18 @@ func (a *Agent) Process(t model.Trace) {
 	go a.Concentrator.Add(pt)
 	go a.Sampler.Add(pt)
 }
+
+func (a *Agent) watchdog() {
```
LotharSee: Is this function fast enough? If it happen to take time, it will impact the Agent main loop.

config/agent.go
```diff
@@ -120,6 +124,9 @@ func NewDefaultAgentConfig() *AgentConfig {
 
 		LogLevel:    "INFO",
 		LogFilePath: "/var/log/datadog/trace-agent.log",
+
+		MaxMemory:        1e9,
+		WatchdogInterval: time.Duration(1) * time.Minute,
```
LotharSee: Can we go with a shorter interval default? something like 10s? The sampler will certainly need fresh data.

watchdog/info.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/shirou/gopsutil/process"
+)
+
+// CPUInfo contains very basic CPU information
+type CPUInfo struct {
+	// UserAvg is the average of the user CPU usage since last time
+	// it was polled. 0 means "not used at all" and 1 means "1 CPU was
+	// totally full for that period". So it might be greater than 1 if
+	// the process is monopolizing several cores.
+	UserAvg float64
```
LotharSee: Would it be hard to normalize it by the number of CPUs?

ufoot: No, it's not hard. It's just a choice. I feel what we want is an agent that does not go over 3% of one core. An agent eating up 3% of global CPU on a 16-core machine is consuming almost 50% of a single core, so it's misbehaving.

config/agent.go
ufoot: Yes, 10 sec is fine.

agent/agent.go
ufoot: According to the go bench, `Mem()` represents 90 microseconds and `CPU()` 6 microseconds. So it should be fine to call them every X seconds. But the `statsd` part seem to be quite time consuming, the complete thing is almost 1 millisecond. It's fine if it's only running once per minute.

clutchski:
i'm interested in this, but let's wait until we have a clear strategy to merge into sampling before we merge / proceed.

watchdog/info.go
LotharSee: Isn't the Agent properly running in multiple threads/core? As long as we don't set a bad GOMAXPROCS.
Plus, if you are on a 16-core machine, you can expect a high throughput of traces, so it is normal to be proportional there (like if it were N machines).

LotharSee:
Since we haven't implemented a usage for that, it could wait before merging it.
Just to be sure what we bring here will really suit what we need.

But if we really want it to go out first, I'd be good with it.

agent/agent.go
```diff
@@ -147,11 +148,17 @@ func (a *Agent) Process(t model.Trace) {
...
 	c := watchdog.CPU()
 	statsd.Client.Gauge("trace_agent.process.cpu.pct", c.UserAvg*100.0, tags, 1)
 	m := watchdog.Mem()
 	statsd.Client.Gauge("trace_agent.process.mem.alloc", float64(m.Alloc), tags, 1)
 	statsd.Client.Gauge("trace_agent.process.mem.alloc_per_sec", m.AllocPerSec, tags, 1)
+
+	if float64(m.Alloc) > a.conf.MaxMemory {
+		msg := fmt.Sprintf("exceeded max memory (current=%d, max=%d)", m.Alloc, int64(a.conf.MaxMemory))
+		log.Error(msg)
+		panic(msg)
```
galdor: There is no reason to panic here, the stacktrace won't be useful in any way. Using `die()` will both log and error and exit.

watchdog/info.go
```diff
@@ -0,0 +1,200 @@
...
+	"sync"
+	"time"
+
+	log "github.com/cihub/seelog"
+	"github.com/shirou/gopsutil/net"
+	"github.com/shirou/gopsutil/process"
+)
+
+const (
+	cacheDelay = time.Second
```
ufoot: @LotharSee there, you probably want something longer (10 sec?) if you want the CPU info, typically, to be somewhat stable

agent/listener.go
```diff
@@ -67,6 +67,7 @@ func (sl *StoppableListener) Accept() (net.Conn, error) {
 		select {
 		case <-sl.exit:
 			log.Debug("stopping listener")
+			sl.TCPListener.Close()
```
LotharSee: Were we missing that? What's the impact?

config/agent.go
```diff
@@ -133,6 +138,10 @@ func NewDefaultAgentConfig() *AgentConfig {
 
 		LogLevel:    "INFO",
 		LogFilePath: "/var/log/datadog/trace-agent.log",
+
+		MaxMemory:        1e9,
+		MaxConnections:   5000,
+		WatchdogInterval: time.Duration(1) * time.Minute,
```
LotharSee: `1 * time.Minute` should be enough

glide.lock
```diff
@@ -1,5 +1,5 @@
-hash: 82b158363973fe2a76243c75d649380286b2dd97e6f263789a3217f0f503316e
-updated: 2017-02-22T16:29:03.841539063+01:00
+hash: 6ee4eb022b387621ddb234375954dc60256fc2df4dcf2099dfb2f55f645756e4
```
LotharSee: Can't we avoid putting subdeps in this file?
Stuff like `w32` or `wmi` are not directly imported by our code, it would be better if we could manage them differently.

watchdog/info_test.go
```diff
@@ -0,0 +1,248 @@
...
+	c = CPU()
+	t.Logf("CPU (sleep): %v", c)
+
+	// checking that CPU is low enough, this is theorically flaky,
+	// but eating 50% of CPU for a time.Sleep is still not likely to happen often
+	assert.Condition(func() bool { return c.UserAvg >= 0.0 }, fmt.Sprintf("cpu avg should be positive, got %f", c.UserAvg))
+	assert.Condition(func() bool { return c.UserAvg <= 0.5 }, fmt.Sprintf("cpu avg should be below 0.5, got %f", c.UserAvg))
+}
+
+func doTestCPUHigh(t *testing.T, n int) {
```
LotharSee: How were these 2 tests behaving locally/in the CI? Are they stable?

watchdog/info_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+	// Checking that CPU is high enough, a very simple ++ loop should be
+	// enough to stimulate one core and make it over 50%. One of the goals
+	// of this test is to check that values are not wrong by a factor 100, such
+	// as mismatching percentages and [0...1]  values.
+	assert.Condition(func() bool { return c.UserAvg >= 0.5 }, fmt.Sprintf("cpu avg is too low, got %f", c.UserAvg))
+	assert.Condition(func() bool { return c.UserAvg <= float64(n+1) }, fmt.Sprintf("cpu avg is too high, target is %d, got %f", n, c.UserAvg))
+}
+
+func TestCPUHigh(t *testing.T) {
```
LotharSee: What's the purpose of testing with more goroutines?

watchdog/info.go
```diff
@@ -0,0 +1,204 @@
...
+	if err != nil {
+		log.Errorf("unable to create global Process: %v", err)
+	}
+}
+
+// NewCurrentInfo creates a new CurrentInfo referring to the current running program.
+func NewCurrentInfo() (*CurrentInfo, error) {
+	p, err := process.NewProcess(int32(os.Getpid()))
+	if err != nil {
+		log.Debugf("unable to create Process: %v", err)
```
LeoCavaille: This is a duplicate with the log above I think?

watchdog/info_test.go
```diff
@@ -0,0 +1,248 @@
...
+	globalCurrentInfo.cacheDelay = testDuration
+	time.Sleep(testDuration)
+	m := Mem()
+	t.Logf("Mem (sleep): %v", m)
+
+	// Checking that Mem is low enough, this is theorically flaky,
+	// unless some other random GoRoutine is running, figures should remain low
+	assert.Condition(func() bool { return int64(m.Alloc)-int64(oldM.Alloc) <= 1e4 }, "over 10 Kb allocated since last call, way to high for almost no operation")
+	assert.Condition(func() bool { return m.Alloc <= 1e8 }, "over 100 Mb allocated, way to high for almost no operation")
+	assert.Condition(func() bool { return m.AllocPerSec >= 0.0 }, "allocs per sec should be positive")
```
LeoCavaille: Maybe just writing those tests as `assert.True(a > b, "error")` is simpler?

agent/listener.go
ufoot: Basically, if you don't close it, then you obviously can't listen again on the same port. This is very annoying in tests. And it could be annoying if agent was embedded in some other program that would start/stop it. As now what we do in the agent is just start the program, start the listener, and when we stop, quit the program, you'd never notice as when you exit the program, all file descriptors are closed by the OS. I still advocate for closing it cleanly.

watchdog/info_test.go
ufoot: Yes, stable.

watchdog/info_test.go
ufoot: Oh, just checking if there are no corner cases, in practice 10 and 100 yield the same kind of results, but 1 and 10 are quite different as 10 uses several procs for real.

agent/agent_test.go
```diff
@@ -1,12 +1,76 @@
...
+	"runtime"
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/DataDog/datadog-trace-agent/config"
 	"github.com/DataDog/datadog-trace-agent/fixtures"
 )
 
+func TestWatchdog(t *testing.T) {
```
ufoot: Thinking of it, one way to do it is to set ridiculously low values for thresholds, such as "10 kb max memory" and "1 network connection opened". In those cases the program should crash in black-box testing. And those can be provided through a conf file, so would need some testing plumbery, but really doable.

PR #184, opened 2017-01-10
 [profile] using one struct to aggregate stats

ufoot:
Using one struct to aggregate all stats, instead of several structs pointed by multiple maps. This speeds up lookup operations. A side effect is the introduction of an "anti corruption layer" (the data we use for internal computation is not the one used to communicate with the API).


model/statsraw.go
```diff
@@ -0,0 +1,228 @@
...
+
+type statsKey struct {
+	name string
+	aggr string
+}
+
+// StatsRawBucket is used to compute span data and aggregate it
+// within a time-framed bucket. This should not be used outside
+// the agent, use StatsBucket for this.
+type StatsRawBucket struct {
```
clutchski: there's no need for this to be used outside of the agent. can we make it private?

model/statsraw.go
```diff
@@ -0,0 +1,228 @@
...
+
+	"github.com/DataDog/datadog-trace-agent/quantile"
+)
+
+// Most "algorithm" stuff here is tested with stats_test.go as what is important
+// is that the final data, the one with send after a call to Export(), is correct.
+
+type groupedStats struct {
+	tgs                  TagSet
+	hitsCount            int64
```
clutchski: these are obviously counts. i think `hits`, `errors`, `duration` works. (also durationCount isn't even a count :) )

model/statsraw.go
```diff
@@ -0,0 +1,228 @@
...
+	"sort"
+
+	"github.com/DataDog/datadog-trace-agent/quantile"
+)
+
+// Most "algorithm" stuff here is tested with stats_test.go as what is important
+// is that the final data, the one with send after a call to Export(), is correct.
+
+type groupedStats struct {
+	tgs                  TagSet
```
clutchski: tags

model/statsraw.go
```diff
@@ -0,0 +1,228 @@
...
+func (sb StatsRawBucket) addSublayer(s Span, aggr string, tgs TagSet, sub SublayerValue) {
+	// This is not as efficient as a "regular" add as we don't update
+	// all sublayers at once (one call for HITS, and another one for ERRORS, DURATION...)
+	// when logically, if we have a sublayer for HITS, we also have one for DURATION,
+	// they should indeed come together. Still room for improvement here.
+
+	var gs groupedStats
+	var ok bool
+
+	subAggr := aggr + "," + sub.Tag.Name + ":" + sub.Tag.Value
```
clutchski: do we need this?

model/statsraw.go
ufoot: I would *really* like to make it private but then `./agent` and `./model` are different packages. We could technically move it and stuff everything in `./agent` but then it defeats modularity (if we look at it, pretty much all the code in this repo is used by the agent only, including Quantizer & Sampler)

model/statsraw.go
ufoot: https://github.com/DataDog/datadog-trace-agent/blob/master/model/sublayers.go#L141 I think so, we really aggregate on different totals depending on this tag. Getting rid of this is a bigger refactoring I think.

clutchski: ok let's punt for now

PR #187, opened 2017-01-13
do not log received traces/spans when there are zero of them

galdor:
This way, we do not fill the logs with useless messages.

### DIFF
--- a/agent/receiver.go
+++ b/agent/receiver.go
@@ -285,7 +285,10 @@ func (r *HTTPReceiver) logStats() {
 		statsd.Client.Count("trace_agent.receiver.span_dropped", sdropped, nil, 1)
 		statsd.Client.Count("trace_agent.receiver.trace_dropped", tdropped, nil, 1)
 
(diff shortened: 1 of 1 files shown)


galdor:
We now keep the message in debug mode.

agent/receiver.go
```diff
@@ -285,7 +285,10 @@ func (r *HTTPReceiver) logStats() {
 		statsd.Client.Count("trace_agent.receiver.span_dropped", sdropped, nil, 1)
 		statsd.Client.Count("trace_agent.receiver.trace_dropped", tdropped, nil, 1)
 
-		log.Infof("receiver handled %d spans, dropped %d ; handled %d traces, dropped %d", spans, sdropped, traces, tdropped)
+		if opts.debug || spans > 0 || sdropped > 0 || traces > 0 || tdropped > 0 {
+			log.Infof("receiver handled %d spans, dropped %d ; handled %d traces, dropped %d",
```
talwai: this is kind of convoluted. why can't we just drop the `if` clause and entirely and `log.Debugf` this?

galdor: Because we may want to have the numbers in non-debug mode ?

awang:
completely addresses my point

PR #188, opened 2017-01-13
change the redis resource format

galdor:
We remove the '*' mark for duplicate commands in a pipeline.

Before:
BenchmarkTestRedisQuantizer-4  1000000  2228 ns/op  1012 B/op  8 allocs/op

After:
BenchmarkTestRedisQuantizer-4  1000000  2045 ns/op   996 B/op  7 allocs/op

### DIFF
--- a/quantizer/redis_test.go
+++ b/quantizer/redis_test.go
@@ -46,10 +46,10 @@ func TestRedisQuantizer(t *testing.T) {
 			"MSET"},
 
 		{"MULTI\nSET k1 v1\nSET k2 v2\nSET k3 v3\nSET k4 v4\nDEL to_del\nEXEC",
-			"MULTI SET* DEL ..."},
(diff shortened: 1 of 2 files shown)


quantizer/redis_test.go
```diff
@@ -73,10 +73,10 @@ func TestRedisQuantizer(t *testing.T) {
 			"GET ..."},
 
 		{"GET k1\nDE...\nGET k2",
```
LotharSee: Out of curiosity, are these tests making sense?
Is it possible to receive weirdly-truncated resources like these? `GET k...\nDEL k2\nMS...`, `GET k1\nDE...\nGET k2`, ...
I'd prefer tests to focus on real type of data, instead of testing impossible edge cases.

galdor: Both lines and pipelines can get truncated by the sender.

PR #190, opened 2017-01-19
[decoder] graceful accept uint64 values if they're sent as int64 for ids

palazzem:
### What it does

In JRuby interpreter, unsigned int are sent as ``int`` type (Java related type system). This patch will handle IDs so that they're cast from ``int`` to ``uint``.

### DIFF
--- a/model/decoder.go
+++ b/model/decoder.go
@@ -120,6 +120,36 @@ func parseInt64(dc *msgp.Reader) (int64, error) {
 	}
 }
 
+// parseUint64 parses an uint64 even if the sent value is an int64;
+// this is required because the language used for the encoding library
+// may not have unsigned types. An example is early version of Java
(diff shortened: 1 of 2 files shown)


mzaccari:
Thank you putting this together. I built the latest version of the trace agent with this code change to see if it resolved the issue for our JRuby 9.1.5.0 application. The error in the `trace-agent.log` was (pre merging this):

```
2017-02-22 12:56:45 ERROR (receiver_logger.go:21) - error when decoding msgpack traces
```

With this change I now see:

```
2017-02-22 12:45:00 ERROR (receiver_logger.go:23) - cannot decode v0.3 traces payload: msgp: attempted to decode type "int" with method for "uint"
```

I'm attempting to debug the issue locally, I'll let you know if I'm able to resolve it. In the mean time do you have any suggestions as to why the new error is being thrown?

PR #191, opened 2017-01-19
fix panic when exiting

galdor:
3ee25af22bb2597cd505c2ed389664eb1e9949ed introduced a bug by not exiting the main loop after closing the
exit chan. Therefore the program panics when close is called a second
time.

We also use (*Writer).Stop() since it exists.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -90,8 +90,9 @@ func (a *Agent) Run() {
 		case <-a.exit:
 			log.Info("exiting")
 			close(a.Receiver.exit)
-			close(a.Writer.exit)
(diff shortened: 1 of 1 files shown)


LotharSee:
Good catch! In fact this commit goes with 9ef64aa2ff1684c53c26a84d495e4327a01adf94 (so the change was https://github.com/DataDog/datadog-trace-agent/compare/c9d4d9de14fe1c262b1d02b3ec16fe82d8d89c2c...3ee25af22bb2597cd505c2ed389664eb1e9949ed ). And the `return` got lost :/

Thanks for the fix :+1:

PR #193, opened 2017-01-24
[profile] using BestSpeed GZIP compression

ufoot:
Using fastest compression saves about 10% CPU cycles but OTOH produces 10% more data. Obviously depends on the data.

### DIFF
--- a/agent/endpoint_test.go
+++ b/agent/endpoint_test.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/DataDog/datadog-trace-agent/fixtures"
 	"github.com/DataDog/datadog-trace-agent/model"
-	// "github.com/stretchr/testify/assert"
 )
 
 func newBenchPayload(traces, spans, stats int) model.AgentPayload {
@@ -33,13 +32,11 @@ func newBenchPayload(traces, spans, stats int) model.AgentPayload {
 
 func BenchmarkEncodeAgentPayload(b *testing.B) {
 	payload := newBenchPayload(10, 1000, 100)
-	var err error
 
 	b.ResetTimer()
 	b.ReportAllocs()
 	for i := 0; i < b.N; i++ {
-		_, err = model.EncodeAgentPayload(payload)
-		if err != nil {
+		if _, err := model.EncodeAgentPayload(payload); err != nil {
 			b.Fatalf("error encoding payload: %v", b)
 		}
 	}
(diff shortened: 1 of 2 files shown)


agent/endpoint_test.go
```diff
@@ -0,0 +1,46 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/DataDog/datadog-trace-agent/fixtures"
+	"github.com/DataDog/datadog-trace-agent/model"
+	// "github.com/stretchr/testify/assert"
```
palazzem: we can remove that

model/payload.go
```diff
@@ -47,7 +47,10 @@ func EncodeAgentPayload(p AgentPayload) ([]byte, error) {
 
 	switch GlobalAgentPayloadVersion {
 	case AgentPayloadV01:
-		gz := gzip.NewWriter(&b)
+		gz, err := gzip.NewWriterLevel(&b, gzip.BestSpeed)
```
palazzem: I'm not sure about that. What would be great, is having numbers about how much the payload is bigger, related to some of our high traffic nodes (so big traces etc...).

model/payload.go
```diff
@@ -47,7 +47,10 @@ func EncodeAgentPayload(p AgentPayload) ([]byte, error) {
...
+		if err != nil {
+			return []byte{}, err
```
galdor: You could just return `nil` instead of `[]byte{}`.

agent/endpoint_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+func BenchmarkEncodeAgentPayload(b *testing.B) {
+	payload := newBenchPayload(10, 1000, 100)
+	var err error
+
+	b.ResetTimer()
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		_, err = model.EncodeAgentPayload(payload)
+		if err != nil {
```
galdor: nitpick: you don't need to declare `err` outside the loop, you can just use:

```go
if _, err := model.EncodeAgentPayload(payload); err != nil {
	b.Fatalf("error encoding payload: %v", b)
}
```

model/payload.go
ufoot: It's sort of tricky to get the exact numbers in real usage as the streaming API does not ease up compression ratio calculation, without modifying the code. OTOH compressing the input data (composed of pretty much the same elements, ids & resources) I get always the same kind of result. About 10% faster, yet 10% bigger. No hard proof yet, as it really depends on the data.

PR #194, opened 2017-01-24
receiver improvements

galdor:


agent/receiver.go
```diff
@@ -266,7 +266,7 @@ func (r *HTTPReceiver) handleServices(v APIVersion, w http.ResponseWriter, req *
 
 // logStats periodically submits stats about the receiver to statsd
 func (r *HTTPReceiver) logStats() {
-	for range time.Tick(60 * time.Second) {
+	for range time.Tick(10 * time.Second) {
```
ufoot: Not sure we really want to be more verbose here, how is this related to the new way of dropping spans?

LeoCavaille: can we keep the log at 60 and the rest at 10?

agent/agent.go
```diff
@@ -111,6 +114,8 @@ func (a *Agent) Process(t model.Trace) {
 
 	if root.End() < model.Now()-2*a.conf.BucketInterval.Nanoseconds() {
 		log.Debugf("skipping trace with root too far in past, root:%v", *root)
+		atomic.AddInt64(&a.Receiver.stats.TracesDropped, 1)
+		atomic.AddInt64(&a.Receiver.stats.SpansDropped, int64(len(t)))
 		return
 	}
```
LeoCavaille: Can we do this easy check before computing sublayers which is expensive?

model/normalizer.go
```diff
@@ -143,41 +143,29 @@ func (s *Span) Normalize() error {
...
+		if s.TraceID != traceID {
+			return t, errors.New("trace id mismatch")
 		}
-		id = s.TraceID
 
-		err := t[i].Normalize()
-		if err != nil {
-			toRemove = append(toRemove, i)
+		if err := t[i].Normalize(); err != nil {
+			return t, fmt.Errorf("invalid span %d: %v", s.SpanID, err)
```
LeoCavaille: maybe print the whole span as %v for more info (service/name/...)?

agent/agent.go
```diff
@@ -111,6 +114,8 @@ func (a *Agent) Process(t model.Trace) {
 
 	if root.End() < model.Now()-2*a.conf.BucketInterval.Nanoseconds() {
```
ufoot: Maybe move this a step above, thinking of it, now that this test is quite simple and quick, maybe have it just as a first shot before we even start processing any sublayer stuff (which is quite time-consuming).

model/normalizer_test.go
```diff
@@ -229,3 +229,47 @@ func TestSpecialZipkinRootSpan(t *testing.T) {
...
+	span1 := testSpan()
+	span1.TraceID = 1
+
+	span2 := testSpan()
+	span2.TraceID = 2
+
+	trace := Trace{span1, span2}
+
+	_, err := NormalizeTrace(trace)
+	assert.NotNil(t, err)
```
ufoot: Cool, can't believe this one was not already here, thanks!

agent/receiver.go
galdor: Yep, I'm going to modify it to log every 60 seconds but update stats every 10 seconds.

agent/agent.go
galdor: Leo made the same comment :)

galdor:
I need to fix the stat log line.

PR #195, opened 2017-01-24
[ci] run go test -bench on Circle CI

ufoot:
Simply run our aleady written go benchmarks and store the artifacts on CircleCI

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -44,6 +44,16 @@ task :test do
   PACKAGES.each { |pkg| go_test(pkg) }
 end
 
+desc "Bench Datadog Trace agent"
+task :bench do
+  sh "install -d ./gobench"
+  rev = `git rev-parse --short HEAD`.strip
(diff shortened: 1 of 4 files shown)


Rakefile
```diff
@@ -44,6 +44,16 @@ task :test do
   PACKAGES.each { |pkg| go_test(pkg) }
 end
 
+desc "Bench Datadog Trace agent"
+task :bench do
+  sh "install -d ./gobench"
+  rev = `git rev-parse --short HEAD`.gsub(/[\n\s]+/, '')
```
galdor: Instead of `gsub(/[\n\s]+/, '')`, you can just use `strip()`.

galdor:
Why don't just squash the last two patches ? Since the typo was introduced in the first patch, you may as well modify it.

ufoot:
Probably better on Gitlab anyway.

PR #196, opened 2017-01-25
fix signal handling to avoid closing the exit chan twice

galdor:
- It's possible to receive more than one SIGINT/SIGTERM during the
shutdown process. So we return after closing the exit chan to avoid
panicking when closing it a second time.
- Only request notification for SIGINT/SIGTERM since these are the only
ones we handle.

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -22,12 +22,15 @@ import (
 // handleSignal closes a channel to exit cleanly from routines
 func handleSignal(exit chan struct{}) {
 	sigChan := make(chan os.Signal, 10)
-	signal.Notify(sigChan)
-	for signal := range sigChan {
-		switch signal {
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+	for signo := range sigChan {
+		switch signo {
 		case syscall.SIGINT, syscall.SIGTERM:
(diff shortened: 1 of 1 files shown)


agent/main.go
```diff
@@ -22,12 +22,15 @@ import (
...
 	sigChan := make(chan os.Signal, 10)
-	signal.Notify(sigChan)
-	for signal := range sigChan {
-		switch signal {
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+	for signo := range sigChan {
+		switch signo {
 		case syscall.SIGINT, syscall.SIGTERM:
-			log.Info("received interruption signal")
+			log.Infof("received signal %d (%v)", signo, signo)
```
ufoot: Since we already display signo as a `%v`, not sure displaying it too as a `%d` really adds something.

galdor: `%v` displays the string representation of the signal; `%d` displays the signal number. For example, for SIGINT, you get `interrupt` and `2`. I would prefer Go to use  the standard signal name `sigint` instead of `interrupt`, but since it does not, I like having the signal number to avoid having to look at the Go sources to figure out what the signal actually is.

PR #197, opened 2017-01-25
add a :coverage rake task

galdor:
It generates a coverage file for all tests on all packages.


LeoCavaille:
Running coverage tests is great, but without tracking the coverage over time (and regressions) it's pretty much pointless I think. If people don't get a red flag when running the CI they keep moving on.

PR #199, opened 2017-01-25
add more tag tests

galdor:


model/tags_test.go
```diff
@@ -102,16 +119,29 @@ func TestFilterTags(t *testing.T) {
...
+func TestTagSetUnset(t *testing.T) {
 	assert := assert.New(t)
-	ts := NewTagSetFromString("service:mcnulty,resource:template,name:magicfunc")
-	ts2 := ts.Unset("name")
-	assert.Len(ts, 3)
-	assert.Equal("mcnulty", ts.Get("service").Value)
-	assert.Equal("template", ts.Get("resource").Value)
-	assert.Equal("magicfunc", ts.Get("name").Value)
+
+	ts1 := NewTagSetFromString("service:mcnulty,resource:template,name:magicfunc")
```
ufoot: Nitpick, I know you did not change this one and the `name:magicfunc` was there before but I would advocate replacing it by `custom:mymetadata` or something else as using `name` as a tag is a legacy behavior, we now use `span.Name`, typically. Putting `resource`, `service` and `name` together as tags is misleading. But otherwise, the test is very fine, and nice to have.

model/tags_test.go
```diff
@@ -53,12 +53,29 @@ func TestTagMerge(t *testing.T) {
...
+	group, value := SplitTag("a:b")
+	assert.Equal(t, group, "a")
+	assert.Equal(t, value, "b")
+
+	group, value = SplitTag("k:v:w")
+	assert.Equal(t, group, "k")
+	assert.Equal(t, value, "v:w")
+
+	group, value = SplitTag("a")
+	assert.Equal(t, group, "")
```
ufoot: :+1: it's nice to have this corner case self-documented by a test.

PR #200, opened 2017-01-25
config: exit 0 if Datadog APM is not enabled explicitly

talwai:
This makes the trace-agent an opt-in process, and is important for hosts
that see an unattended upgrade to datadog-agent@5.11.0 (which will be several)
but don't actually intend to submit to the trace-agent

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -81,7 +81,6 @@ func main() {
 	// command-line arguments
 	flag.StringVar(&opts.ddConfigFile, "ddconfig", "/etc/dd-agent/datadog.conf", "Classic agent config file location")
 	// FIXME: merge all APM configuration into dd-agent/datadog.conf and deprecate the below flag
-	flag.StringVar(&opts.configFile, "config", "/etc/datadog/trace-agent.ini", "Trace agent ini config file.")
 	flag.BoolVar(&opts.debug, "debug", false, "Turn on debug mode")
 	flag.BoolVar(&opts.version, "version", false, "Show version information and exit")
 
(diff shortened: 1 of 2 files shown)


agent/main.go
```diff
@@ -119,6 +119,11 @@ func main() {
 		panic(err)
 	}
 
+	// Exit if tracing is not enabled
+	if !agentConf.Enabled {
+		os.Exit(0)
```
talwai: The behavior of `os.Exit` is that it does not execute any `defer`ed function calls prior to the goroutine exiting. This is _probably_ fine since we don't defer anything explicitly before this point

agent/main.go
```diff
@@ -119,6 +119,15 @@ func main() {
 		panic(err)
 	}
 
+	// Exit if tracing is not enabled
+	if !agentConf.Enabled {
+		// a sleep is necessary to ensure that supervisor registers this process as "STARTED"
+		// If the exit is "too quick", we enter a BACKOFF->FATAL loop even though this is an expected exit
+		// http://supervisord.org/subprocess.html#process-states
+		time.Sleep(5 * time.Second)
+		os.Exit(0)
```
talwai: The behavior of `os.Exit` is that it pre-empts the execution of any `defer`ed calls in this goroutine. This is probably fine since we don't explicitly `defer` anything before this point, but I'm curious to hear views on this.

clutchski: you could just return no?

clutchski: key thing is that it doesn't open ports, write to files, etc.

PR #202, opened 2017-01-30
log errors when loading configuration files

galdor:
- Do not wrap the error when loading fails, so that we can test it.
- Log loading errors unless it's an ENOENT error.
- Log the fact that we are using a configuration file.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,9 +1,10 @@
 package config
 
 import (
-	// stdlib
 	"fmt"
+	"os"
 	"strings"
+	"syscall"
 
 	"gopkg.in/ini.v1"
 )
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -23,12 +24,24 @@ type File struct {
...
-		return nil, fmt.Errorf("unable to read config file at %v, err=%v", configPath, err)
+		return nil, err
 	}
 	globalConfig = &File{instance: config, Path: configPath}
 	return globalConfig, nil
 }
 
+// NewIfExists works as New, but does not return an error if the file does not
+// exist. Instead, it returns a null File pointer.
+func NewIfExists(configPath string) (*File, error) {
```
talwai: possible to add a small test for this behavior?

talwai:
just the 1 comment

PR #206, opened 2017-01-31
[decoder] big buffers are discarded

palazzem:
### What it does

Prevent to keep really big buffers if clients are sending too much data. Currently we're dropping buffers bigger than 10MB

#### To bear in mind

* if the clients are sending the same huge payload (so the average is bigger than 10MB), the decoder will be created over and over increasing the CPU usage.

### DIFF
--- a/model/client_api.go
+++ b/model/client_api.go
@@ -12,7 +12,10 @@ import (
 // average size of a buffer; because usually the payloads are huge,
 // this value ensures that initialized buffers are big enough so that
 // the resize operation is not usually called while reading the response.
-const minBufferSize = 512
+const (
+	minBufferSize = 512
+	maxBufferSize = 10485760
+)
 
 // readAll reads from source until an error or EOF and writes to dest;
 // if the dest buffer contains data, it is truncated
@@ -42,6 +45,7 @@ type ClientDecoder interface {
 	Decode(body io.Reader, v interface{}) error
 	BufferReader() *bytes.Reader
 	ContentType() string
+	Cap() int
 }
 
 type jsonDecoder struct {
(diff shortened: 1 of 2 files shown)


model/client_api.go
```diff
@@ -12,7 +12,10 @@ import (
 // average size of a buffer; because usually the payloads are huge,
 // this value ensures that initialized buffers are big enough so that
 // the resize operation is not usually called while reading the response.
-const minBufferSize = 512
+const (
+	minBufferSize = 512
+	maxBufferSize = 10485760
```
ufoot: I'd probably prefer `1e7` here because I find it quicker to find out this is 10 megs (starting with `1048...` I tend to infer it's `2^20` when it's not. But this is a detail, 10485760 is fine, really)

palazzem: yea actually I didn't think about that... just a quick fix to deploy and see the real effect. We can make it better later

model/client_api.go
```diff
@@ -179,6 +191,11 @@ func (p *DecoderPool) Borrow(contentType string) ClientDecoder {
 // limit is reached, the decoder is discarded so that the call is not blocking.
 // This operation is thread-safe since channels are used.
 func (p *DecoderPool) Release(dec ClientDecoder) {
+	// dropping the decoder if it reaches the maxBufferSize
+	if dec.Cap() > maxBufferSize {
```
clutchski: maybe we could also add a `maximum used times condition` i.e. if it's been borrowed 1000 times always kill it.

palazzem: Agree, I think that we should add as many conditions as we need to let the GC do the job when, at some point, is the time without implementing something too custom

PR #207, opened 2017-01-31
change the listening port

galdor:


agent/receiver.go
```diff
@@ -88,26 +92,38 @@ func (r *HTTPReceiver) Run() {
 	http.HandleFunc("/v0.3/services", httpHandleWithVersion(v03, r.handleServices))
 
 	addr := fmt.Sprintf("%s:%d", r.conf.ReceiverHost, r.conf.ReceiverPort)
-	log.Infof("listening for traces at http://%s/", addr)
+	r.Listen(addr)
+
+	legacyAddr := fmt.Sprintf("%s:%d", r.conf.ReceiverHost, legacyReceiverPort)
+	r.Listen(legacyAddr)
```
ufoot: As I understand it, if 7777 is already binded, here, the agent won't start, and the user has no real way to change the port and/or discard it. Technically, a customer installing this version of the agent for the first time would need to clear port 7777 whatever happens, while we would probably not use it for real. Maybe only log an error in that case? OTOH such an agent could silently run along with an old deprecated one.

talwai:
i'd like us to emit a metric or log something when we see requests come in on the legacy port. this gives us some idea of when we can yank the legacy listener altogether.

galdor:
I can add a metric for this if you wish, but I'm not sure it's that useful. As long as the port 7777 is available, those who use it will continue to use it. If you want to make people use the new port, you have to actively do it (i.e. mark it as deprecated for the next stable version, then remove it for the one after that).

PR #208, opened 2017-01-31
add a /status route

galdor:


LeoCavaille:
Should we close that in favor of #214 ?

PR #209, opened 2017-01-31
check the request method in http handlers

galdor:


talwai:
this change breaks the python client because it does a `PUT` instead of a `POST`. 
https://github.com/DataDog/dd-trace-py/blob/master/ddtrace/api.py#L47

PR #214, opened 2017-02-01
[status/info] add status info

ufoot:
This patch does two things:
1. expose program information with expvar (`GET http://127.0.0.1:<agent-port>/debug/vars` returns a JSON)
2. call this, parse the data, and display them on stdout when calling the program with the info switch (eg `trace-agent -info`)


agent/info.go
```diff
@@ -0,0 +1,58 @@
...
+	infoConfig        config.AgentConfig
+	start             time.Time
+)
+
+func init() {
+	start = time.Now()
+}
+
+func publishUptime() interface{} {
+	return int(time.Now().Sub(start) / time.Second)
```
galdor: You can use `Since` and `Seconds`:

```go
return int(time.Since(start).Seconds())
```

agent/info.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+var (
+	infoMu            sync.RWMutex
+	infoReceiverStats ReceiverStats // only for the last minute
+	infoConfig        config.AgentConfig
+	start             time.Time
+)
+
+func init() {
```
galdor: Careful with the naming, this is a very generic function name and it's in `main`.

agent/info.go
```diff
@@ -29,30 +35,138 @@ func updateReceiverStats(rs ReceiverStats) {
...
+	}
+	infoJsonConfig = string(buf)
+	expvar.Publish("config", infoConfig{})
+	return nil
+}
+
+type infoVersion struct{}
+
+// String implements the expvar.Var interface
+func (infoVersion) String() string { return `"` + Version + `"` }
```
galdor: When you want to quote a string, juste use `strconv.Quote`. This way, it will add the double quotes, and will escape those in the string:

```go
return strconv.Quote(Version)
```

agent/info.go
```diff
@@ -29,30 +35,138 @@ func updateReceiverStats(rs ReceiverStats) {
...
+
 func init() {
 	expvar.NewInt("pid").Set(int64(os.Getpid()))
 	expvar.Publish("uptime", expvar.Func(publishUptime))
+	version := expvar.NewMap("version")
+	version.Set("Version", infoVersion{})
+	version.Set("GitCommit", infoGitCommit{})
+	version.Set("GitBranch", infoGitBranch{})
+	version.Set("BuildDate", infoBuildDate{})
+	version.Set("GoVersion", infoGoVersion{})
```
galdor: Generally speaking, `Set` expects an object implementing `String()`. Instead of using a different type for each value, why not just use for example:

```go
type infoString string
func (s infoString) string() string { return string(s) }
```

Then:

```go
version.Set("Version", infoString(Version))
```

agent/info.go
```diff
@@ -29,30 +35,138 @@ func updateReceiverStats(rs ReceiverStats) {
...
+	} `json:"version"`
+	Receiver ReceiverStats      `json:"receiver"`
+	Config   config.AgentConfig `json:"config"`
+}
+
+// Info returns a printable string, suitable for the `-info` option.
+func Info(conf *config.AgentConfig) (string, error) {
+	host := conf.ReceiverHost
+	if host == "0.0.0.0" {
+		host = "127.0.0.1" // [FIXME:christian] not fool-proof
```
galdor: Why not just print `0.0.0.0` ?

agent/info.go
```diff
@@ -29,30 +35,138 @@ func updateReceiverStats(rs ReceiverStats) {
...
+		return "", fmt.Errorf("unable to contact Datadog Trace Agent on '%s', not running\nERROR: %v\n", url, err)
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", fmt.Errorf("unable to read response from Datadog Trace Agent on '%s'\nERROR: %v\n", url, err)
+	}
+
+	var info StatusInfo
+	err = json.Unmarshal(body, &info)
```
galdor: As always, it's easier to stream decode:

```go
if err := json.NewDecoder(resp.Body).Decode(&info); err != nil {
	// TODO
}
```

agent/info.go
```diff
@@ -29,30 +35,138 @@ func updateReceiverStats(rs ReceiverStats) {
...
+		"  API Endpoints: " + strings.Join(info.Config.APIEndpoints, ", ") + "\n" +
+		"  Receiver Host: " + info.Config.ReceiverHost + "\n" +
+		"  Receiver port: " + strconv.Itoa(info.Config.ReceiverPort) + "\n" +
+		"  Mem alloc: " + fmt.Sprintf("%d", info.MemStats.Alloc) + "\n" +
+		"\n" +
+		"  Spans received (1 min): " + strconv.Itoa(int(info.Receiver.SpansReceived)) + "\n" +
+		"  Traces received (1 min): " + strconv.Itoa(int(info.Receiver.TracesReceived)) + "\n" +
+		"  Spans dropped (1 min): " + strconv.Itoa(int(info.Receiver.SpansDropped)) + "\n" +
+		"  Traces dropped (1 min): " + strconv.Itoa(int(info.Receiver.TracesDropped)) + "\n" +
+		"\n"), nil
```
galdor: For this code, a `text/template` would have been handy.

agent/info.go
ufoot: 👍

agent/info.go
ufoot: This is used not to display, but to connect to it, `GET http://0.0.0.0:7777` won't do the job.

agent/info.go
```diff
@@ -0,0 +1,216 @@
...
+		}{
+			Banner:       banner,
+			Program:      program,
+			ReceiverPort: conf.ReceiverPort,
+		})
+		return false, err
+	}
+
+	var info StatusInfo
+	if err := json.NewDecoder(resp.Body).Decode(&info); err != nil {
```
LeoCavaille: nit but you're responsible for closing the `resp.Body`

agent/info.go
```diff
@@ -0,0 +1,216 @@
...
+	err = infoTmpl.Execute(w, struct {
+		Banner       string
+		Program      string
+		CmdLine      string
+		APIEndpoints string
+		Status       *StatusInfo
+	}{
+		Banner:       banner,
+		Program:      program,
+		CmdLine:      strings.Join(info.CmdLine, " "), // [FIXME:christian] find a way to do this in text/template
```
LeoCavaille: You can do ranges in templates IIRC.

PR #215, opened 2017-02-01
limit the length of request bodies

galdor:


agent/receiver.go
```diff
@@ -70,22 +66,24 @@ func NewHTTPReceiver(conf *config.AgentConfig) *HTTPReceiver {
 		conf:        conf,
 		logger:      &errorLogger{},
 		exit:        make(chan struct{}),
+
+		maxRequestBodyLength: 10 * 1024 * 1024,
```
LeoCavaille: Can we change that to a constant?

model/limited_reader.go
```diff
@@ -0,0 +1,48 @@
...
+
+// LimitedReader reads from a reader up to a specific limit. When this limit
+// has been reached, any subsequent read will return
+// ErrLimitedReaderLimitReached.
+// The underlying reader has to implement io.ReadCloser so that it can be used
+// with http request bodies.
+type LimitedReader struct {
+	r     io.ReadCloser
+	limit int64
+}
```
LeoCavaille: https://golang.org/pkg/io/#LimitedReader :) ?

galdor: Read the commit message ;)

LeoCavaille: That's the type of things that should be in the PR description (right now there is none), when there are multiple commits I usually end up concatenating the commit messages in there (when you have a single commit it does it by default). Cool.

PR #216, opened 2017-02-01
do not truncate when logging why a trace could not be normalized

galdor:
Also remove the -debug flag as we discussed.

### DIFF
--- a/agent/receiver.go
+++ b/agent/receiver.go
@@ -55,6 +55,7 @@ type HTTPReceiver struct {
 	exit chan struct{}
 
 	maxRequestBodyLength int64
+	debug                bool
 }
 
 // NewHTTPReceiver returns a pointer to a new HTTPReceiver
@@ -69,6 +70,7 @@ func NewHTTPReceiver(conf *config.AgentConfig) *HTTPReceiver {
 		exit:        make(chan struct{}),
 
 		maxRequestBodyLength: maxRequestBodyLength,
+		debug:                strings.ToLower(conf.LogLevel) == "debug",
 	}
 }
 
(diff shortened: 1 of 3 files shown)


agent/receiver.go
```diff
@@ -200,13 +200,7 @@ func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *ht
 		if err != nil {
 			atomic.AddInt64(&r.stats.TracesDropped, 1)
 			atomic.AddInt64(&r.stats.SpansDropped, int64(spans))
-
-			// this is a potentially very spammy log message, so extra care
-			errorMsg := fmt.Sprintf("dropping trace reason: %s (debug for more info), %v", err, normTrace)
-			if len(errorMsg) > 150 {
```
talwai: don't we still want to continue truncating except in debug mode? so maybe `if len(errorMsg) > 150 && !agentConf.LogLevel == DEBUG`

galdor: There is zero reason to truncate here. If you see a truncated line and restart the agent in debug, there is no guarantee that you'll be able to reproduce the problem.

talwai: `// this is a potentially very spammy log message, so extra care`
isn't this the reason?
i don't think we can dump every dropped trace in its entirety. let's add the debug clause and run the agent with log_level: debug for the period we need info

galdor: It used to write the entire trace which was indeed very verbose. I removed it since the errors indicate precisely what is wrong.

LeoCavaille: I agree with Aaditya, printing an error without a hint of the data that is wrong is probably useless. If your agent is sending 1000 traces/s, how can you know which span failed to normalize because of "missing service"?
I think your reasoning of "no guarantee of being able to reproduce the problem" is also valid for logging an error only.

Keeping the previous behavior but just logging the full message without truncation in debug mode allows you to spot what kind of data causes issues and gives you a chance to reproduce and get more info when setting the agent in debug mode.

galdor: Ok that makes sense, I'll make the patch right away.

agent/receiver.go
```diff
@@ -201,9 +201,8 @@ func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *ht
 			atomic.AddInt64(&r.stats.TracesDropped, 1)
 			atomic.AddInt64(&r.stats.SpansDropped, int64(spans))
 
-			// this is a potentially very spammy log message, so extra care
 			errorMsg := fmt.Sprintf("dropping trace reason: %s (debug for more info), %v", err, normTrace)
-			if len(errorMsg) > 150 {
+			if len(errorMsg) > 150 && r.conf.LogLevel != "DEBUG" {
```
LeoCavaille: Just by looking quickly at the code not sure this config attribute is really normalized ? It could be "debug" for instance? Maybe find a more reliable way to check the logger log level?

galdor: There does not seem to be a better way, but I can cache a flag, it will be cleaner.

PR #217, opened 2017-02-01
receiver: set a WriteTimeout, and avoid blocking on a full channel

talwai:
These changes are in response to a situation where we leaked file descriptors on a high-throughput node,
eventually hitting kernel limits and being unable to accept new traces

It's likely that we were unable to close `request.Body` and release our connection, due to blocking
in the main HTTP loop. To avoid this we:

- set a WriteTimeout on the server, defaulting to 5 seconds, and configurable via `receiver_timeout`
- drop traces on the floor when we can't write them to the outbound `traces` channel

### DIFF
--- a/agent/receiver.go
+++ b/agent/receiver.go
@@ -219,7 +219,7 @@ func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *ht
 			select {
 			case r.traces <- normTrace:
 			default:
-			    r.logger.Errorf(fmt.Sprintf("dropping trace reason: rate-limited")
+				r.logger.Errorf("dropping trace reason: rate-limited")
 			}
 		}
 
--- a/agent/receiver.go
+++ b/agent/receiver.go
@@ -102,7 +102,10 @@ func (r *HTTPReceiver) Run() {
 	if r.conf.ReceiverTimeout > 0 {
 		timeout = r.conf.ReceiverTimeout
 	}
-	server := http.Server{ReadTimeout: time.Second * time.Duration(timeout)}
+	server := http.Server{
+		ReadTimeout:  time.Second * time.Duration(timeout),
+		WriteTimeout: time.Second * time.Duration(timeout),
+	}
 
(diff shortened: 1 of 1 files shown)


agent/receiver.go
```diff
@@ -209,7 +212,15 @@ func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *ht
...
 			atomic.AddInt64(&r.stats.SpansDropped, int64(spans-len(normTrace)))
-			r.traces <- normTrace
+
+			// if our downstream consumer is slow, we drop the trace on the floor
+			// this is a safety net against us using too much memory
+			// when clients flood us
+			select {
+			case r.traces <- normTrace:
+			default:
+				r.logger.Errorf("dropping trace reason: rate-limited")
```
ufoot: We might still want to keep stats for this, see https://github.com/DataDog/datadog-trace-agent/pull/217/files#diff-e8961f81006d547c70ab424318805bbaR204 above 
```
atomic.AddInt64(&r.stats.TracesDropped, 1)
atomic.AddInt64(&r.stats.SpansDropped, int64(spans))
```

ufoot: Also: https://github.com/DataDog/datadog-trace-agent/pull/217/files#diff-e8961f81006d547c70ab424318805bbaL64 following some work on our internal servers, I came to the conclusion the right size for a channel is usually "the amount of data you get during a 'period'", with 'period' being a timeout, a flush timeout, whatever... We flush every 10 sec and have a default http timeout of 5 sec so it's reasonnable to say this channel should hold "as many traces we can get during 5 or 10 sec". One of our goals is to fluently handle about 1000 traces/sec so logically, the traces buffer should be 5000 or 10000 (think of those very simple traces with one single span...) else I suspect it's going to drop a lot of stuff when it still has enough power to do the job. Of course this is not very precise and there are places downstream where things could still be buffered, but being conservative, let's assume until it's not flushed, it's still hanging there.

PR #218, opened 2017-02-02
[test] fixed flaky test, making sure server is started

ufoot:


galdor:
I'm not very comfortable with this kind of hack. The problem is that we have no synchronous way to start the agent. But if it fixes the immediate problem, ok.

PR #219, opened 2017-02-02
[status/info] refining the output of trace-agent -info

ufoot:
This patch:
- removes the statsd info
- compacts the receiver host and port into one line
- compacts the receiver stats into one line, isolated in a subsection


ufoot:
Sample output here: https://github.com/DataDog/datadog-trace-agent/blob/c7603728a15a156c33a69277511577e36535bed7/agent/info.go#L177-L214

LotharSee:
Comments on what we could display there:

- command line, API Endpoints, receiver aren't thing that people are supposed to change (and they come straight from the config), so I'm not sure they are that useful
- uptime is useful, but it lacks a unit suffix/pretty formatting
- +1 on hostname and pid 
- +1 on displaying stats ; we might want to display even more? A few ideas:
 - count of payloads in/flushes out, traces/spans in, traces/spans out
 - sizes of payloads in/out?
 - maxTPS/CPU limit (not for now, I need to change the sampler first)
 - last API status code?
- let's also inline them with a simple format and not long lines

ufoot:
- the API Endpoints *can* come from the conf, but defaults are provided, if you don't override them (which we typically do, as a default setup), you have no clue on your filesystem to guess what they are, and pretty much have to read the source or doc to figure out AFAIK
- 👍  for the uptime, yes, adding a unit is a must-have
- about the inlining and stats, I think we should align and synchronize what we do in that status line and what we show in the logs. I mean, if it's worth displaying as live data here, it has a place in the logs. Will check that along those lines.

ufoot:
New status output looks like:

```
======================
Trace Agent (v 0.99.0)
======================

  Command line: ./trace-agent
  Pid: 4642
  Uptime: 61 seconds
  Mem alloc: 3142568 bytes

  Hostname: ubuntu-1204.vagrantup.com
  Receiver: 0.0.0.0:8126
  API Endpoints: http://localhost:8012

  Bytes received (1 min): 57103
  Traces received (1 min): 240
  Spans received (1 min): 360

  Bytes sent (1 min): 75827
  Traces sent (1 min): 1201
  Stats sent (1 min): 4
```

If there are some traces dropped by the receiver or some endpoint errors, it displays them with a `WARNING:`, see https://github.com/DataDog/datadog-trace-agent/pull/219/files#diff-6f43e4800a4f57a19ba9f109cde762a9R225 for details

agent/info.go
```diff
@@ -178,44 +205,60 @@ func getProgramBanner(version string) (string, string) {
...
+//   Traces received (1 min): 240
+//   Spans received (1 min): 360
+//   WARNING: Traces dropped (1 min): 5
+//   WARNING: Spans dropped (1 min): 10
+//
+//   Bytes sent (1 min): 3245
+//   Traces sent (1 min): 6
+//   Stats sent (1 min): 60
+//   WARNING: Traces API errors (1 min): 1/3
+//   WARNING: Services API errors (1 min): 1/1
```
LotharSee: Great idea these warnings, they will be more useful that what we think.

LotharSee:
Good work, it is all good to me! And good catch for the `die()` thing.

ufoot:
Last nitpick -> I removed the command line (indeed, it's not something users are supposed to change)

PR #225, opened 2017-02-06
sql: quantize as aliases.

clutchski:
some of them can be unbounded and don't materially change the contents
of the query. here's the direct case:

    SELECT pg_try_advisory_lock ( ? ) AS tabcdef1235

this could also be quantized to ` SELECT pg_try_advisory_lock ( ? )` but i din't do that.

### DIFF
--- a/quantizer/main.go
+++ b/quantizer/main.go
@@ -2,7 +2,6 @@ package quantizer
 
 import (
 	"regexp"
-	"strings"
 
 	"github.com/DataDog/datadog-trace-agent/model"
 )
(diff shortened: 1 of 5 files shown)


quantizer/main.go
```diff
@@ -25,28 +24,16 @@ type QuantizeFunction func(model.Span) model.Span
 func Quantize(span model.Span) model.Span {
 	switch span.Type {
 	case sqlType:
-		fallthrough
+		return QuantizeSQL(span)
```
clutchski: explicit is better than implicit

palazzem: +1

quantizer/sql.go
```diff
@@ -51,6 +51,8 @@ func (f *ReplaceFilter) Filter(token, lastToken int, buffer []byte) (int, []byte
 		return Filtered, []byte("?")
 	case Limit:
 		return token, buffer
+	case As:
+		return token, []byte("?")
```
palazzem: happy that is quite easy to extend quantizer rules

quantizer/sql_test.go
```diff
@@ -244,6 +244,10 @@ func TestSQLQuantizer(t *testing.T) {
 			"SELECT * FROM public.table ( array [ ROW ( array [ 'magic', 'foo',",
 			"SELECT * FROM public.table ( array [ ROW ( array [ ?",
 		},
+		{
+			"SELECT pg_try_advisory_lock (123) AS t46eef3f025cc27feb31ca5a2d668a09a",
+			"SELECT pg_try_advisory_lock ( ? ) AS ?",
```
palazzem: For now I think it's ok, we may want (as you've already suggested) to drop all tables/store procedures aliasing. We may keep track of this change for the future.

LotharSee:
@palazzem we talked about completely dropping all the `AS ?`, would it be hard to adapt this PR to make it work?

PR #226, opened 2017-02-07
weight spans using their sample rate

galdor:
The root span of a trace can contain a _sample_rate value if the client
sampled spans. When this is the case, we scale stats (hit and error
counts, and durations) by a factor, or weight, equal to `1.0 /
_sample_rate`.

This also means that we have to use floating point values to accumulate
values for `StatsRawBucket`.

### DIFF
--- a/agent/concentrator.go
+++ b/agent/concentrator.go
@@ -45,11 +45,16 @@ func (c *Concentrator) Add(t processedTrace) {
 			c.buckets[btime] = b
 		}
 
+		weight := 1.0
+		if t.Root != nil {
+			weight = t.Root.Weight()
+		}
+
 		if t.Root != nil && s.SpanID == t.Root.SpanID && t.Sublayers != nil {
 			// handle sublayers
-			b.HandleSpan(s, t.Env, c.aggregators, &t.Sublayers)
+			b.HandleSpan(s, t.Env, c.aggregators, weight, &t.Sublayers)
 		} else {
-			b.HandleSpan(s, t.Env, c.aggregators, nil)
+			b.HandleSpan(s, t.Env, c.aggregators, weight, nil)
 		}
 	}
 
--- a/model/span.go
+++ b/model/span.go
@@ -5,6 +5,11 @@ import (
 	"math/rand"
 )
 
+const (
+	// SpanSampleRateMetricKey is the metric key holding the sample rate
+	SpanSampleRateMetricKey = "_sample_rate"
+)
+
 // Span is the common struct we use to represent a dapper-like span
 type Span struct {
 	// Mandatory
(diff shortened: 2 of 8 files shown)


model/span.go
```diff
@@ -59,3 +64,14 @@ func NewFlushMarker() Span {
 func (s *Span) End() int64 {
 	return s.Start + s.Duration
 }
+
+// Weight returns the weigth of the span as defined for sampling, i.e. the
```
LeoCavaille: typo: weight

model/span.go
```diff
@@ -59,3 +64,14 @@ func NewFlushMarker() Span {
...
+// inverse of the sampling rate.
+func (s *Span) Weight() float64 {
+	sampleRate, found := s.Metrics[SpanSampleRateMetricKey]
```
LeoCavaille: nitpick: but for consistency we use `ok` rather than `found` for maps

agent/concentrator.go
```diff
@@ -47,9 +47,9 @@ func (c *Concentrator) Add(t processedTrace) {
 
 		if t.Root != nil && s.SpanID == t.Root.SpanID && t.Sublayers != nil {
 			// handle sublayers
-			b.HandleSpan(s, t.Env, c.aggregators, &t.Sublayers)
+			b.HandleSpan(s, t.Env, c.aggregators, t.Root, &t.Sublayers)
 		} else {
-			b.HandleSpan(s, t.Env, c.aggregators, nil)
+			b.HandleSpan(s, t.Env, c.aggregators, t.Root, nil)
```
LeoCavaille: For the API if we just use the weight of the root span can we consider passing just the weight rather than a pointer to the full span? Would avoid having to check for nil pointers downstream etc.

model/span.go
galdor: Having an element in a map is not "ok" (or not), but as you wish.

agent/concentrator.go
galdor: Sure.

model/statsraw.go
```diff
@@ -12,9 +12,9 @@ import (
 
 type groupedStats struct {
 	tags                 TagSet
-	hits                 int64
-	errors               int64
-	duration             int64
+	hits                 float64
+	errors               float64
+	duration             float64
```
LotharSee: Since we are `ns` here, we can stick to `in64` without any significant lose of precision.

model/statsraw.go
```diff
@@ -12,9 +12,9 @@ import (
 
 type groupedStats struct {
 	tags                 TagSet
-	hits                 int64
-	errors               int64
-	duration             int64
+	hits                 float64
```
LotharSee: Is the move from `int64` to `float64` going to impact anything in the serialization/payload to the API? Is the API compatible?

LeoCavaille: I think this in an internal format only that we then export to the API payload format.

galdor: These counters are for internal computations in `StatsRawBucket`. `(*StatsRawBucket).Export()` returns a `StatsBucket` which already store data using `Count` structures which  use `float64` values.

model/statsraw.go
galdor: Or we can use `float64` without losing precision, since it will be exported to a `float64` value anyway (see next comment).

PR #227, opened 2017-02-08
implement stream decoding

galdor:
See commit messages for more information.

### DIFF
--- a/agent/receiver.go
+++ b/agent/receiver.go
@@ -1,17 +1,21 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"strings"
 	"sync/atomic"
 	"time"
 
+	log "github.com/cihub/seelog"
+	"github.com/tinylib/msgp/msgp"
+
 	"github.com/DataDog/datadog-trace-agent/config"
 	"github.com/DataDog/datadog-trace-agent/model"
 	"github.com/DataDog/datadog-trace-agent/statsd"
-	log "github.com/cihub/seelog"
 )
 
 // The trace agent used to listen on port 7777, but now uses port 8126. Keep
@@ -19,7 +23,6 @@ import (
 const legacyReceiverPort = 7777
 
 const (
-	decoderSize          = 10 // Max size of decoders pool
 	maxRequestBodyLength = 10 * 1024 * 1024
 	tagTraceHandler      = "handler:traces"
 	tagServiceHandler    = "handler:services"
@@ -46,10 +49,9 @@ const (
 // HTTPReceiver is a collector that uses HTTP protocol and just holds
 // a chan where the spans received are sent one by one
 type HTTPReceiver struct {
-	traces      chan model.Trace
-	services    chan model.ServicesMetadata
-	decoderPool *model.DecoderPool
-	conf        *config.AgentConfig
+	traces   chan model.Trace
+	services chan model.ServicesMetadata
+	conf     *config.AgentConfig
 
 	// due to the high volume the receiver handles
 	// custom logger that rate-limits errors and track statistics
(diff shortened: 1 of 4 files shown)


agent/receiver.go
```diff
@@ -254,39 +241,11 @@ func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *ht
...
-				r.logger.Errorf(model.HumanReadableJSONError(dec.BufferReader(), err))
-			} else {
-				r.logger.Errorf("error when decoding msgpack traces")
-			}
-			HTTPDecodingError(err, []string{tagServiceHandler, fmt.Sprintf("v:%s", v)}, w)
-			return
-		}
-	default:
-		HTTPEndpointNotSupported([]string{tagServiceHandler, fmt.Sprintf("v:%s", v)}, w)
+	contentType := req.Header.Get("Content-Type")
```
ufoot: I'm mixed here. One one hand, doing it this way makes the code more compact, and slightly more permissive, but this is fine. OTOH, doing this, we loose the information that "v01 and v02 are JSON only while v03 is JSON & MsgPack and has to decide depending on Content-Type". As I understand it, this code will accept any protocol on any enpoint. Maybe just state with a comment that MsgPack should only be for v03? Because for now, we know it, but we can still "forget it" when we migrate.

galdor:
Small update to add more tests for the LimitedReader/EOF workaround.

model/limited_reader.go
```diff
@@ -38,6 +38,13 @@ func (r *LimitedReader) Read(buf []byte) (n int, err error) {
 	}
 	n, err = r.r.Read(buf)
 
+	// Some libraries (e.g. msgp) will ignore read data if err is not nil.
+	// We reset err if something was read, and the next read will return
+	// io.EOF with no data.
+	if err == io.EOF && n > 0 {
```
ufoot: :+1:

agent/receiver.go
galdor: We filter msgpack payloads for v0.1 and v0.2 in `httpHandleWithVersion`. Since decoding is handled in `decodeReceiverPayload`, handlers don't have to think about the format of the payload. The only exception is the manual JSON decoding for v0.1, but we're going to deprecate v0.1 anyway.

ufoot: OK, my bad, I missed the fact httpHandleWithVersion was doing the filtering, sorry for the noise.

agent/receiver.go
```diff
@@ -157,72 +166,46 @@ func (r *HTTPReceiver) httpHandleWithVersion(v APIVersion, f func(APIVersion, ht
...
 	default:
 		HTTPEndpointNotSupported([]string{tagTraceHandler, fmt.Sprintf("v:%s", v)}, w)
 		return
 	}
 
 	HTTPOK(w)
 
-	if contentLength > 0 {
-		atomic.AddInt64(&r.stats.TracesBytes, int64(contentLength))
+	bytesRead := req.Body.(*model.LimitedReader).Count
```
ufoot: 👍

agent/receiver.go
```diff
@@ -157,72 +166,46 @@ func (r *HTTPReceiver) httpHandleWithVersion(v APIVersion, f func(APIVersion, ht
...
-		if err != nil {
-			r.logger.Errorf(model.HumanReadableJSONError(dec.BufferReader(), err))
-			r.decoderPool.Release(dec)
-			HTTPDecodingError(err, []string{tagTraceHandler, fmt.Sprintf("v:%s", v)}, w)
-			return
-		}
 
-		r.decoderPool.Release(dec)
+	case v02:
+		fallthrough
```
clutchski: nitpick: i don't like fallthrough because it sort of hinders local resoni kind of hate fallthrough because it hinders local reasoning.

why not just `case v01, v02`

agent/receiver.go
```diff
@@ -375,3 +327,20 @@ type receiverStats struct {
 	// SpansReceived is the number of traces dropped
 	TracesDropped int64
 }
+
+func decodeReceiverPayload(r io.Reader, dest msgp.Decodable, v APIVersion, contentType string) error {
+	switch contentType {
+	case "application/msgpack":
+		return msgp.Decode(r, dest)
+
+	case "application/json":
```
clutchski: same comment about fall through

agent/receiver.go
```diff
@@ -46,10 +49,9 @@ const (
 // HTTPReceiver is a collector that uses HTTP protocol and just holds
 // a chan where the spans received are sent one by one
 type HTTPReceiver struct {
-	traces      chan model.Trace
-	services    chan model.ServicesMetadata
-	decoderPool *model.DecoderPool
```
clutchski: this is the second time i've made this comment over the last year, but the decoder pool was added to a fix a performance bottleneck (i think it's so we could re-use the decoding buffers, but it's been a while).

can you verify this has no impact?

PR #228, opened 2017-02-09
Adaptive sampling — Step 1

LotharSee:
Make sampling smarter with some adaptive scoring.

### Main concept change

The 2 main internal scoring parameters (offset and slope) were constants, and they weren't suitable for all distributions of traces.
Now, they are frequently adjusted to better fit the input and the maxTPS limit.

- `offset` is now computed based on the ratio between the number of Traces Per Second we keep and the maxTPS set by the config. Every 5s, we increase/decrease the offset (to sample more or less)  accordingly.
- `slope` is now a function of the input distribution (the number of different signatures and the total TPS are parameters). 

The implementation/equation is done is such a way that it has some interesting properties/protections. I won't explain in details why here, but I plan on writing a real algorithmic explanation of it at some point.

### Small changes

- Smaller `decayPeriod` (from 30s to 5s). It basically makes the sampler counters shorter-term: more reactive but decay/forget faster (half-life from 3 minutes to 30 seconds). 
- Expose internal parameters in a `InternalState` structure.
- Report all these internal parameters as statsd metrics. Not sure we will keep them forever, but at least for now it is very useful to watch over sampler behavior.

### Next step (future PR)

This brings a scoring sampler <-> maxTPS feedback loop.
The next step will be a maxTPS <-> CPU usage feedback loop.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -67,6 +67,7 @@ end
 
 desc "Run Datadog Trace agent"
 task :run do
+  ENV['DD_APM_ENABLED'] = 'true'
   sh "./trace-agent -config ./agent/trace-agent.ini"
 end
 
--- a/sampler/sampler.go
+++ b/sampler/sampler.go
@@ -134,8 +134,10 @@ func (s *Sampler) AdjustScoring() {
 		}
 	} else if TPSratio < 0.8 {
 		// If below, increase the offset
-		// Don't do it if we already keep all traces or if offset above maxTPS
-		if currentTPS < s.Backend.GetTotalScore() && s.signatureScoreOffset < s.maxTPS {
+		// Don't do it if:
+		//  - we already keep all traces (with a 1% margin because of stats imprecision)
+		//  - offset above maxTPS
+		if currentTPS < 0.99*s.Backend.GetTotalScore() && s.signatureScoreOffset < s.maxTPS {
 			coefficient = 1.1
 			if TPSratio < 0.5 {
 				coefficient = 1.3
--- a/sampler/sampler.go
+++ b/sampler/sampler.go
@@ -17,10 +17,7 @@ import (
 	"math"
 	"time"
 
-	log "github.com/cihub/seelog"
-
 	"github.com/DataDog/datadog-trace-agent/model"
-	"github.com/DataDog/datadog-trace-agent/statsd"
 )
 
 const (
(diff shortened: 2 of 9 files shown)


sampler/sampler.go
```diff
@@ -78,12 +88,82 @@ func (s *Sampler) UpdateMaxTPS(maxTPS float64) {
...
+		}
+	} else if TPSratio < 0.8 {
+		// If below, increase the offset
+		// Don't do it if:
+		//  - we already keep all traces (with a 1% margin because of stats imprecision)
+		//  - offset above maxTPS
+		if currentTPS < 0.99*totalTPS && s.signatureScoreOffset < s.maxTPS {
+			coefficient = 1.1
+			if TPSratio < 0.5 {
+				coefficient = 1.3
```
ufoot: This looks like it's crafted/fine-tuned from some real use case, I'm wondering if we should make all these numbers dynamically tunable (conf?). I feel once we've run this for weeks and month it's really fine to hardcode them. In between, our mileage may vary. Depends on how sure you are about these, if they are already here to stay, good to go, but if we still need to tweak them...

sampler/sampler.go
```diff
@@ -78,12 +88,82 @@ func (s *Sampler) UpdateMaxTPS(maxTPS float64) {
...
+
+	// Safeguard to avoid too small offset (for guaranteed very-low volume sampling)
+	if newOffset < minSignatureScoreOffset {
+		newOffset = minSignatureScoreOffset
+	}
+
+	// Default slope value
+	newSlope := defaultSignatureScoreSlope
+
+	// TODO: explain why this formula
```
ufoot: Indeed ;)

sampler/sampler.go
```diff
@@ -78,12 +88,82 @@ func (s *Sampler) UpdateMaxTPS(maxTPS float64) {
...
+}
+
+// AdjustScoring modifies sampler coefficients to fit better the `maxTPS` condition
+func (s *Sampler) AdjustScoring() {
+	// See how far we are from our maxTPS limit and make signature sampler harder/softer accordingly
+	currentTPS := s.Backend.GetSampledScore()
+	totalTPS := s.Backend.GetTotalScore()
+	TPSratio := currentTPS / s.maxTPS
+	offset := s.signatureScoreOffset
+	cardinality := float64(s.Backend.GetCardinality())
```
ufoot: Once we reach this point, I believe the rest of the function is pretty much "stateless". Pattern is "get info from s, do things, then update s". I think you could gain in testability (if not readability) by pulling this logic out of the method and make it a pure stateless method which takes as an input, say `currentTPS`, `totalTPS`, `TPSratio` and `cardinality` and then outputs `newOffset` and `newSlope`. This way writing unit tests would be much easier and could partially achieve documentation goals. Just my 2c.

Rakefile
```diff
@@ -67,6 +67,7 @@ end
 
 desc "Run Datadog Trace agent"
 task :run do
+  ENV['DD_APM_ENABLED'] = 'true'
```
ufoot: Much needed!

PR #234, opened 2017-02-21
end date validation

galdor:


model/normalizer.go
```diff
@@ -94,6 +97,11 @@ func (s *Span) Normalize() error {
 		return fmt.Errorf("span.normalize: invalid `Start` (must be nanosecond epoch): %d", s.Start)
 	}
 
+	// If the end date is too far away in the future, it's probably a mistake.
+	if s.Start+s.Duration > time.Now().Add(MaxEndDateOffset).UnixNano() {
```
clutchski: do we need a `UTC` in here just in case?

model/normalizer.go
```diff
@@ -94,6 +97,11 @@ func (s *Span) Normalize() error {
...
+		return fmt.Errorf("span.normalize: more than %v in the future", MaxEndDateOffset)
```
clutchski: how does this error get handled on the server side? want to make sure we increment a stat or log something, but it shouldn't be an error level log. Error should be for things that require a dev's attention.

model/normalizer.go
galdor: The documentation states that `UnixNano returns t as a Unix time, the number of nanoseconds elapsed since January 1, 1970 UTC.

Running a test program printing both `time.Now().UnixNano()` and `time.Now().UTC().UnixNano()` yields two (nearly) identical values, as expected.

model/normalizer.go
galdor: In the agent receiver, if normalization fails, the trace is dropped. This increments two metrics, one for dropped traces, and one for dropped spans, and an error message is logged. Do you want me to only log it in debug ?

PR #235, opened 2017-02-22
[build] fixed possible cycle import issue in build (does not show on ci)

ufoot:
This problem does not show in CircleCI as it checkouts the code in special directories, but running `rake ci` on a standard dev host, with the code in `$GOPATH/src/github.com/DataDog/datadog-trace-agent` the test on `./model` fails, as it imports config, which imports model... Fixed this by disabling the log using log functions directly, and not relying on the helpers in config.

### DIFF
--- a/model/client_api_test.go
+++ b/model/client_api_test.go
@@ -11,7 +11,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


LeoCavaille:
Closing as this has been fixed in 92c92a5be25ba3c46855a37eae30375196e94c36

PR #236, opened 2017-02-22
switch to glide

galdor:


GLOCKFILE
```diff
@@ -1,8 +0,0 @@
-cmd github.com/golang/lint/golint
-github.com/DataDog/datadog-go a9c7a9896c1847c9cc2b068a2ae68e9d74540a5d
-github.com/cihub/seelog d2c6e5aa9fbfdd1c624e140287063c7730654115
-golang.org/x/tools 344080534465c5613c3563c30fd38c031073841d
-github.com/golang/lint 32a87160691b3c96046c0c678fe57c5bef761456
-github.com/stretchr/testify 8ce79b9f0b77745113f82c17d0756771456ccbd3
-github.com/tinylib/msgp 362bfb3384d53ae4d5dd745983a4d70b6d23628c
-gopkg.in/ini.v1 12f418cc7edc5a618a51407b7ac1f1f512139df3
```
LeoCavaille: :wave: glock

Rakefile
```diff
@@ -2,12 +2,27 @@ require "./gorake.rb"
 
 desc 'Bootstrap CI environment'
 task :bootstrap do
-  sh 'go get github.com/robfig/glock'
+  sh 'go get github.com/Masterminds/glide'
```
LeoCavaille: Should we pin glide to a given version or even download the binary directly (even locally not only in the CI)?

galdor: We do it for CircleCI, I guess we could use the same version in local. But then any other repository would have to use the same version, or bootstrap calls will reverse each other. I'll do the chance, but I don't think it really matters.

PR #238, opened 2017-02-24
[quantizer] add new HTTP quantizer

LeoCavaille:
For all spans that have type set to 'http', we try to extract the "host"
part of the URL and set it as a resource or if we can't fallback to a
placeholder so that we don't spam with zillions of resources.

### DIFF
--- a/quantizer/http.go
+++ b/quantizer/http.go
@@ -1,0 +1,37 @@
+package quantizer
+
+import (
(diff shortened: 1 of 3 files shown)


LeoCavaille:
@palazzem does this still need some work? Happy to add missing features or close if we decided another way. Thanks.

PR #239, opened 2017-02-28
[sampling] add pre-sampling support

ufoot:


model/span.go
```diff
@@ -75,3 +75,17 @@ func (s *Span) Weight() float64 {
 
 	return 1.0 / sampleRate
 }
+
+// ApplyRate applies a given rate over the existing one.
+func (s *Span) ApplyRate(rate float64) {
```
LotharSee: Duplicated logic with `SetTraceAppliedSampleRate`, need some refactor.

sampler/presampler.go
```diff
@@ -0,0 +1,218 @@
...
+func (ps *PreSampler) RealRate() float64 {
+	ps.mu.RLock()
+	rate := ps.stats.RealRate()
+	ps.mu.RUnlock()
+	return rate
+}
+
+// RealRate calcuates the current real pre-sample rate from
+// the stats data. If no data is available, returns the target rate.
+func (stats *PreSamplerStats) RealRate() float64 {
```
LotharSee: Because this is from the beginning of the app, the rate will be averaged over way too long and be very imprecise.

Our count of Traces Seen/Dropped should be based on recent data only. The current sampler `Backend` already provides some logic for that, it might be reusable?

ufoot: I will see the current sampler Backend logic but this value is reset on a regular basis so it's not *that* wrong, exponential decay is more precise anyway.

sampler/presampler.go
```diff
@@ -0,0 +1,250 @@
...
+const (
+	// TraceCountHeader is the header client implementation should fill
+	// with the number of traces contained in the payload.
+	TraceCountHeader = "X-Datadog-Trace-Count"
+
+	// Every 100 payload, reset counters. This means that we can not
+	// go below 1% pre-sampling, as the first payload will always be
+	// accepted, since by default the RealRate is the Rate when there
+	// is no data.
+	preSamplerResetPayloads = 100
```
LotharSee: No longer used.

ufoot:
Superseded by https://github.com/DataDog/datadog-trace-agent/pull/240

PR #241, opened 2017-03-07
add an env variable to control the inclusion of build vars in -version

galdor:
We need to be able to build the agent without build variables (commit,
branch, etc.) in omnibus.

### DIFF
--- a/gorake.rb
+++ b/gorake.rb
@@ -6,7 +6,8 @@ def go_build(program, opts={})
   end
   opts = {
     :cmd => default_cmd,
-    :race => false
+    :race => false,
+    :add_build_vars => true
   }.merge(opts)
 
   dd = 'main'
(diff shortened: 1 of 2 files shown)


gorake.rb
```diff
@@ -16,24 +17,27 @@ def go_build(program, opts={})
...
+  vars["#{dd}.Version"] = agentversion
+  if opts[:add_build_vars]
+    vars["#{dd}.BuildDate"] = date
+    vars["#{dd}.GitCommit"] = commit
+    vars["#{dd}.GitBranch"] = branch
+    vars["#{dd}.GoVersion"] = goversion
+  end
+
+  ldflags = vars.map do |k, v|
     if goversion.include?("1.4")
```
talwai: drop this clause since we'll never be building against go1.4?

galdor: With pleasure :)

PR #242, opened 2017-03-09
config: shell out to dd-agent to retrieve hostname

talwai:
The infrastructure agent implements more refined logic for
inferring hostname of the machine this process is running on
https://github.com/DataDog/dd-agent/blob/7cd0986e137683195f129b2a988dd60ca9ccf0ea/utils/hostname.py#L48

Prior to this change, the trace-agent would identify as a different host
than the infra agent in subtle cases (EC2 / GCE for example). This
created phantom hosts and general inconsistency in the UI

We shell out to the infra agent, via the following command, falling back
to running `os.Hostname()` when it fails

```
PYTHONPATH=/opt/datadog-agent/agent /opt/datadog-agent/embedded/bin/python -c "from utils.hostname import get_hostname; print get_hostname()"
```

### DIFF
--- a/config/agent.go
+++ b/config/agent.go
@@ -1,8 +1,10 @@
 package config
 
 import (
+	"bytes"
 	"errors"
 	"os"
+	"os/exec"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


LotharSee:
Can we check with Croissant to know which platform is compatible with that?
And check that it still works well on the others? (including Windows)

config/config_test.go
```diff
@@ -149,3 +149,9 @@ func TestConfigNewIfExists(t *testing.T) {
 	assert.Nil(t, conf)
 	os.Remove(filename)
 }
+
+func TestGetHostname(t *testing.T) {
```
ufoot: This test is very fine but will probably fall back on `os.Hostname` in most cases. OTOH I can't see any way to really trigger the "real test" without either black-box test the thing (so, non-Go solution, complex...) or changing the func prototype and passing `python` and `command` as args (sounds like over engineering for the sake of testability...). So well, good to go, but will keep in mind this is slightly under tested.

config/agent.go
```diff
@@ -104,9 +106,38 @@ func mergeEnv(c *AgentConfig) {
...
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	err := cmd.Run()
+	if err != nil {
+		log.Infof("error retrieving dd-agent hostname, falling back to os.Hostname(): %s", stderr)
+		return os.Hostname()
+	}
+
+	hostname := strings.TrimSpace(stdout.String())
```
ufoot: 👍

talwai:
@LotharSee - confirmed that `utils.get_hostname()` is the canonical source of hostname for all platforms supported by the infra agent, including windows. However the path to the embedded agent Python will differ on Windows so `getHostname()` will end up defaulting to `os.Hostname()` anyway.

I've convinced myself this works on the Linux systems we support, and i think pursuing windows support is out of the scope of this PR (since the hostname logic is not the only thing that will need tweaking)

PR #248, opened 2017-03-17
[sampler] reduce log verbosity and exposing state through expvar

ufoot:


agent/sampler.go
```diff
@@ -78,8 +78,11 @@ func (s *Sampler) Flush() []model.Trace {
 // TODO: remove (or clean it) in a real released build
 func (s *Sampler) logState() {
 	state := s.samplerEngine.(*sampler.Sampler).GetState()
-	log.Infof("inTPS: %f, outTPS: %f, maxTPS: %f, offset: %f, slope: %f, cardinality: %d",
+	log.Debugf("inTPS: %f, outTPS: %f, maxTPS: %f, offset: %f, slope: %f, cardinality: %d",
 		state.InTPS, state.OutTPS, state.MaxTPS, state.Offset, state.Slope, state.Cardinality)
+
+	updateSamplerState(state) // publish through expvar
+
 	statsd.Client.Gauge("datadog.trace_agent.sampler.scoring.offset", state.Offset, nil, 1)
```
ufoot: Open question: should we keep those direct calls to statsd? Because since the values are in expvar, it's rather trivial to get them with https://www.datadoghq.com/blog/instrument-go-apps-expvar-datadog/

PR #251, opened 2017-03-20
config: respect proxy_* settings in agent

### LINKED ISSUE #245: APM Trace agent doesn't use http proxy configuration for reporting
I'm attempting to use APM in an environment that restricts outbound http unless through an http proxy, and from what I can tell this agent is not supporting the http proxy config (or I'm missing the correct place to put it).

I do have at present (`proxy_host`, `proxy_port`, `proxy_user`, `proxy_password`) set inside `/etc/dd-agent/datadog.conf` and have restarted the datadog-agent.

I see the following in my logs:

```
tail -f /var/log/datadog/trace-agent.log
2017-03-16 14:21:50 ERROR (endpoint.go:119) - error when requesting to endpoint https://trace.agent.datadoghq.com/api/v0.1/collector: Post https://trace.agent.datadoghq.com/api/v0.1/collector?api_key=[redacted]: dial tcp 23.23.189.219:443: i/o timeout
2017-03-16 14:21:52 INFO (receiver.go:329) - receiver handled 51 spans, dropped 0 ; handled 51 traces, dropped 0
2017-03-16 14:22:06 ERROR (endpoint.go:119) - error when requesting to endpoint https://trace.agent.datadoghq.com/api/v0.1/collector: Post https://trace.agent.datadoghq.com/api/v0.1/collector?api_key=[redacted]: dial tcp 50.16.188.192:443: i/o timeout
```

talwai:
when [proxy
settings](https://github.com/DataDog/dd-agent/blob/master/datadog.conf.example#L7-L10) are defined in datadog.conf
update the default client in the writer to submit to a proxy.

fixes #245

### DIFF
--- a/config/proxy.go
+++ b/config/proxy.go
@@ -1,0 +1,33 @@
+package config
+
+import (
+	"github.com/go-ini/ini"
+)
+
+const DefaultProxyPort = 3128
+
+type ProxySettings struct {
+	User     string
+	Password string
+	Host     string
+	Port     int
+}
+
+func getProxySettings(m *ini.Section) *ProxySettings {
+	p := ProxySettings{Port: DefaultProxyPort}
+
+	if v := m.Key("proxy_host").MustString(""); v != "" {
+		p.Host = v
+	}
+	if v := m.Key("proxy_port").MustInt(-1); v != -1 {
+		p.Port = v
+	}
+	if v := m.Key("proxy_user").MustString(""); v != "" {
+		p.User = v
+	}
+	if v := m.Key("proxy_password").MustString(""); v != "" {
+		p.Password = v
+	}
+
+	return &p
+}
--- a/config/proxy.go
+++ b/config/proxy.go
@@ -48,12 +48,21 @@ func getProxySettings(m *ini.Section) *ProxySettings {
 // URL turns ProxySettings into an idiomatic URL struct
 func (p *ProxySettings) URL() (*url.URL, error) {
 	// construct scheme://user:pass@host:port
-	var userpass string
+	var userpass *url.Userinfo
 	if p.User != "" {
 		if p.Password != "" {
-			userpass = fmt.Sprintf("%s:%s@", p.User, p.Password)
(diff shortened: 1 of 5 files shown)


config/proxy.go
```diff
@@ -0,0 +1,59 @@
...
+	return &p
+}
+
+// URL turns ProxySettings into an idiomatic URL struct
+func (p *ProxySettings) URL() (*url.URL, error) {
+	// construct scheme://user:pass@host:port
+	var userpass string
+	if p.User != "" {
+		if p.Password != "" {
+			userpass = fmt.Sprintf("%s:%s@", p.User, p.Password)
```
ufoot: I think we could URL encode here (cf https://www.cyberciti.biz/faq/unix-linux-export-variable-http_proxy-with-special-characters/ or http://unix.stackexchange.com/questions/42273/export-http-proxy-and-special-characters-in-passwd ). Probably using https://golang.org/pkg/net/url/#UserPassword would do the job.

config/proxy.go
```diff
@@ -0,0 +1,59 @@
+package config
```
ufoot: I like the idea of having a separate chunk of code / struct to handle this. Indeed, it is not really agent specific.

config/proxy_test.go
```diff
@@ -0,0 +1,68 @@
...
+	s, err = getURL(f)
+	assert.Nil(err)
+	assert.Equal("http://aaditya:password_12@myhost:3129", s.String())
+
+	f, _ = ini.Load([]byte(strings.Join([]string{
+		"[Main]",
+		"proxy_host = https://myhost",
+		"proxy_port = 3129",
+		"proxy_user = aaditya",
+		"proxy_password = password_12",
```
ufoot: You should probably add a test with some special chars such as `/:\!?&;=@éÔγλῶσσα`, and test that `ProxySettings.URL()` returns the expected value.

talwai:
Yes good catch on the special chars handling. I've addressed this in https://github.com/DataDog/datadog-trace-agent/commit/9486db3

PR #254, opened 2017-03-29
ensure the agent recovers from and logs its panics

talwai:


agent/main.go
```diff
@@ -122,6 +122,20 @@ func main() {
...
 
+	defer func() {
+		if err := recover(); err != nil {
+			// Full print of the trace in the logs
+			buf := make([]byte, 4096)
+			length := runtime.Stack(buf, false)
+			stacktrace := string(buf[:length])
+			msg := fmt.Sprintf("%s: %s\n%s", "Unexpected error", err, stacktrace)
+
+			log.Info(msg)
```
clutchski: i think this should be logged at Error level

agent/main.go
```diff
@@ -122,6 +122,20 @@ func main() {
...
+		}
```
clutchski: do you need to re-raise?

PR #257, opened 2017-03-30
writer: don't discard the client on api errors

talwai:
Our payload buffering logic passes along a stateful `APIEndpoint`
when it receives upstream errors. A prior PR ( #251  ) added a `client`
to individual APIEndpoints without propagating this value to the error handler

Thus when an endpoint flush failed, all future flushes would use the
zero value for `APIEndpoint.client` which is `nil`

### DIFF
--- a/agent/endpoint.go
+++ b/agent/endpoint.go
@@ -31,6 +31,10 @@ func (err *apiError) IsEmpty() bool {
 	return len(err.errs) == 0
 }
 
+func (err *apiError) SetClient(client *http.Client) {
+	err.endpoint.client = client
+}
+
 func (err *apiError) Append(url, apiKey string, e error) {
 	err.errs = append(err.errs, e)
 	err.endpoint.urls = append(err.endpoint.urls, url)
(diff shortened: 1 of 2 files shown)


agent/endpoint.go
```diff
@@ -124,6 +128,11 @@ func (a *APIEndpoint) Write(p model.AgentPayload) (int, error) {
 
 	endpointErr := newAPIError()
 
+	// if this payload cannot be flushed due to API errors
+	// we need to pass the client along for future submissions
+	// FIXME(aaditya)
```
ufoot: Short-term fix is fine IMHO but as we see this now only, probably no test is going through that path for real. This is beyond the scope of this PR anyway.

agent/writer.go
```diff
@@ -70,6 +70,7 @@ func NewWriter(conf *config.AgentConfig) *Writer {
 		if conf.Proxy != nil {
 			// we have some kind of proxy configured.
 			// make sure our http client uses it
+			log.Infof("configuring proxy through host %s", conf.Proxy.Host)
```
ufoot: :+1:

PR #258, opened 2017-03-31
handle parallel and async spans for sublayer computations

galdor:


model/sublayers.go
```diff
@@ -1,153 +1,100 @@
...
-		if s.Tag.Name != "" {
-			b.WriteRune('.')
-			b.WriteString(s.Tag.Name)
-			b.WriteRune(':')
-			b.WriteString(s.Tag.Value)
+		// In-place filtering
+		nonAsyncChildren := Spans(children[:0])
+		for _, child := range children {
+			if child.End() <= span.End() {
+				nonAsyncChildren = append(nonAsyncChildren, child)
```
ufoot: Does this imply that if child ends before span, it's considered sync (nonAsync) ? Because, technically, I could imagine a caller firing 100 goroutines and having a sync point waiting for all this goroutines -> all of them finish before the caller, but they are truely async (meaning, the total of their durations is way above the duration of the caller). But I might miss the point.

galdor: Yep, you are 100% right. However we currently do not have the information we need, i.e. some kind of flag indicating that a spans actually blocks (or not) its parent. We may try to improve that in the future.

MattHauglustaine:
This PR is included as part of #296

PR #261, opened 2017-04-11
normalizer: reject traces containing duplicate span ids

galdor:


model/normalizer_test.go
```diff
@@ -276,6 +276,16 @@ func TestNormalizeTraceInvalidSpan(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestNormalizeTraceDuplicateSpanID(t *testing.T) {
+	span1 := testSpan()
+	span2 := testSpan()
```
ufoot: Nitpick, I'd rather set explicitly the span id to the same value, you never know how testSpan() can evolve, could end up behaving differently in the future.

PR #262, opened 2017-04-13
config: allow disabling agent

olivielpeau:
When `apm_enabled` is set to a falsy value the agent should pick it up
and exit.

### DIFF
--- a/config/agent.go
+++ b/config/agent.go
@@ -243,8 +243,8 @@ APM_CONF:
 		goto ENV_CONF
 	}
 
-	if v, _ := conf.Get("Main", "apm_enabled"); v == "true" {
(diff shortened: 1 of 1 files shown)


config/agent.go
```diff
@@ -243,8 +243,8 @@ APM_CONF:
 		goto ENV_CONF
 	}
 
-	if v, _ := conf.Get("Main", "apm_enabled"); v == "true" {
-		c.Enabled = true
+	if v, _ := conf.Get("Main", "apm_enabled"); v == "no" || v == "false" {
```
talwai: maybe `strings.ToLower(...)` as well, which is perhaps more consistent with how dd-agent does it

PR #263, opened 2017-04-20
[log] using receiver limited logger instead of plain log debug

ufoot:
When skipping messages, a debug message is of no use in production mode,
as we get no clue on why traces where dropped. Especially when it's for
an obscure reason such as clock skew. This patch uses the receiver logger
object which has a limiter preventing log to be filled with the same error
repeating many times, while still displaying a few of them.


MattHauglustaine:
#273 set's up a global throttled logger, fixing this issue if approved

MattHauglustaine:
@ufoot #273 has been merged, if it's ok for you I think we can close this issue.

ufoot:
Superseded by #273

PR #266, opened 2017-04-27
Make binding to the legacy port optional

dirceu:
Binding to the legacy port (in addition to the current one) should be optional, although we can keep the default as `true` to keep the current behavior.

### DIFF
--- a/agent/receiver.go
+++ b/agent/receiver.go
@@ -101,9 +101,11 @@ func (r *HTTPReceiver) Run() {
 		die("%v", err)
 	}
 
-	legacyAddr := fmt.Sprintf("%s:%d", r.conf.ReceiverHost, legacyReceiverPort)
-	if err := r.Listen(legacyAddr, " (legacy)"); err != nil {
-		log.Error(err)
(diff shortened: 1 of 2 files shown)


elbuo8:
Default `false` ? 💃

talwai:
I think we may be at a point when we can drop the behavior of binding to 7777 entirely . I don't think any of our older client versions (which reported to 7777) see much use. and in the case they are being used there's a clear upgrade path

palazzem:
I second @talwai idea. Versions that are still using the port `7777` are quite old so I think we can simply drop that legacy port. At the time of writing, the following is the list of libraries that are required to work with an agent *without* the legacy port:
* `dd-trace-py`: version 0.6.0+ (current 0.8.2)
* `dd-trace-rb`: version 0.5.0+ (current 0.7.0)
* `dd-trace-go`: version 0.3.0+ (current 0.3.0 but we need to cut a new "release" here)

dirceu:
Even better :)

LotharSee:
+1 we will certainly drop this legacy port with the next 5.14.0 release.

PR #268, opened 2017-05-04
[toplevelname] identifying top-level name spans

ufoot:
This adds:
- a `TopLevel` func on spans which returns a boolean telling wether it's top-level or not
- a `TopLevel`member to stats, which contains the number of top-level spans contributing to it

### DIFF
--- a/model/statsraw.go
+++ b/model/statsraw.go
@@ -201,9 +201,6 @@ func (sb *StatsRawBucket) HandleSpan(s Span, env string, aggregators []string, w
 }
 
 func (sb *StatsRawBucket) add(s Span, weight float64, aggr string, tags TagSet) {
-	// [TODO:christian] the day we want to skip stats on non top-level spans
-	// totally, do it here.
-
 	var gs groupedStats
 	var ok bool
 
--- a/model/statsraw.go
+++ b/model/statsraw.go
@@ -190,6 +190,11 @@ func (sb *StatsRawBucket) HandleSpan(s Span, env string, aggregators []string, w
 }
 
 func (sb *StatsRawBucket) add(s Span, weight float64, aggr string, tags TagSet) {
+	// if this is not a top-level name -> don't keep track of any stats
+	if _, ok := s.Meta[topLevelTag]; !ok {
+		return
+	}
+
 	var gs groupedStats
 	var ok bool
 
--- a/model/statsraw.go
+++ b/model/statsraw.go
@@ -13,8 +13,7 @@ import (
 type groupedStats struct {
 	tags TagSet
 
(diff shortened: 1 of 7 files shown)


model/top_level.go
```diff
@@ -0,0 +1,69 @@
+package model
+
+const (
+	// TraceMetricsTagKey is a tag key which, if set to true,
+	// ensures all statistics are computed for this span.
+	TraceMetricsTagKey = "datadog.trace_metrics"
+
+	subNameTag   = "_sub_name"
```
LotharSee: Since we reverted the initial sub_name logic to top_level, should we do the same with the metadata? (ie. set a `_top_level` meta with a value `true` when we have one).
Mostly for consistency/simplicity.

model/top_level.go
```diff
@@ -0,0 +1,69 @@
+package model
+
+const (
+	// TraceMetricsTagKey is a tag key which, if set to true,
+	// ensures all statistics are computed for this span.
+	TraceMetricsTagKey = "datadog.trace_metrics"
```
LotharSee: :+1: We can just add a comment that it isn't used for now.

model/statsraw.go
```diff
@@ -83,44 +89,49 @@ func (sb *StatsRawBucket) Export() StatsBucket {
...
-			Key:     key,
-			Name:    k.name,
-			Measure: k.measure,
-			TagSet:  v.tags,
-			Value:   float64(v.value),
+			Key:      key,
+			Name:     k.name,
+			Measure:  k.measure,
+			TagSet:   v.tags,
+			TopLevel: v.topLevel,
```
LotharSee: Initially I'd have said that this is not useful/doesn't have a real meaning, but in practice it is good to keep it consistent with the other counts / doesn't cost more :+1:

model/top_level_test.go
```diff
@@ -0,0 +1,163 @@
+package model
```
LotharSee: Great test suite :+1:

model/statsraw.go
ufoot: Actually, writing the test suite I found out it flags some sublayers counts as useless, such as in https://github.com/DataDog/datadog-trace-agent/blob/0965a171ee2936fad4213b886555c96ab563691b/model/stats_test.go#L390-L393 so while in the ultimate solution we might just drop data earlier in the pipeline and have no use for it, as a transition it could prove useful.

model/top_level.go
ufoot: Need to double-check how it copes with "zero, default data" but yes, you're right, giving it a try.

PR #273, opened 2017-05-09
[logging] Global Throttled Logger

MattHauglustaine:
For now, only the receiver has a throttled logger, dropping messages after a specific amount has been logged.

This patch series sets up Seelog differently, and uses a custom receiver in order to throttle warning and error messages for the whole agent. With this, we could log as usual `log.Error("Oups")` and we are guaranteed to not flood the system.

It thus removes the custom logging system of the receiver and falls back to the usual logging API.

In order to test this PR, you can start the agent and send lots of unacceptable HTTP requests to it. For example, running `./wrk -R 4 -d 30 http://localhost:8126/v0.3/traces` should output: 

```
2017-05-09 12:05:59 ERROR (receiver.go:197) - cannot decode v0.3 traces payload: EOF
2017-05-09 12:05:59 ERROR (receiver.go:197) - cannot decode v0.3 traces payload: EOF
2017-05-09 12:05:59 ERROR (receiver.go:197) - cannot decode v0.3 traces payload: EOF
2017-05-09 12:05:59 ERROR (receiver.go:197) - cannot decode v0.3 traces payload: EOF
2017-05-09 12:05:59 ERROR (receiver.go:197) - cannot decode v0.3 traces payload: EOF
2017-05-09 12:05:59 ERROR (receiver.go:197) - cannot decode v0.3 traces payload: EOF
2017-05-09 12:05:59 ERROR (receiver.go:197) - cannot decode v0.3 traces payload: EOF
2017-05-09 12:05:59 ERROR (receiver.go:197) - cannot decode v0.3 traces payload: EOF
2017-05-09 12:05:59 ERROR (receiver.go:197) - cannot decode v0.3 traces payload: EOF
2017-05-09 12:05:59 ERROR (log.go:98) - Too many messages to log, skipping for a bit...
```


agent/main.go
```diff
@@ -195,8 +196,14 @@ func main() {
 		return
 	}
 
-	// Initialize logging (replacing the default logger)
-	err = config.NewLoggerLevelCustom(agentConf.LogLevel, agentConf.LogFilePath)
+	// Initialize logging (replacing the default logger). No need
+	// to defer log.Flush, it was already done when calling
+	// "SetupDefaultLogger" earlier.
+	logLevel, ok := log.LogLevelFromString(strings.ToLower(agentConf.LogLevel))
```
MattHauglustaine: Here we set up the agent's logger, with a throttling of maximum 10 messages every 10 seconds.

We might want some other setting, and perhaps even a boolean config flag "log_throttling" enabling/disabling it?

agent/log.go
```diff
@@ -0,0 +1,207 @@
...
+	r.maxLogsPerInterval = int64(maxLogsPerInterval)
+	r.rawLogger = rawLogger
+	r.rawLoggerNoFmt = rawLoggerNoFmt
+
+	r.logCount = 0
+	r.tick = time.Tick(time.Duration(interval))
+	r.done = make(chan struct{})
+
+	// Start the goroutine resetting the log count
+	go func() {
```
ufoot: At some point we started to use this: https://github.com/DataDog/datadog-trace-agent/blob/master/watchdog/logonpanic.go#L46 to catch panics within the code. I'm not 100% sure it's relevant in that case because this is about logs (so no tricky code) but generally speaking, having a panic() and no log is... annoying.

quantizer/sql_test.go
```diff
@@ -28,8 +29,8 @@ func SQLSpan(query string) model.Span {
 func TestMain(m *testing.M) {
 	flag.Parse()
 
-	// neutralize logs for tests
-	config.NewLoggerLevelCustom("critical", "")
+	// Disable debug logs in these tests
+	log.UseLogger(log.Disabled)
```
ufoot: :+1: this is cleaner, thanks

agent/main.go
```diff
@@ -195,8 +196,14 @@ func main() {
...
+	if !ok {
+		logLevel = log.InfoLvl
+	}
+	err = SetupLogger(logLevel, agentConf.LogFilePath, 10*time.Second, 10)
```
ufoot: Now that we have this fancy mechanic around log rate limiting, we might want to have this as a config parameter (just to tweak it when, typically, on support and trying to debug something on a customer's box). Could probably be added here: https://github.com/DataDog/datadog-trace-agent/blob/master/config/agent.go#L178

MattHauglustaine: Yes this was my point with an earlier comment.
I'm leaning towards a simple boolean flag used to enable/disable throttling instead of 2 integers for duration and max log counts, what do you think?

agent/log.go
MattHauglustaine: Will do, thanks for spotting this!

MattHauglustaine:
2nd patch series

* Made sure throttling can be disabled by setting duration to 0
* Added the config flag "log_throttling" enabling/disabling throttling. Should we have more fine grained config flags to control max log messages and flush interval?

config/agent.go
```diff
@@ -263,6 +265,11 @@ APM_CONF:
 		c.LogFilePath = v
 	}
 
+	if v := strings.ToLower(conf.GetDefault("trace.config", "log_throttling", "")); v == "no" || v == "false" {
+		log.Error("foo")
```
ufoot: `s/foo/bar/g`

MattHauglustaine:
Merged on master 🥇

PR #274, opened 2017-05-10
[normalizer] Normalize Third Party Integrations

MattHauglustaine:
Sometimes, our client libraries third party integrations might end up sending malformed spans. For example, we noticed some spans tracing a web framework with a missing `http.status_code` tag.

This patch series adds the `normalizeThirdPartyIntegrations` function, called when a span is normalized. For now, the function makes sure web framework integrations sets a `http.status_code` tag, and if not will manually set it to 500.

### DIFF
--- a/model/normalizer.go
+++ b/model/normalizer.go
@@ -33,6 +33,28 @@ var (
 	Year2000NanosecTS = time.Date(2000, time.January, 1, 0, 0, 0, 0, time.UTC).UnixNano()
 )
 
+// webFrameworkIntegrationNames is a set of names corresponding to
+// span names of the web frameworks our client libraries handle.Used
+// in normalizeThirdPartyIntegrations().
+var webFrameworkIntegrationNames = map[string]struct{}{
+	// dd-trace-py
+	"pylons.request":  struct{}{},
+	"bottle.request":  struct{}{},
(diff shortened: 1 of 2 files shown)


model/normalizer.go
```diff
@@ -192,6 +218,28 @@ func NormalizeTrace(t Trace) (Trace, error) {
...
 }
 
+// normalizeThirdPartyIntegrations normalizes the span if we detect
+// it's coming from our own third-party integrations (for example, the
+// Django support of dd-trace-py.
+func (s *Span) normalizeThirdPartyIntegrations() {
+	// If it's coming from one of our web framework integrations,
+	// make sure we always have an HTTP status code.
+	if s.Type == "http" {
+		_, isHandled := webFrameworkIntegrationNames[s.Name]
```
talwai: do we need this additional check? there will be custom instrumentation that uses arbitrary span names. it'd be nice if we could tell people "just set span type to http" and then the agent will handle all the rest.

MattHauglustaine:
We're keeping this on hold for now, as we are thinking of a better solution.

MattHauglustaine:
Closing. We've decided to not do any additional magic, and just fix our third-party integrations.

PR #275, opened 2017-05-10
[build] experimental windows build

ufoot:
Proposal on how to build agent. This breaks, obviously, as some deps are not available for windows (the connection watchdog related tools AFAIK) but at least it should provide a command-line tool that does build a real windows exe.

### DIFF
--- a/watchdog/net_windows.go
+++ b/watchdog/net_windows.go
@@ -1,20 +1,6 @@
 package watchdog
 
-import (
-	"time"
-)
-
 // Net for windows returns basic network info without the number of connections.
 func (pi *CurrentInfo) Net() NetInfo {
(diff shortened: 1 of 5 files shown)


watchdog/net_windows.go
```diff
@@ -0,0 +1,20 @@
+package watchdog
+
+import (
+	"time"
+)
+
+// Net for windows returns basic network info without the number of connections.
+func (pi *CurrentInfo) Net() NetInfo {
+	pi.mu.Lock()
```
ufoot: you can simplify this, just `return NetInfo{}` is enough, this is going to be 0 all the time on MS-Windows anyway.

ufoot:
LGTM !

PR #279, opened 2017-05-15
Remove duplicate feature

gabsn:
Drop duplicating endpoint feature.


agent/endpoint.go
```diff
@@ -61,212 +37,182 @@ type AgentEndpoint interface {
...
-	atomic.AddInt64(&a.stats.TracesBytes, int64(payloadSize))
-	atomic.AddInt64(&a.stats.TracesCount, int64(len(p.Traces)))
-	atomic.AddInt64(&a.stats.TracesStats, int64(len(p.Stats)))
-
-	endpointErr := newAPIError()
-
-	// if this payload cannot be flushed due to API errors
-	// we need to pass the client along for future submissions
-	// FIXME(aaditya)
-	endpointErr.SetClient(a.client)
```
talwai: i think we probably want to keep this? can we scope this change to behave exactly as before but without looping over URLs?

PR #282, opened 2017-05-19
Improve metrics for status codes and number of retries

gabsn:


agent/endpoint.go
```diff
@@ -148,6 +154,11 @@ func (ae *APIEndpoint) Write(p model.AgentPayload) (int, error) {
 	log.Infof("flushed payload to the API, time:%s, size:%d", flushTime, len(data))
 	statsd.Client.Gauge("datadog.trace_agent.writer.flush_duration", flushTime.Seconds(), nil, 1)
 
+	if ae.nbRetry != 0 {
+		tagNbRetry := "nb_retry:" + strconv.FormatUint(uint64(ae.nbRetry), 10)
+		statsd.Client.Incr("datadog.trace_agent.writer.nb_retry", []string{tagNbRetry}, 1)
```
talwai: this seems strange - increment by the value of `ae.nbRetry` rather than using the numeric value as a tag.

LotharSee: I guess here we wanted to track how many times agent are retrying?
So yes, you should instead use a gauge which value is the number of retry. That way we can easily graph max, distributions across hosts, ...

agent/endpoint.go
```diff
@@ -129,6 +131,10 @@ func (ae *APIEndpoint) Write(p model.AgentPayload) (int, error) {
 	}
 	defer resp.Body.Close()
 
+	// We monitor the status code received
+	tagStatusCode := "status_code:" + strconv.Itoa(resp.StatusCode)
```
LotharSee: You should use `fmt` instead

agent/endpoint.go
```diff
@@ -23,6 +23,7 @@ type apiError struct {
 }
 
 func newAPIError(err error, endpoint *APIEndpoint) *apiError {
+	endpoint.nbRetry++
```
talwai: call this `retries` and put it on the `endpointStats`

agent/endpoint.go
```diff
@@ -148,6 +153,7 @@ func (ae *APIEndpoint) Write(p model.AgentPayload) (int, error) {
 	flushTime := time.Since(startFlush)
 	log.Infof("flushed payload to the API, time:%s, size:%d", flushTime, len(data))
 	statsd.Client.Gauge("datadog.trace_agent.writer.flush_duration", flushTime.Seconds(), nil, 1)
+	statsd.Client.Gauge("datadog.trace_agent.writer.nb_retry", float64(ae.nbRetry), nil, 1)
```
talwai: `statsd.Client.Count("datadog.trace_agent.writer.retries", int64(ae.endpointStats.retries), nil, 1)`

PR #285, opened 2017-05-23
[toplevelname] fix concurrent map access

ufoot:
I basically followed the pattern used when we met the same problem with the weight: compute the value beforehand and cache it in some external data struct. Main pros are:
- avoids locking the span fields
- avoids touching the span structure, as everything is out-of-band, not in the main data model

Cons:
- more parameters to functions, which already carry the span + the weight + the list of top_level spans within the trace + the sublayers + ...

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -167,10 +167,13 @@ func (a *Agent) Process(t model.Trace) {
 		pt.Env = tenv
 	}
 
-	weight := pt.weight() // need to do this now because sampler edits .Metrics map
-	topLevelSpans := t.TopLevelSpans()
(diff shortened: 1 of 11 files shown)


LotharSee:
So, what about the reverse approach: it was already "dirty" to pass around the weight. It gets worse with the addition of this map.
So why not adding two private attributes to the Span structure? That would greatly simplify the code. 
And it should not hurt us more to do it that way?

gabsn:
In the long term, shouldn't we replace our maps accessed concurrently by that kind of map: https://godoc.org/golang.org/x/sync/syncmap

I think it can simplify a lot the code at the end.

PR #287, opened 2017-05-30
Raphael/log errors

furmmon:
# Problem
After a trace-agent crash `/var/log/datadog/trace-agent.log` did not contain the error.
In  `/var/log/datadog/supervisord.log` you could find this error:
```
2017-05-22 20:15:01,832 INFO exited: trace-agent (exit status 2; not expected)
2017-05-22 20:15:02,482 INFO gave up: trace-agent entered FATAL state, too many start retries too quickly
```
And by running the trace agent manually you could see the goroutines spans.
More details on trello

# Fix and test
The root of the problem is watchdog.Go() function : if you add a `panic(...)` in it https://github.com/DataDog/datadog-trace-agent/blob/master/watchdog/logonpanic.go#L49, you will reproduce the error.
Now if you go back to the old way (before factoring) of creating go routines
```
	go func() {
		defer watchdog.LogOnPanic()
 		server.Serve(stoppableListener)
	})
	}()
```
and add panics in it, the crash will properly log.

Did tests by running different versions of the trace-agent on one node of staging, adding panics and changing code.

Adding a panic at this exact line: https://github.com/DataDog/datadog-trace-agent/blob/master/model/statsraw.go#L211 reproduces the error explained in the trello card. Doing it with the reverts below will solve the issue (tested it and it does log)

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -168,12 +168,14 @@ func (a *Agent) Process(t model.Trace) {
 	}
 
 	weight := pt.weight() // need to do this now because sampler edits .Metrics map
-	watchdog.Go(func() {
+	go func() {
+		defer watchdog.LogOnPanic()
 		a.Concentrator.Add(pt, weight)
-	})
-	watchdog.Go(func() {
+	}()
+	go func() {
+		defer watchdog.LogOnPanic()
 		a.Sampler.Add(pt)
-	})
+	}()
 }
 
 func (a *Agent) watchdog() {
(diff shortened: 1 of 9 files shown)


agent/agent.go
```diff
@@ -167,10 +167,14 @@ func (a *Agent) Process(t model.Trace) {
 		pt.Env = tenv
 	}
 
-	weight := pt.weight() // need to do this now because sampler edits .Metrics map
+	// Need to do this computation before entering the concentrator
+	// as they access the Metrics map, which is not thread safe.
+	t.ComputeWeight(*root)
+	t.ComputeTopLevel()
```
ufoot: :+1: thanks

PR #288, opened 2017-05-30
config: add the http.status_code aggregator by default

talwai:
`ExtraAggregators` allow certain span metadata fields to be
elevated to top-level keys in `StatsBucket`s. The downstream
effect is to generate metrics with this additional dimension in Datadog's backend

A default `http.status_code` aggregator will inspect incoming spans for the eponymous
metadata, and create the following metrics downstream:
- `trace.<name>.errors.by_http_status`
- `trace.<name>.hits.by_http_status`
both tagged by `http.status_code` (e.g. `500`) and `http.status_class` (e.g. `5xx`)

### DIFF
--- a/config/agent.go
+++ b/config/agent.go
@@ -162,7 +162,7 @@ func NewDefaultAgentConfig() *AgentConfig {
 		APIPayloadBufferMaxSize: 16 * 1024 * 1024,
 
 		BucketInterval:   time.Duration(10) * time.Second,
-		ExtraAggregators: []string{},
+		ExtraAggregators: []string{"http.status_code"},
 
 		ExtraSampleRate: 1.0,
 		PreSampleRate:   1.0,
@@ -296,7 +296,7 @@ APM_CONF:
 	}
 
 	if v, e := conf.GetStrArray("trace.concentrator", "extra_aggregators", ","); e == nil {
-		c.ExtraAggregators = v
+		c.ExtraAggregators = append(c.ExtraAggregators, v...)
 	} else {
 		log.Debug("No aggregator configuration, using defaults")
 	}
--- a/config/agent.go
+++ b/config/agent.go
@@ -296,7 +296,7 @@ APM_CONF:
 	}
 
 	if v, e := conf.GetStrArray("trace.concentrator", "extra_aggregators", ","); e == nil {
-		c.ExtraAggregators = append(c.ExtraAggregators, v...)
+		c.ExtraAggregators = v
 	} else {
 		log.Debug("No aggregator configuration, using defaults")
 	}
--- a/config/agent.go
+++ b/config/agent.go
@@ -296,7 +296,7 @@ APM_CONF:
 	}
 
 	if v, e := conf.GetStrArray("trace.concentrator", "extra_aggregators", ","); e == nil {
-		c.ExtraAggregators = v
+		c.ExtraAggregators = append(c.ExtraAggregators, v...)
 	} else {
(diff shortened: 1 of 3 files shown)


config/agent.go
```diff
@@ -162,7 +162,7 @@ func NewDefaultAgentConfig() *AgentConfig {
 		APIPayloadBufferMaxSize: 16 * 1024 * 1024,
 
 		BucketInterval:   time.Duration(10) * time.Second,
-		ExtraAggregators: []string{},
+		ExtraAggregators: []string{"http.status_code"},
```
LotharSee: What about putting it directly into the defaultAggregators?

ufoot: @LotharSee I don't know, sounds like a good idea yes, but at some point `service` or `service,resource` is much more "universal" than this one. In some cases http just makes no sense, whereas service or resource can always be interpreted as something within the real use-case. So IMHO it's fine to let it here, as it is different from our base "all purpose" aggregators, and in some cases we might even want *not* to aggregate on this. Just my 2c.

LotharSee: Technically, we will require this dimension to be there in all our pages (since the graphs of HTTP stats will be break down by http_status), so removing it just breaks some of the core of the app (and I don't see what we get from removing it).
Plus, it implicitly means that if you want to add your own extra aggregators, you will have to add `http.status_code` to the list.

talwai: I can't locate any source of truth for `defaultAggregators` in the config? The implicit default grain of `env:,service:,resource:` is a private decision made when initializing the StatsBuckets https://github.com/DataDog/datadog-trace-agent/blob/master/model/statsraw.go#L192

Because of this, the special first-order treatment of 'env,service,resource' leaks into many other parts of the code, and i'm a bit wary of going the same route for http status because:
- it affects readability
- the service(s) sending spans to the agent may not even be speaking http at all, and for them we pass around a bunch of empty strings in the code
- (less important) you can't "turn off" the behavior of aggregating by http status - the only reason i can think of that you'd want to do this is memory consumption by the additional per-status buckets

If the issue is that "http.status_code needs to be manually preserved" in the config, maybe we can just change the behavior to append rather than replace the existing `ExtraAggregator`

LotharSee: Right, this `env:,service:,resource:` isn't a real aggregator thing, and `c.aggregators` https://github.com/DataDog/datadog-trace-agent/blob/5d833cd1e6be1aa0e3a8da549599176d19be614f/agent/concentrator.go#L28 is in fact only the extra one https://github.com/DataDog/datadog-trace-agent/blob/0fdebf6c96f6d3afef16453e128db07d779f3160/agent/agent.go#L55

So let's go with the append :+1:

ufoot: I see your points. Interesting. And I maintain what I said above, what made me think adding it along with `service` and `resource` was sensible is the fact that having this in a field named `ExtraAggregators` and which is blatantly here for users to fill with *their* extra aggregator was... confusing.

It's fine to append this to the aggregators given in `ExtraAggregators`. This would mean `ExtraAggregators` <- "`UserExtraAggregators` merged with  `DatadogExtraAggregators`". Any implementation would do, as long as the config file stays unchanged (we don't want users to hack around the conf, then revert it back).

In short: append rather than replace -> 👍

talwai: addressed in https://github.com/DataDog/datadog-trace-agent/pull/288/commits/6dfdaf1f51f72ae753cc4fcea6eede8283bc0ed4

PR #290, opened 2017-06-13
Add protection against invalid status codes

gabsn:
- [x] **tested on staging**

### DIFF
--- a/model/normalizer.go
+++ b/model/normalizer.go
@@ -3,6 +3,7 @@ package model
 import (
 	"errors"
 	"fmt"
+	"strconv"
 	"time"
 
 	log "github.com/cihub/seelog"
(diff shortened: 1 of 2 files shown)


talwai:
apply the same comments from here please https://github.com/DataDog/dd-go/pull/3294

model/normalizer.go
```diff
@@ -152,6 +153,13 @@ func (s *Span) Normalize() error {
 		s.Meta["env"] = NormalizeTag(env)
 	}
 
+	// Status Code
+	if sc, ok := s.Meta["http.status_code"]; ok {
+		if !isValidStatusCode(sc) {
+			delete(s.Meta, "http.status_code")
```
talwai: add a `log.Debug` line just so we tell the user we're dropping their meta.

talwai:
quick comment. Looks good otherwise

model/normalizer.go
gabsn: if we do that, is will quickly become unreadable, is it ok?

talwai: use `DEBUG` level so it will not spam by default. and if we turn it on it quickly makes the problem obvious

PR #296, opened 2017-06-26
[Model] Handle parallel spans when computing sublayers

MattHauglustaine:
This branch imports the work already done by galdor on sublayer duration computation. Changes are done to the `ComputeSublayers` function in order to better handle parallel spans (i.e., multiple active spans for a given time frame).

Now, the total duration is always the total duration of the trace, and if multiple spans are active, they each contribute proportionally when computing the sum of duration for a given attribute (service or type). The new algorithm is described as a comment on the function, and an example is give.

Also, unit test have been updated and dead code has been removed.


model/sublayers.go
```diff
@@ -1,153 +1,262 @@
...
+	// Generate sublayers values
+	values := make([]SublayerValue, 0,
+		len(durationsByService)+len(durationsByType)+1,
+	)
+
+	for service, duration := range durationsByService {
+		values = append(values, SublayerValue{
+			Metric: "_sublayers.duration.by_service",
+			Tag:    Tag{"sublayer_service", service},
+			Value:  float64(int64(duration)),
```
MattHauglustaine: Here, as well as line 124, I'm truncating the duration. My thought were that we don't need to have sub-nanoseconds precision, and might as well have simpler values to deal with (for unit tests, for example). Floats are not precise by nature... It does, however, mean that technically, the sum of each sublayer duration might be of by a unit of 1 from the total duration of the trace.

model/sublayers.go
```diff
@@ -1,153 +1,262 @@
...
 
-	if span.Metrics == nil {
-		span.Metrics = make(map[string]float64, len(sv))
+func (a int64Slice) Len() int           { return len(a) }
+func (a int64Slice) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a int64Slice) Less(i, j int) bool { return a[i] < a[j] }
+
+// buildTraceTimestamps returns the timestamps of a trace, i.e the set
+// of start/end times of each spans
+func buildTraceTimestamps(trace Trace) []int64 {
```
ufoot: Would probably write a quick unit test for that func.

model/sublayers.go
```diff
@@ -1,153 +1,262 @@
...
 }
 
-func insertTS(sts []timeSpan, ts timeSpan) []timeSpan {
-	// don't do anything with unnamed
-	if ts.Name == "" {
-		return sts
+// buildTraceActiveSpansMapping returns a mappging from timestamps to
+// a set of active spans
+func buildTraceActiveSpansMapping(trace Trace, timestamps []int64) map[int64]Spans {
+	activeSpans := make(activeSpansMap, len(timestamps))
```
ufoot: And that one too. This is not a very complex func and it's largely explained in comments, but a simple unit test with clear input/output is nice to have.

model/trace_test.go
```diff
@@ -31,3 +31,25 @@ func TestGetRootFromPartialTrace(t *testing.T) {
 
 	assert.Equal(trace.GetRoot().SpanID, uint64(12341))
 }
+
+func TestTraceChildrenMap(t *testing.T) {
```
ufoot:
```
// <-1-------------------------------->
//    <-2------------>  <-3----------->
//       <-4------->          <-5---->
//              <-6--->
```
^ I love those ASCII art graphs ;)

model/sublayers.go
```diff
@@ -1,153 +1,262 @@
...
-	if span.Metrics == nil {
-		span.Metrics = make(map[string]float64, len(sv))
+func (a int64Slice) Len() int           { return len(a) }
+func (a int64Slice) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a int64Slice) Less(i, j int) bool { return a[i] < a[j] }
+
+// buildTraceTimestamps returns the timestamps of a trace, i.e the set
+// of start/end times of each spans
+func buildTraceTimestamps(trace Trace) []int64 {
+	tsSet := make(map[int64]struct{})
```
ufoot: Would it be possible to have a raw idea of the hash size beforehand? I feel something like `len(trace)` could give a nice hint for sizing. Theorically `2*len(trace)` is the absolute max.

model/sublayers.go
ufoot: As well as a bench, see remark below, sublayers computing has historically be a significant part of agent CPU usage, as well as map access and memory alloc.

model/sublayers.go
MattHauglustaine: Strange I remember thinking of that. Don't know how it didn't end up in the PR. Will do.

model/sublayers.go
MattHauglustaine: Makes sense.

model/sublayers.go
```diff
@@ -1,153 +1,262 @@
...
-			ss.Add(cur)
-		}
+// GoString returns a description of a sublayer value.
+func (v SublayerValue) GoString() string {
+	return v.String()
+}
+
+// ComputeSublayers extracts sublayer values by type and service for a trace
+//
+// Description of the algorithm, with the following trace as an example:
```
LotharSee: Amazing description, great work 👍

model/sublayers_test.go
```diff
@@ -3,101 +3,314 @@ package model
...
+
+	assert.Equal(map[string]float64{
+		"_sublayers.span_count":                                      2.0,
+		"_sublayers.duration.by_type.sublayer_type:db":               30.0,
+		"_sublayers.duration.by_service.sublayer_service:pgsql":      30.0,
+		"_sublayers.duration.by_service.sublayer_service:pgsql-read": 20.0,
+	}, span.Metrics)
 }
 
 func BenchmarkSublayerThru(b *testing.B) {
```
LotharSee: I'm curious here: can you share the comparison of the benchmark before/after this change?

MattHauglustaine: 👍

model/sublayers.go
```diff
@@ -1,153 +1,262 @@
...
-		// FIXME does NOT work for parallel stuff
-		for i, ots := range sts {
-			// >======================< ots
-			//          >======<        ts
-			if ots.Start <= ts.Start && ots.Start+ots.Duration >= ts.Start+ts.Duration {
-				// do not account for if same name
-				if ots.Name == ts.Name {
-					return sts
+	spanChildren := trace.ChildrenMap()
+	for sIdx, span := range trace {
```
MattHauglustaine: @LotharSee I'm iterating over the spans instead of walking in BFS like in the JS implem. I'm having second thoughts on this... what do you think?

MattHauglustaine:
Patch update after review. Benchmark results (regression):

```
    new:    BenchmarkComputeSublayers-4  100000  12750 ns/op  3264 B/op  37 allocs/op
    master: BenchmarkComputeSublayers-4  500000   3822 ns/op  2112 B/op  18 allocs/op
```

model/sublayers.go
```diff
@@ -1,153 +1,262 @@
...
 
-	for _, ts := range ss.byType {
-		mType[ts.Name] += float64(ts.Duration)
-	}
-	for _, ts := range ss.byService {
-		mService[ts.Name] += float64(ts.Duration)
+		durationPortion := float64(end-start) / float64(len(activeSpans))
+
+		for _, span := range activeSpans {
+			key := selector(span)
```
LotharSee: Is there a noticeable performance difference if `selector` is simply a string/key and here we a `if` to test if it is by service or by type?

model/sublayers.go
LotharSee: After looking longer at it, I'm fine with this implementation / it looks equivalent.
A recursive implementation wouldn't be more efficient since you still need to map each span to its children.

MattHauglustaine:
Nope:

model/sublayers.go
MattHauglustaine: Sounded fine to me too, just wanted to be sure I didn't miss anything.

PR #297, opened 2017-06-28
Parse the header to get meta data about the language

gabsn:
- [x] Deployed on staging

New tags available for trace-agent metrics:
- `lang`
- `lang_version`
- `interpreter`
- `trace_version`
- `endpoint`

## Benchmarks

- [master version](https://github.com/DataDog/datadog-trace-agent/commit/1a2473c47d170054325c9998836939a4ae0c23fb): `go test -run=NONE -bench=. > master.txt`

```
BenchmarkAgentTraceProcessing-4           	   10000	    157463 ns/op	   50315 B/op	     705 allocs/op
BenchmarkWatchdog-4                       	 3000000	       477 ns/op	      48 B/op	       1 allocs/op
BenchmarkEncodeAgentPayload-4             	     200	   9962172 ns/op	 2183083 B/op	   21829 allocs/op
BenchmarkHandleSpanRandom-4               	   20000	     65146 ns/op	   28907 B/op	     381 allocs/op
BenchmarkHandleTracesFromOneApp-4         	  200000	      7630 ns/op	    1362 B/op	      22 allocs/op
BenchmarkHandleTracesFromMultipleApps-4   	  200000	      8303 ns/op	    1362 B/op	      22 allocs/op
BenchmarkDecoderJSON-4                    	      10	 121390220 ns/op	25726412 B/op	  329443 allocs/op
BenchmarkDecoderMsgpack-4                 	     100	  20649946 ns/op	 9225670 B/op	  118953 allocs/op
```

- new version: `go test -run=NONE -bench=. > meta.txt`

```
BenchmarkAgentTraceProcessing-4           	   10000	    142788 ns/op	   50233 B/op	     705 allocs/op
BenchmarkWatchdog-4                       	 3000000	       459 ns/op	      48 B/op	       1 allocs/op
BenchmarkEncodeAgentPayload-4             	     200	   6760406 ns/op	 2188907 B/op	   21830 allocs/op
BenchmarkHandleSpanRandom-4               	   20000	     65688 ns/op	   28907 B/op	     381 allocs/op
BenchmarkHandleTracesFromOneApp-4         	  200000	      8342 ns/op	    1362 B/op	      22 allocs/op
BenchmarkHandleTracesFromMultipleApps-4   	  200000	      8682 ns/op	    1362 B/op	      22 allocs/op
BenchmarkDecoderJSON-4                    	      10	 117905562 ns/op	25726409 B/op	  329443 allocs/op
BenchmarkDecoderMsgpack-4                 	     100	  21261181 ns/op	 9225620 B/op	  118953 allocs/op
```

- diff: `benchcmp master.txt meta.txt`

```
benchmark                                   old ns/op     new ns/op     delta
BenchmarkAgentTraceProcessing-4             157463        142788        -9.32%
BenchmarkWatchdog-4                         477           459           -3.77%
BenchmarkEncodeAgentPayload-4               9962172       6760406       -32.14%
BenchmarkHandleSpanRandom-4                 65146         65688         +0.83%
BenchmarkHandleTracesFromOneApp-4           7630          8342          +9.33%
BenchmarkHandleTracesFromMultipleApps-4     8303          8682          +4.56%
BenchmarkDecoderJSON-4                      121390220     117905562     -2.87%
BenchmarkDecoderMsgpack-4                   20649946      21261181      +2.96%

benchmark                                   old allocs     new allocs     delta
BenchmarkAgentTraceProcessing-4             705            705            +0.00%
BenchmarkWatchdog-4                         1              1              +0.00%
BenchmarkEncodeAgentPayload-4               21829          21830          +0.00%
BenchmarkHandleSpanRandom-4                 381            381            +0.00%
BenchmarkHandleTracesFromOneApp-4           22             22             +0.00%
BenchmarkHandleTracesFromMultipleApps-4     22             22             +0.00%
BenchmarkDecoderJSON-4                      329443         329443         +0.00%
BenchmarkDecoderMsgpack-4                   118953         118953         +0.00%

benchmark                                   old bytes     new bytes     delta
BenchmarkAgentTraceProcessing-4             50315         50233         -0.16%
BenchmarkWatchdog-4                         48            48            +0.00%
BenchmarkEncodeAgentPayload-4               2183083       2188907       +0.27%
BenchmarkHandleSpanRandom-4                 28907         28907         +0.00%
BenchmarkHandleTracesFromOneApp-4           1362          1362          +0.00%
BenchmarkHandleTracesFromMultipleApps-4     1362          1362          +0.00%
BenchmarkDecoderJSON-4                      25726412      25726409      -0.00%
BenchmarkDecoderMsgpack-4                   9225670       9225620       -0.00%
```


agent/receiver.go
```diff
@@ -299,14 +306,13 @@ func (r *HTTPReceiver) logStats() {
...
 
-		statsd.Client.Gauge("datadog.trace_agent.heartbeat", 1, []string{"version:" + Version}, 1)
-
-		statsd.Client.Count("datadog.trace_agent.receiver.traces", tracesBytes, []string{"endpoint:traces"}, 1)
-		statsd.Client.Count("datadog.trace_agent.receiver.services", servicesBytes, []string{"endpoint:services"}, 1)
-		statsd.Client.Count("datadog.trace_agent.receiver.span", spans, nil, 1)
-		statsd.Client.Count("datadog.trace_agent.receiver.trace", traces, nil, 1)
-		statsd.Client.Count("datadog.trace_agent.receiver.span_dropped", sdropped, nil, 1)
-		statsd.Client.Count("datadog.trace_agent.receiver.trace_dropped", tdropped, nil, 1)
+		statsd.Client.Gauge("datadog.trace_agent.heartbeat", 1, append(tags, "version:"+Version), 1)
```
palazzem: do we need these tags in the `heartbeat`?

agent/receiver.go
```diff
@@ -52,6 +53,7 @@ type HTTPReceiver struct {
 	conf     *config.AgentConfig
 
 	stats      receiverStats
+	tags       receiverTags
```
palazzem: Probably I'm missing something, but can't we just:
```go
for now := range time.Tick(10 * time.Second) {
    tags := getHeaderTags(req)
    // use the tags variable
```
instead of attaching that to the receiver object and making a lot of copies / locks?

agent/receiver.go
```diff
@@ -336,6 +342,47 @@ type receiverStats struct {
...
+		"lang":           req.Header.Get("Datadog-Meta-Lang"),
+		"lang_version":   req.Header.Get("Datadog-Meta-Lang-Version"),
+		"interpreter":    req.Header.Get("Datadog-Meta-Lang-Interpreter"),
+		"tracer_version": req.Header.Get("Datadog-Meta-Tracer-Version"),
+	}
+
+	tags := []string{}
+	for k, v := range headerFields {
+		if v != "" {
+			tags = append(tags, fmt.Sprintf("%s:%s", k, v))
```
palazzem: can't we just do this instead of allocating `headerFields` that is GC later? also header strings (`Datadog-Meta-Lang`) may be a `const`?

agent/receiver.go
gabsn: Actually we only have access to the req object when it arrives in the handler and the function that sends the metrics every 10 seconds is in another goroutine, so we need to have an intermediary struct to store the tags.

agent/receiver.go
gabsn: I think, the more tags we have, the best it is when we use the app (we can see precisely what we want).

agent/receiver.go
gabsn: Using a map is just an easy way to add tags later, et I don't think it has a huge impact on the GC. What do you mean by `Datadog-Meta-Lang` is a const?

agent/receiver.go
palazzem: but that is not so useful when attached to the heartbeat that is used to see the agent version (and if it's alive)

agent/receiver.go
palazzem: It's more to keep the implementation easy without creating useless objects. The point is don't waste memory just because we have memory. About the const, I mean declaring that string as a const since it never change.

agent/receiver.go
palazzem: especially if the agent is shared across different applications in different languages (so some heartbeats are for one language, and some for another, right?)

gabsn:
For the moment the tests break (because I didn't modify the info part of the agent). @palazzem can you check if the global implementation satisfies you ?

agent/receiver.go
```diff
@@ -44,14 +47,40 @@ const (
...
+	// headerFields is a map used to decode the header metas
+	headerFields = map[string]string{
+		"lang":           "Datadog-Meta-Lang",
+		"lang_version":   "Datadog-Meta-Lang-Version",
+		"interpreter":    "Datadog-Meta-Lang-Interpreter",
+		"tracer_version": "Datadog-Meta-Tracer-Version",
+	}
+
+	// initStats is the map we use to initialize the payload stats
+	initStats = map[string]int64{
```
ufoot: I'm wondering if we have a truely dynamic list of keys here (as if: keys depending on user data at runtime) or if it's a finite list. If it's a finite list (like, a dozen of well-known values) we should keep a Go struct I think. Hash maps make the code easier to read & write as it act as a nice pass through, but here, in the agent, we're doing a great (great) effort to save CPU cycles, and a struct saves quite a bunch of them compared to a map.

agent/receiver.go
```diff
@@ -262,78 +266,226 @@ func (r *HTTPReceiver) handleServices(v APIVersion, w http.ResponseWriter, req *
...
+			accStats = getInitStatsCopy()
+			lastLog = now
+		}
+	}
+}
 
-		tdropped := atomic.SwapInt64(&r.stats.TracesDropped, 0)
-		accStats.TracesDropped += tdropped
+type receiverStats struct {
+	sync.Mutex
```
ufoot: CI complains about that one being copied by value in info.go (here: https://github.com/DataDog/datadog-trace-agent/blob/master/agent/info.go#L86 for example), so probably you need to "`*`" it.

agent/receiver.go
palazzem: Keys are finite but they change over the time because they depend by the current language, interpreter version and so on during each client flush. Once a string is created for a particular application, it will always be the same for the rest of the execution.

If there is something we can do here to reduce the number of CPU cycles it would be great, but we should still keep separated stats for each unique string (or hash) composed by these values: https://github.com/DataDog/datadog-trace-agent/pull/297/files?diff=split#diff-e8961f81006d547c70ab424318805bbaR52

ufoot: Oh, I see, so indeed if it depends on language & possibly client version, it's truely dynamic. It might be constant in 99.99% of the cases but we definitely want to isolate that 1 out of 10000 case where a payload comes that has an unusual language/version/whatever. Fine then. But worth a comment on top of the struct possibly, to sum up this quick tech discussion.

agent/agent.go
```diff
@@ -139,8 +138,8 @@ func (a *Agent) Process(t model.Trace) {
 	root := t.GetRoot()
 	if root.End() < model.Now()-2*a.conf.BucketInterval.Nanoseconds() {
 		log.Errorf("skipping trace with root too far in past, root:%v", *root)
-		atomic.AddInt64(&a.Receiver.stats.TracesDropped, 1)
-		atomic.AddInt64(&a.Receiver.stats.SpansDropped, int64(len(t)))
+		//atomic.AddInt64(&a.Receiver.stats.TracesDropped, 1)
+		//atomic.AddInt64(&a.Receiver.stats.SpansDropped, int64(len(t)))
```
ufoot: I think since we now have a `receiverStats` struct which is lockable, those stats can still be kept. Only acquire the lock and update the values.

agent/receiver.go
palazzem: Yes I think it's better to explain the reason in a comment since it may be counter-intuitive in a few months

agent/receiver.go
```diff
@@ -262,78 +266,226 @@ func (r *HTTPReceiver) handleServices(v APIVersion, w http.ResponseWriter, req *
...
+	tags  []string
+}
+
+func newPayloadStats(req *http.Request) *payloadStats {
+	stats := getInitStatsCopy()
+	tags := getTags(req)
+	return &payloadStats{stats, tags}
+}
+
+func (s *payloadStats) clone() *payloadStats {
```
palazzem: I'm a bit worried about all these copies. Can't we simply:
* get the string in the map -> if it's not here we initialize a new [put your implementation here]; if it's here, we return and use it to increment
* at some point the goroutine that flushes data needs that map so, instead of cloning over and over, why not "steal" the reference? so using the mutex we lock the map, steal that one in a local reference and set the previous reference (in the receiver) to `nil`, or whatever is needed to initialize the map again at the next received payload.

In this way:
* we don't need to allocate over and over the same thing
* we're still thread-safe
* we're faster with lesser allocations
* the code should be simpler

Probably I'm missing some implementation details, but let's iterate over them trying to reach that plan ^^ (or a good compromise)

agent/receiver.go
```diff
@@ -262,78 +266,226 @@ func (r *HTTPReceiver) handleServices(v APIVersion, w http.ResponseWriter, req *
...
+		if value != "" {
+			tags = append(tags, fmt.Sprintf("%s:%s", meta, value))
+		}
+	}
+	return tags
+}
+
+// getHash returns the hash of the tag map
+func getHash(tags []string) uint64 {
+	h := fnv.New64()
```
palazzem: Because we're introducing a lot of new stuff in the receiver pipeline, I'm pretty sure we need some benchmarks before/after your changes. The goal is to add this aggregation with the minimum possible overhead.

agent/receiver.go
```diff
@@ -262,78 +266,226 @@ func (r *HTTPReceiver) handleServices(v APIVersion, w http.ResponseWriter, req *
...
+	var buf bytes.Buffer
+	enc := gob.NewEncoder(&buf)
+	err := enc.Encode(key)
+	if err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+func getTraces(v APIVersion, w http.ResponseWriter, req *http.Request) (model.Traces, bool) {
```
palazzem: @ufoot do we have metrics about which endpoints are used? I'm really incline to remove old endpoints.

agent/agent.go
gabsn: I was thinking: why not create a `Stats` structure part of the agent (like `Receiver`, `Concentrator`, etc.) where we send all the stats of the agent instead of duplicating the code everywhere (for example, in `writer.go` we write the metrics directly but we could save a lot of network by simply use the `Stats` struct that would send all metrics every 10s (like we already do in the `Receiver`.

agent/receiver.go
```diff
@@ -354,3 +342,23 @@ func decodeReceiverPayload(r io.Reader, dest msgp.Decodable, v APIVersion, conte
 		panic(fmt.Sprintf("unhandled content type %q", contentType))
 	}
 }
+
+// headerFields is a map used to decode the header metas
+var headerFields = map[string]string{
+	"lang":           "Datadog-Meta-Lang",
+	"lang_version":   "Datadog-Meta-Lang-Version",
+	"interpreter":    "Datadog-Meta-Lang-Interpreter",
+	"tracer_version": "Datadog-Meta-Tracer-Version",
```
tylerbenson: I think an additional header would be beneficial:
`Datadog-Meta-Tracer-SHA`

Thoughts?

agent/endpoint.go
```diff
@@ -55,10 +55,6 @@ type APIEndpoint struct {
 // of URL (such as https://trace.agent.datadoghq.com) and API
 // keys
 func NewAPIEndpoint(url, apiKey string) *APIEndpoint {
-	if apiKey == "" {
```
palazzem: @gabsn this is not related to `Parse the header to get meta data about the language` so it should be part of another PR and discussed separately.

agent/receiver.go
gabsn: @tylerbenson What would be the difference with Datadog-Meta-Tracer-Version?

agent/endpoint.go
gabsn: I know it was just a quick hack to be able to run the benchs.

agent/stats.go
```diff
@@ -0,0 +1,156 @@
...
+	if rs.Stats[ts.Hash] != nil {
+		rs.Stats[ts.Hash].update(ts.Stats)
+	} else {
+		rs.Stats[ts.Hash] = ts.clone()
+	}
+	rs.Unlock()
+}
+
+func (rs *receiverStats) acc(new *receiverStats) {
+	new.RLock()
```
talwai: don't call it new - this is a restricted word in go. surprised this compiles

agent/endpoint.go
palazzem: be sure to remove this one, or make a proper fix and then rebase that branch on top of that fix

agent/receiver.go
tylerbenson: I decided to have "tracer-version" be a combination of the public facing version plus the git sha, so my suggestion isn't needed anymore.

Basically the SHA is useful when dealing with non-official released versions such as "SNAPSHOT" versions in java.

palazzem: Yes let's keep that way so we have only `Datadog-Meta-Tracer-Version` that should/may contain the client git hash if defined.

agent/stats.go
```diff
@@ -0,0 +1,192 @@
...
+
+func (rs *receiverStats) publish() {
+	rs.RLock()
+	for _, tagStats := range rs.Stats {
+		tagStats.publish()
+	}
+	rs.RUnlock()
+}
+
+// tot returns a Stats struct holding the sum of all stats, independtly of their tags.
```
palazzem: nit: `independently`

agent/stats.go
```diff
@@ -0,0 +1,192 @@
...
+	// Atomically load the stats from ts
+	tracesReceived := atomic.LoadInt64(&ts.TracesReceived)
+	tracesDropped := atomic.LoadInt64(&ts.TracesDropped)
+	tracesBytes := atomic.LoadInt64(&ts.TracesBytes)
+	spansReceived := atomic.LoadInt64(&ts.SpansReceived)
+	spansDropped := atomic.LoadInt64(&ts.SpansDropped)
+	servicesReceived := atomic.LoadInt64(&ts.ServicesReceived)
+	servicesBytes := atomic.LoadInt64(&ts.ServicesBytes)
+
+	// Publish the stats
```
palazzem: Maybe everyone should be aware (and agree) with that change in metric names? /cc @talwai @LotharSee @ufoot

gabsn: @talwai said he's fine with it

palazzem: Good to me so!

agent/receiver.go
```diff
@@ -265,77 +246,96 @@ func (r *HTTPReceiver) handleServices(v APIVersion, w http.ResponseWriter, req *
...
+		r.stats.reset()
 
 		if now.Sub(lastLog) >= time.Minute {
-			updateReceiverStats(accStats)
-			log.Infof("receiver handled %d spans, dropped %d ; handled %d traces, dropped %d",
-				accStats.SpansReceived, accStats.SpansDropped,
-				accStats.TracesReceived, accStats.TracesDropped)
+			// We expose the stats accumulated to expvar
+			updateReceiverStats(accStats.tot())
+			log.Infof("Stats (for the the last minute): %s", accStats.String())
```
palazzem: Do we need to prefix all log lines with `Stats (for the the last minute):`? I think it's a bit redundant

gabsn: Yeah, customers are not supposed to now that the stats we dump are only those gathered during the last minute. Moreover, if there is no traces sent to the agent, we can see it because it will print: `Stats (for the the last minute): `. If we don't prefix the log, a blank will print which is weird.

gabsn: But if you prefer to write another prefix, I'm opened to suggestions (although I think we should indicate that stats are for the last minute).

palazzem: but in that case we should print:
```
Stats (for the the last minute): receiver handled 0 spans, dropped 0 ; handled 0 traces, dropped 0
Stats (for the the last minute): receiver handled 0 spans, dropped 0 ; handled 0 traces, dropped 0
Stats (for the the last minute): receiver handled 0 spans, dropped 0 ; handled 0 traces, dropped 0
```
rather then:
```
Stats (for the the last minute):
Stats (for the the last minute):
Stats (for the the last minute):
```

otherwise it seems an error in the logging system

palazzem: Thinking about it, I'm not sure we need to write something here. The fact that it's related to the last minute can be found using the timestamp (that is always available). So I'd say:
* remove the prefix like it was before
* when we don't have stats, we should print an empty record so:
```
receiver handled 0 spans, dropped 0 ; handled 0 traces, dropped 0
```

I got your point and your suggestion to improve the logs, but I'm not sure this is going to help people since they can rely on timestamps (that are accurate). And we're writing redundant chars for each line.

agent/stats.go
```diff
@@ -0,0 +1,216 @@
...
+	rs.Lock()
+	for _, tagStats := range rs.Stats {
+		tagStats.reset()
+	}
+	rs.Unlock()
+}
+
+// GetLanguages returns the list of the languages used in the traces the agent receives.
+// Eventually this list will be send to our backend through the payload header.
+func (rs *receiverStats) GetLanguages() string {
```
talwai: just call it `Languages()`

agent/receiver.go
gabsn: Solution 1:
```
2017-07-25 18:42:55 INFO (receiver.go:293) - Stats (for the the last minute):
	[endpoint:traces] -> traces received: 19476, traces dropped: 0, traces amount: 7450302 bytes, services received: 0, services amount: 0 bytes
	[endpoint:services] -> traces received: 0, traces dropped: 0, traces amount: 0 bytes, services received: 3, services amount: 342 bytes
```
Solution 2:
```
2017-07-25 18:42:55 INFO (receiver.go:293) -
	[endpoint:traces] -> traces received: 19476, traces dropped: 0, traces amount: 7450302 bytes, services received: 0, services amount: 0 bytes
	[endpoint:services] -> traces received: 0, traces dropped: 0, traces amount: 0 bytes, services received: 3, services amount: 342 bytes
```

If there is no data it will look like this for both solutions:
```
2017-07-25 18:42:55 INFO (receiver.go:293) - No data received
```

If I understand your point, you prefer the second solution, but I personally think the solution 2 is better...

agent/stats.go
talwai: also we need this on the `Receiver` itself so export it one level higher

agent/receiver.go
palazzem: @gabsn why the PR has been merged without addressing my review?

PR #298, opened 2017-07-06
[logger] flush raw loggers when throttled logger is flushed

ufoot:
It's all in the comments, I was trapped seeing an agent dying with no log message at all, and it took me quite a time to find out what the error was.

### DIFF
--- a/agent/log.go
+++ b/agent/log.go
@@ -165,7 +165,13 @@ func (r *ThrottledReceiver) AfterParse(args log.CustomReceiverInitArgs) error {
 }
 
 // Flush implements log.CustomReceiver
-func (*ThrottledReceiver) Flush() {}
+func (r *ThrottledReceiver) Flush() {
+	// Flush all raw loggers, a typical use cases for log is showing an error at startup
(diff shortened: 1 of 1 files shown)


agent/log.go
```diff
@@ -165,7 +165,13 @@ func (r *ThrottledReceiver) AfterParse(args log.CustomReceiverInitArgs) error {
 }
 
 // Flush implements log.CustomReceiver
-func (*ThrottledReceiver) Flush() {}
+func (r *ThrottledReceiver) Flush() {
+	// Flush all raw loggers, a typical use cases for log is showing an error at startup
+	// (eg: "cannot listen on localhost:8126: listen tcp 127.0.0.1:8126: bind: address already in use")
+	// and those are not shown if we don't Flush for real.
+	r.rawLogger.Flush()
+	r.rawLoggerNoFmt.Flush()
```
MattHauglustaine: I think we should safeguard since it's `AfterParse` that sets  `rawLogger` and `rawLoggerNoFmt`, just in case.

PR #300, opened 2017-07-24
Gabin/meta2

gabsn:


agent/stats.go
```diff
@@ -0,0 +1,184 @@
...
+	atomic.StoreInt64(&s.SpansDropped, 0)
+	atomic.StoreInt64(&s.ServicesBytes, 0)
+	atomic.StoreInt64(&s.ServicesMeta, 0)
+}
+
+// hash returns the hash of the tag slice
+func hash(tags []string) uint64 {
+	h := fnv.New64()
+	s := strings.Join(tags, "")
+	h.Write([]byte(s))
```
ufoot: I think here you could avoid joining, AFAIR you could do something like:
```
for _,v := range tags {
    h.Write([]byte(v))
}
```
might be a nitpick but worth a quick benchmark I think

PR #301, opened 2017-07-24
Fix quantizer edge case

furmmon:
The quantizer would return "Non-parsable SQL query" whenever those patterns would be found:
```
`letter-letter` `letter.letter` "letter.letter" "letter-letter"
```
This bug was found by a client with the following sql query that won't be parsed : 
```
INSERT INTO `table-21`.issues
```

### DIFF
--- a/quantizer/sql_test.go
+++ b/quantizer/sql_test.go
@@ -250,8 +250,8 @@ func TestSQLQuantizer(t *testing.T) {
 			"SELECT pg_try_advisory_lock ( ? )",
 		},
 		{
-			"`a-a` `a.a` \"a.a\" \"a-a\"",
-			"a-a a.a a.a a-a",
+			"INSERT INTO `qual-aa`.issues (alert0 , alert1) VALUES (NULL, NULL)",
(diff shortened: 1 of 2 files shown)


furmmon:
[backticks seem to break the sql query parsing](https://trello.com/c/6l1HNBy3/1130-backticks-seem-to-break-the-sql-query-parsing)

quantizer/sql_test.go
```diff
@@ -249,6 +249,10 @@ func TestSQLQuantizer(t *testing.T) {
 			"SELECT pg_try_advisory_lock (123) AS t46eef3f025cc27feb31ca5a2d668a09a",
 			"SELECT pg_try_advisory_lock ( ? )",
 		},
+		{
+			"`a-a` `a.a` \"a.a\" \"a-a\"",
```
palazzem: We should write a real SQL query here. Probably the one that we're skipping right now.

palazzem: Also I think we should add some MySQL queries here, since we're only covering PostgreSQL examples (that's why backticks weren't taken in consideration).

PR #302, opened 2017-07-25
propagate language data in payload headers

talwai:
Since the Receiver now collects data about incoming languages, compose an HTTP header that contains this info for reporting to the backend. 

Header value is a pipe-separated alphabetical string of languages, like so:
```
X-Datadog-Reported-Languages: go|java|python
```

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -15,6 +15,7 @@ import (
 )
 
 const processStatsInterval = time.Minute
+const languageHeaderKey = "X-Datadog-Reported-Languages"
 
 type processedTrace struct {
 	Trace     model.Trace
@@ -99,7 +100,7 @@ func (a *Agent) Run() {
 				Env:      a.conf.DefaultEnv,
 			}
 			var wg sync.WaitGroup
-			wg.Add(3)
+			wg.Add(2)
 			go func() {
 				defer watchdog.LogOnPanic()
 				p.Stats = a.Concentrator.Flush()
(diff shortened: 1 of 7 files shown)


agent/endpoint.go
```diff
@@ -119,7 +119,8 @@ func (ae *APIEndpoint) Write(p model.AgentPayload) (int, error) {
 	queryParams := req.URL.Query()
 	queryParams.Add("api_key", ae.apiKey)
 	req.URL.RawQuery = queryParams.Encode()
-	model.SetAgentPayloadHeaders(req.Header)
+
+	model.SetAgentPayloadHeaders(req.Header, p.Extras())
 	resp, err := ae.client.Do(req)
```
gabsn: Why not calling it `p.Headers()` instead of `p.Extras()`, it'd be less confusing?

talwai: i tried not to use an HTTP specific word. since the payload is independent of the protocol used to transfer it

agent/receiver.go
```diff
@@ -309,13 +309,13 @@ func (r *HTTPReceiver) Languages() string {
 	r.stats.RLock()
 	for tags := range r.stats.Stats {
 		if _, ok := langs[tags.Lang]; !ok {
-			str += tags.Lang + ","
+			str += tags.Lang + "|"
```
LotharSee: Could we have it sorted? Good to have it deterministic even if it doesn't matter for now.

agent/agent.go
```diff
@@ -110,6 +110,11 @@ func (a *Agent) Run() {
 				p.Traces = a.Sampler.Flush()
 				wg.Done()
 			}()
+			go func() {
+				defer watchdog.LogOnPanic()
+				p.SetExtra("X-Datadog-Reported-Languages", a.Receiver.Languages())
```
LotharSee: Make this string a const, it will be more readable/easier to find.

LotharSee: Also, what is the purpose of having it as a goroutine while we could do it just after the `wg.Wait`? Looks simpler/lighter this `SetExtra` is super cheap.

PR #303, opened 2017-07-27
Improve info output of the agent

gabsn:
## Before
```
======================
Trace Agent (v 0.99.0)
======================

  Pid: 38149
  Uptime: 15 seconds
  Mem alloc: 773552 bytes

  Hostname: localhost.localdomain
  Receiver: localhost:8126
  API Endpoint: https://trace.agent.datadoghq.com

  Bytes received (1 min): 11000
  Traces received (1 min): 240
  Spans received (1 min): 360
  WARNING: Traces dropped (1 min): 5
  WARNING: Spans dropped (1 min): 10
  WARNING: Pre-sampling traces: 42.1 %
  WARNING: Pre-sampler: raising pre-sampling rate from 3.1 % to 5.0 %

  Bytes sent (1 min): 3591
  Traces sent (1 min): 6
  Stats sent (1 min): 60
  WARNING: Traces API errors (1 min): 3/4
  WARNING: Services API errors (1 min): 1/2
```

## After
```
======================
Trace Agent (v 0.99.0)
======================

  Pid: 38149
  Uptime: 15 seconds
  Mem alloc: 773552 bytes

  Hostname: localhost.localdomain
  Receiver: localhost:8126
  API Endpoint: https://trace.agent.datadoghq.com

  --- Receiver stats (1 min) ---

  -> tags: python, 2.7.6, CPython, 0.9.0

    Traces received: 70 (10679 bytes)
    Spans received: 984
    Services received: 0 (0 bytes)
    Total data received : 10679 bytes

    WARNING: Traces dropped: 23
    WARNING: Spans dropped: 184

  ------------------------------

  WARNING: Pre-sampling traces: 42.1 %
  WARNING: Pre-sampler: raising pre-sampling rate from 3.1 % to 5.0 %


  Bytes sent (1 min): 3591
  Traces sent (1 min): 6
  Stats sent (1 min): 60
  WARNING: Traces API errors (1 min): 3/4
  WARNING: Services API errors (1 min): 1/2
```


agent/info.go
```diff
@@ -43,16 +43,26 @@ const (
...
-  Spans received (1 min): {{.Status.Receiver.SpansReceived}}
-{{if gt .Status.Receiver.TracesDropped 0}}  WARNING: Traces dropped (1 min): {{.Status.Receiver.TracesDropped}}
-{{end}}{{if gt .Status.Receiver.SpansDropped 0}}  WARNING: Spans dropped (1 min): {{.Status.Receiver.SpansDropped}}
-{{end}}{{if lt .Status.PreSampler.Rate 1.0}}  WARNING: Pre-sampling traces: {{percent .Status.PreSampler.Rate}} %
+  API Endpoint: {{.Status.Config.APIEndpoint}}{{ range $i, $ts := .Status.Receiver }}
+  
+  --- Receiver stats (1 min) ---
+
+  {{if $ts.Tags.Lang}}-> tags: {{ $ts.Tags.Lang }}, {{ $ts.Tags.LangVersion }}, {{ $ts.Tags.Interpreter }}, {{ $ts.Tags.TracerVersion }}
+  {{else}}-> * no tags for those stats *{{end}}
```
talwai: i don't think you need the `no tags` header. just omit the header entirely if tags are not present

agent/info.go
```diff
@@ -81,8 +91,16 @@ func publishUptime() interface{} {
 	return int(time.Since(infoStart) / time.Second)
 }
 
-func updateReceiverStats(s Stats) {
+func updateReceiverStats(rs *receiverStats) {
 	infoMu.Lock()
+
+	rs.RLock()
+	s := make([]tagStats, 0, len(rs.Stats))
```
talwai: you can release the lock before commencing the for loop. better yet expose a threadsafe getter on the `receiverStats` struct and do `stats := rs.Stats() ... for _, tagStats := range stats`.

agent/agent.go
```diff
@@ -141,8 +141,7 @@ func (a *Agent) Process(t model.Trace) {
 		log.Errorf("skipping trace with root too far in past, root:%v", *root)
 
 		// We get the address of the struct holding the stats associated to the tags
-		tags := Tags{Endpoint: "endpoint:process"}
-		ts := a.Receiver.stats.getTagStats(tags)
+		ts := a.Receiver.stats.getTagStats(Tags{})
```
talwai: maybe this function should just take a pointer? that way you can avoid allocating for the empty struct and do `getTagStats(nil)`

gabsn: This is actually a bad idea since then I would call `tagStats, ok := rs.Stats[*tags]` where `tags` is nil. See [here](https://github.com/DataDog/datadog-trace-agent/blob/master/agent/stats.go#L24)

What I could do is passing an empty interface instead, but is it worth it?

talwai: ok leave it as is

PR #308, opened 2017-08-02
Prevent flush Services <1s

furmmon:
The agent flushes currently service metadata whenever a service meta is changed. This PR aims to give it a time limit (two flushed can't be done in under 1 second)


LotharSee:
Closing this PR as the new system now doesn't allow to update service metadata, preventing flappy flushes.

PR #309, opened 2017-08-15
Trace Filtering

p-lambert:
This PR enables filtering out certain traces based on their *root resource name*. The underlying motivation for it is to prevent traces such as those originated from heartbeat requests to be sent to the backend services.

### Usage

A blacklist of regular expressions can be set through the [agent configuration file](https://github.com/DataDog/datadog-trace-agent/blob/master/config/README.md) or through the environment variable `DD_IGNORE_RESOURCE`.

Entries must surrounded by double quotes and separated by comas.

#### INI File
```ini
[trace.ignore]
resource: "GET|POST /heartbeat", "another_entry", "SELECT COUNT(*) FROM BAR"
```
#### ENV variable
```sh
DD_IGNORE_RESOURCE='"GET|POST /heartbeat", "another_entry"'
```
The regular expressions are evaluated using the RE2 engine. Accepted syntax is described [here](https://golang.org/s/re2syntax).

### Performance impact
The performance impact of this feature was measured against two scenarios:

- running the agent with a set of filters that match some traces – rather than degrading the performance this turned out to improve it since we avoid computation when we drop a trace;
- running the agent with a set of filters that don't match any traces – this is the worst case scenario since we have the overhead of filtering without dropping any trace;

In both cases the impact seems to be negligibe:

``` 
BenchmarkAgentTraceProcessing                         	   10000	    120873 ns/op	   50970 B/op	     706 allocs/op
BenchmarkAgentTraceProcessingWithFiltering            	   10000	    114600 ns/op	   45879 B/op	     633 allocs/op
BenchmarkAgentTraceProcessingWithWorstCaseFiltering   	   10000	    125369 ns/op	   51124 B/op	     706 allocs/op
```

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -8,6 +8,7 @@ import (
 	log "github.com/cihub/seelog"
 
 	"github.com/DataDog/datadog-trace-agent/config"
+	"github.com/DataDog/datadog-trace-agent/filters"
 	"github.com/DataDog/datadog-trace-agent/model"
 	"github.com/DataDog/datadog-trace-agent/quantizer"
 	"github.com/DataDog/datadog-trace-agent/sampler"
@@ -35,7 +36,7 @@ func (pt *processedTrace) weight() float64 {
 type Agent struct {
 	Receiver     *HTTPReceiver
 	Concentrator *Concentrator
-	Filter       Filter
+	Filters      []filters.Filter
 	Sampler      *Sampler
 	Writer       *Writer
 
@@ -57,7 +58,7 @@ func NewAgent(conf *config.AgentConfig) *Agent {
 		conf.ExtraAggregators,
 		conf.BucketInterval.Nanoseconds(),
 	)
-	f := NewResourceFilter(conf)
+	f := filters.Setup(conf)
 	s := NewSampler(conf)
 
 	w := NewWriter(conf)
@@ -66,7 +67,7 @@ func NewAgent(conf *config.AgentConfig) *Agent {
 	return &Agent{
 		Receiver:     r,
 		Concentrator: c,
-		Filter:       f,
+		Filters:      f,
 		Sampler:      s,
 		Writer:       w,
 		conf:         conf,
@@ -153,11 +154,16 @@ func (a *Agent) Process(t model.Trace) {
 		return
 	}
 
-	if !a.Filter.Keep(root) {
-		log.Debugf("dropping trace with blacklisted resource: %v", *root)
+	for _, f := range a.Filters {
+		if f.Keep(root) {
+			continue
+		}
+
+		log.Debugf("rejecting trace by filter: %T  %v", f, *root)
 		ts := a.Receiver.stats.getTagStats(Tags{})
 		atomic.AddInt64(&ts.TracesFiltered, 1)
 		atomic.AddInt64(&ts.SpansFiltered, int64(len(t)))
+
 		return
 	}
 
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -35,6 +35,7 @@ func (pt *processedTrace) weight() float64 {
 type Agent struct {
 	Receiver     *HTTPReceiver
 	Concentrator *Concentrator
+	Filter       Filter
 	Sampler      *Sampler
 	Writer       *Writer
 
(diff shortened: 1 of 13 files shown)


p-lambert:
[agent: exclude traces from processing based on root-level rules](https://trello.com/c/t1Dwe1r0/1171-agent-exclude-traces-from-processing-based-on-root-level-rules)

agent/agent.go
```diff
@@ -217,3 +220,10 @@ func (a *Agent) watchdog() {
 
 	updatePreSampler(*a.Receiver.preSampler.Stats())
 }
+
+func (a *Agent) countAsDropped(t model.Trace) {
+	// We get the address of the struct holding the stats associated to the tags
+	ts := a.Receiver.stats.getTagStats(Tags{})
+	atomic.AddInt64(&ts.TracesDropped, 1)
```
talwai: I think traces discarded to comply with a user request should be counted separately from those discarded for being badly formatted. i.e. let's count `TracesFiltered` as a distinct thing from `TracesDropped`.

agent/agent.go
```diff
@@ -35,6 +35,7 @@ func (pt *processedTrace) weight() float64 {
 type Agent struct {
 	Receiver     *HTTPReceiver
 	Concentrator *Concentrator
+	Filter       Filter
```
talwai: maybe take a `[]Filter` here so it can accommodate several blacklists that check against different fields (`name` `service` etc.)

agent/agent.go
```diff
@@ -56,6 +57,7 @@ func NewAgent(conf *config.AgentConfig) *Agent {
 		conf.ExtraAggregators,
 		conf.BucketInterval.Nanoseconds(),
 	)
+	f := NewResourceFilter(conf)
```
talwai: as per above, call this `NewFilters` so that it is more generalized. implementing a name- and service- blacklist can be out of the scope PR, but let's make it easy for us to add this in the future

config/agent.go
```diff
@@ -65,6 +65,8 @@ type AgentConfig struct {
 
 	// http/s proxying
 	Proxy *ProxySettings
+
+	ResourceBlacklist []string
```
talwai: maybe `Blacklist map[string][]string` and have the current implementation only populate `blacklist["resource"]`

agent/agent.go
```diff
@@ -140,14 +143,14 @@ func (a *Agent) Process(t model.Trace) {
...
-		ts := a.Receiver.stats.getTagStats(Tags{})
-
-		atomic.AddInt64(&ts.TracesDropped, 1)
-		atomic.AddInt64(&ts.SpansDropped, int64(len(t)))
+		log.Errorf("skipping trace with root too far in past, root: %v", *root)
+		a.countAsDropped(t)
+		return
+	}
 
+	if !a.Filter.Keep(root) {
```
talwai: in the new "general" scheme, loop over `a.Filters` and check against each. As of now there'll be only one

agent/agent.go
p-lambert: it also bothers me the amount of coupling between the `Agent` and the `Receiver`. This seems to be a classic example of [feature envy](http://wiki.c2.com/?FeatureEnvySmell). Maybe we should abstract this away? @talwai

filters/base.go
```diff
@@ -0,0 +1,19 @@
+package filters
+
+import (
+	"github.com/DataDog/datadog-trace-agent/config"
+	"github.com/DataDog/datadog-trace-agent/model"
+)
+
+// Interface handles Span filtering
+type Interface interface {
```
talwai: just call it `Filter`

filters/resource.go
```diff
@@ -0,0 +1,47 @@
...
+			continue
+		}
+
+		blacklist = append(blacklist, rule)
+	}
+
+	f.blacklist = blacklist
+
+	return f
+}
```
talwai: just do `func NewResource() *Resource`

then in filters/base you can do `return []Filter{NewResource(conf)}

`NewResource(conf)

filters/resource.go
```diff
@@ -0,0 +1,47 @@
...
+	"regexp"
+
+	"github.com/DataDog/datadog-trace-agent/config"
+	"github.com/DataDog/datadog-trace-agent/model"
+
+	log "github.com/cihub/seelog"
+)
+
+// Resource implements a resource-based filter
+type Resource struct {
```
talwai: call this `ResourceFilter`

talwai: maybe it shouldn't be exported?

PR #310, opened 2017-08-18
[sampling] add distributed tracing capabilities

ufoot:


agent/receiver_responses.go
```diff
@@ -46,3 +60,22 @@ func HTTPOK(w http.ResponseWriter) {
...
+	encoder := json.NewEncoder(w)
+	if err := encoder.Encode(response); err != nil {
+		tags := []string{"error:response-error"}
+		statsd.Client.Count(receiverErrorKey, 1, tags, 1)
+		return
+	}
+
+	for k, v := range response.Rates {
+		tags := strings.Split(k, ",") // could be "service:myapp,env:"
+		statsd.Client.Gauge("datadog.trace_agent.sampling_rate", v, tags, 1)
```
LotharSee: What's the purpose of this metric? It will be high cardinality while not providing such a useful metric.

agent/sampler.go
```diff
@@ -31,25 +35,35 @@ type samplerStats struct {
...
-type SamplerEngine interface {
-	Run()
-	Stop()
-	Sample(t model.Trace, root *model.Span, env string) bool
-}
-
 // NewSampler creates a new empty sampler ready to be started
 func NewSampler(conf *config.AgentConfig) *Sampler {
+	engine := sampler.NewSampler(conf.ExtraSampleRate, conf.MaxTPS)
+	engineType := strings.Replace(fmt.Sprint(reflect.TypeOf(engine))[1:], ".", "_", -1)
```
LotharSee: This makes things more complex and less obvious.
Why not simply making it explicit?

sampler/sampler.go
```diff
@@ -139,23 +162,24 @@ func (s *Sampler) Sample(trace model.Trace, root *model.Span, env string) bool {
 		// Check for the maxTPS limit, and if we require an extra sampling.
 		// No need to check if we already decided not to keep the trace.
 		maxTPSrate := s.GetMaxTPSSampleRate()
-		if maxTPSrate < 1 {
-			sampled = ApplySampleRate(root, maxTPSrate)
+		//		if maxTPSrate < 1 {
+		if maxTPSrate < sampleRate { // [TODO:christian] double-check this
```
LotharSee: Why did you change that? maxTPS and sampleRate are unrelated here.

sampler/sampler.go
```diff
@@ -48,18 +60,29 @@ type Sampler struct {
...
 }
 
 // NewSampler returns an initialized Sampler
-func NewSampler(extraRate float64, maxTPS float64) *Sampler {
+func NewSampler(extraRate float64, maxTPS float64) *ScoreSampler {
+	return newGenericSampler(extraRate, maxTPS, &combinedSignatureComputer{}, &agentSampleRateApplier{})
+}
+
+// newGenericSampler returns an initialized Sampler, allowing to choose the signature computer and the sample rate applier.
+func newGenericSampler(extraRate float64, maxTPS float64, computer SignatureComputer, applier SampleRateApplier) *ScoreSampler {
```
LotharSee: Overall, this attempt to get a "generic" sampler makes things much harder in general.
While in practice, we will never have arbitrary combination of computers/appliers.
I had a hard time to understand the current code flow and it is error prone.

There is one non-distributed SignatureSampler and one distributed PrioritySampler. These are quite independent.
Whatever code we want to share between these should simply be functions.

sampler/rate.go
```diff
@@ -0,0 +1,67 @@
+package sampler
+
+import (
+	"strconv"
+
+	"github.com/DataDog/datadog-trace-agent/model"
+)
+
+const (
+	samplingPriorityKey = "sampling.priority"
```
LotharSee: See https://github.com/DataDog/dd-trace-py/pull/325/files#r136846079 ?

agent/receiver.go
```diff
@@ -220,11 +246,25 @@ func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *ht
...
+
+			// Put traces in different channels depending on endpoint/protocol.
+			switch v {
+			case v01:
+				fallthrough
+			case v02:
+				fallthrough
+			case v03:
+				traceChan = r.traces
+			case v04:
```
LotharSee: Re: https://github.com/DataDog/dd-trace-py/pull/325/files#r136864220

We still want to be able not to use the priority sampling and use the "old" signature sampling.
What should we do?

1. we distinguish by endpoint version? v03 is for signature, v04 is for priority?
2. v04 supports both, and we figure it which one to use by looking at the sent payload.

I'd vote for 2., that way the client will be simpler/won't have to figure out when protocol version to use.
But then, we have to figure out if we use the signature or the priority sampling. Possibilities: payload header, check if traces have a priority meta.

sampler/combinedsignature.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/DataDog/datadog-trace-agent/model"
+)
+
+// combinedSignatureComputer allows signature computing using as many hints as
+// possible to get a unique category for traces.
+type combinedSignatureComputer struct {
+}
+
+// ComputeSignatureWithRootAndEnv generates the signature of a trace knowing its root
+// Signature based on the hash of (env, service, name, resource, is_error) for the root, plus the set of
```
LotharSee: Not to self: it is time to simplify this signature to only care about the root `(env, service, name, resource, is_error)`

agent/agent.go
```diff
@@ -53,26 +54,30 @@ type Agent struct {
...
-		Sampler:      s,
-		Writer:       w,
-		conf:         conf,
-		exit:         exit,
-		die:          die,
+		Receiver:           r,
+		Concentrator:       c,
+		Filters:            f,
+		Sampler:            s,
+		DistributedSampler: ds,
```
LotharSee: Let's call this "prioritySampler" instead.

agent/agent.go
```diff
@@ -112,7 +120,11 @@ func (a *Agent) Run() {
 			}()
 			go func() {
 				defer watchdog.LogOnPanic()
+				// Serializing both flushes, classic agent sampler and distributed sampler,
+				// in most cases only one will be used, so in mainstream case there should
+				// be no performance issue, only in transitionnal mode can both contain data.
 				p.Traces = a.Sampler.Flush()
+				p.Traces = append(p.Traces, a.DistributedSampler.Flush()...)
```
LotharSee: So, all traces in the same payload? How is the backend knowing which sampling logic to apply?

cc @MattHauglustaine https://github.com/DataDog/dd-go/pull/3741/files#diff-a82552013e7d8d23211f360de47ecef8R65

MattHauglustaine: Oh thats fine. I assumed the payload would contain either all new traces, or all old traces, but that would be an easy fix.

agent/agent.go
```diff
@@ -201,10 +211,22 @@ func (a *Agent) Process(t model.Trace) {
...
 		defer watchdog.LogOnPanic()
-		a.Sampler.Add(pt)
+		s.Add(pt)
 	}()
 }
 
+// Process is the default work unit that receives a trace, transforms it and
+// passes it downstream
+func (a *Agent) Process(t model.Trace) {
+	a.processWithSampler(t, a.Sampler)
```
LotharSee: I guess the test between signature vs priority sampler could be done here, by checking if the trace has a sampling priority?

agent/info.go
```diff
@@ -138,6 +144,19 @@ func publishSamplerInfo() interface{} {
...
+func updateRateByService(rbs map[string]float64) {
+	infoMu.Lock()
+	infoRateByService = rbs
+	infoMu.Unlock()
+}
+
+func publishRateByService() interface{} {
+	infoMu.RLock()
+	rbs := infoRateByService
+	infoMu.RUnlock()
```
LotharSee: Simply

```
infoMu.RLock()
defer infoMu.RUnlock()
return infoRateByService
```

agent/receiver_responses.go
```diff
@@ -46,3 +60,22 @@ func HTTPOK(w http.ResponseWriter) {
 	w.WriteHeader(http.StatusOK)
 	io.WriteString(w, "OK\n")
 }
+
+// HTTPRateByService outputs, as a JSON, the recommended sampling rates for all services.
+func HTTPRateByService(w http.ResponseWriter, rates *sampler.RateByService) {
```
LotharSee: Should we also make a endpoint only returning this? Helpful for debug?

I also haven't figured out yet what to do in the client before the first flush: we won't have the per-service rate. Maybe that could be used?

sampler/sampler.go
ufoot: Was not totally clear to me at the time. I probably need to write a test that is broken by this change (it does not exist now I think), they are indeed unrelated and calling twice ApplySampleRate does indeed cumulates the sampling rates and does not "override" the first call (probably why I got it wrong). Will revert that & write the corresponding test.

agent/receiver_responses.go
ufoot: I might only keep it for debugging (for that matter it's quite convenient, as you know exactly what kind of decision is done by the sampler on a per-service basis), but probably yes, the cardinality is dangerous in production use.

agent/receiver.go
ufoot: OK, got it, quite a change I think for we might end up setting explicit priority meta to 0. I'd be against putting something in the header, if we have such a thing that is outside the body of what we send, I'd rather have an explicit v0.4 protocol (and indeed, it is different as the data returned is different) rather than have a channel which has different behaviors depending on some meta, it's like having 0.3 and 0.4 coexisting without explicitely saying it).

agent/agent.go
ufoot: Indeed (but requires sending explicit sampling priority set to 0, which is not done yet)

agent/receiver_responses.go
ufoot: This is already exposed by expvar if you `curl http://localhost:8126/debug/vars` (for debugging, the metric mentionned in the beginning of the PR can be helpful too)

agent/info.go
ufoot: I'm generally speaking a strong advocate of using defer whenever possible but also remember us hunting after them lately ;) Will happily use that pattern anyway, no prob, thanks for suggesting.

agent/sampler.go
ufoot: I think initially I was worried of `"."` and `"*"` not playing nice with JSON serialization. I now realise they should only be hash keys and not object attributes, so should be fine.

agent/agent.go
LotharSee: Yeah, we should have our clients either not reporting the priority at all (in Python, distributed_sampler set to None) or always reporting it (non None distributed_sampler).

agent/info.go
LotharSee: I don't remember the trouble you had with it. But yeah, clearly not a big deal/doesn't really matter.

agent/receiver.go
LotharSee: So what would you suggest?
My concern is that we will keep coming up with new versions of the endpoint/protocol while we will still want to support both versions of the sampler.
So freezing the signature sampling to the v03 might be cumbersome for the future.

sampler/sampler.go
LotharSee: Yeah, happy to dive more into this piece of the code to be sure we do the right thing.

`maxTPSrate < 1` is still the right condition to check for figuring out if we need to apply the maxTPSrate to the incoming trace in the case of signature sampling.

sampler/rate.go
```diff
@@ -0,0 +1,59 @@
...
+type clientSampleRateApplier struct {
+	rates *RateByService
+}
+
+// ApplySampleRate, when using client sampling, works in two steps:
+// - store the sample rate in the rates by service map, so that next time a client
+//   asks for the sampling rate for such a trace, it gest this result
+// - use the information that was in the meta tags ("_sampling_priority_v1") to
+//   decide wether this one should be sampled or not.
+func (csra *clientSampleRateApplier) ApplySampleRate(root *model.Span, sampleRate float64) bool {
```
LotharSee: Why this logic here?

- for priority sampling, there is no extra sample rate to apply here. We only rely on what comes from the client.
- what we send to the client should include the maxTPS component.

agent/receiver.go
ufoot: Combination of both:
- change to v04 because the protocol actually changed (return value of `OK` hardcoded in v03 is indeed different from a valid JSON in v04)
- read the actual content of traces to put them in the right pipeline, regardless of what enveloppe (version) they are coming from

config/agent.go
```diff
@@ -322,6 +326,12 @@ APM_CONF:
 	if v, e := conf.GetFloat("trace.sampler", "max_traces_per_second"); e == nil {
 		c.MaxTPS = v
 	}
+	if v := strings.ToLower(conf.GetDefault("trace.sampler", "priority_sampling", "")); v == "yes" || v == "true" {
+		c.PrioritySampling = true
```
LotharSee: What is the purpose of making it an option? Isn't the check for the sampling meta enough?
Plus, why having it off by default? It would be extra friction, while the priority sampling is already off by default in the clients.

ufoot: Making it an option is convenient because it avoids being forced to restart your app to disable it within the client lib. But I might change the default anyway, indeed, to lower the barrier for using it.

agent/stats.go
```diff
@@ -160,13 +172,18 @@ func (s *Stats) String() string {
...
+	tracesPriority0 := atomic.LoadInt64(&s.TracesPriority0)
+	tracesPriority1 := atomic.LoadInt64(&s.TracesPriority1)
 	tracesBytes := atomic.LoadInt64(&s.TracesBytes)
 	servicesReceived := atomic.LoadInt64(&s.ServicesReceived)
 	servicesBytes := atomic.LoadInt64(&s.ServicesBytes)
 
-	return fmt.Sprintf("traces received: %v, traces dropped: %v, traces filtered: %v, "+
-		"traces amount: %v bytes, services received: %v, services amount: %v bytes",
-		tracesReceived, tracesDropped, tracesFiltered, tracesBytes, servicesReceived, servicesBytes)
+	return fmt.Sprintf("traces received: %d, traces dropped: %d, traces filtered: %d, "+
```
LotharSee: Is that used somewhere for logging? If yes, the line is getting long, we could:

- not repeat `traces`/`services`
- dropped/priority 0/high priority don't mean anything to users. We'd have to phrase it differently.

If it is just debug/internal, it's fine.

sampler/prioritysampler.go
```diff
@@ -0,0 +1,106 @@
...
+	updateSampleRateForPriority(root, sampleRate, defaultSampleRate, s.rates)
+
+	if sampled {
+		// Count the trace to allow us to check for the maxTPS limit.
+		// It has to happen before the maxTPS sampling.
+		s.Sampler.Backend.CountSample()
+
+		// Check for the maxTPS limit, and if we require an extra sampling.
+		// No need to check if we already decided not to keep the trace.
+		maxTPSrate := s.Sampler.GetMaxTPSSampleRate()
```
LotharSee: What is the purpose of this 2 lines? Looks like lines to be removed?

As with the priority sampling, `maxTPS` only drives the internal mechanics of the per-service sampling computation, and once the client made the decision we stick to it.

sampler/prioritysampler.go
```diff
@@ -0,0 +1,106 @@
...
+func (s *PriorityEngine) Run() {
+	s.Sampler.Run()
+}
+
+// Stop stops the main Run loop
+func (s *PriorityEngine) Stop() {
+	s.Sampler.Stop()
+}
+
+func updateSampleRateForPriority(root *model.Span, sampleRate, defaultSampleRate float64, rates *RateByService) {
```
LotharSee: Would deserve a nice docstring, telling that this only updates the `rates`, so that the receiver is up-to-date.

sampler/prioritysampler.go
```diff
@@ -0,0 +1,106 @@
...
+	initialRate := GetTraceAppliedSampleRate(root)
+	newRate := initialRate * sampleRate
+	SetTraceAppliedSampleRate(root, newRate)
```
LotharSee: It is incorrect to chain the sample rates (pick the previous / multiply / store the new one) for priority sampling.
It might even badly affect server-side sampling as it would give a wrong information (this new rate isn't even the actual rate applied by the client + agent).
I'm not sure it is clear enough, ping me for me details.

You can drop these 3 lines.

sampler/prioritysampler.go
```diff
@@ -0,0 +1,106 @@
...
+
+	var env string
+	if root.Meta != nil {
+		env = root.Meta["env"]
+	}
+	rates.Set(root.Service, env, newRate)
```
LotharSee: Updating `rates` only at that moment isn't enough, since it doesn't cover:

- Over time, per-service rates can change (because of the decay). Here we don't take it into consideration and we may provide an "old rate" (which would be too low). 
- When `AdjustScoring` runs, all coefficients can be adjusted and so are resulting rates.

agent/agent.go
```diff
@@ -15,8 +15,11 @@ import (
 	"github.com/DataDog/datadog-trace-agent/watchdog"
 )
 
-const processStatsInterval = time.Minute
-const languageHeaderKey = "X-Datadog-Reported-Languages"
+const (
+	processStatsInterval = time.Minute
+	languageHeaderKey    = "X-Datadog-Reported-Languages"
+	samplingPriorityKey  = "_sampling_priority_v1"
```
ufoot: This is only for v1, private and for internal usage until we have OT support, with the final `sampling.priority` tag. Also, this is a metrics for now.

agent/agent.go
```diff
@@ -143,12 +167,31 @@ func (a *Agent) Process(t model.Trace) {
...
+			s = a.PriorityEngine
+
+			if priority == 0 {
+				priorityPtr = &ts.TracesPriority0
+			} else {
+				priorityPtr = &ts.TracesPriority1
+			}
+		}
+	}
+	atomic.AddInt64(priorityPtr, 1)
```
ufoot: Here this looks redundant with received traces, but not quite, actually traces can be dropped for several reasons , some because they were not normalisable (this would be done before) some because they are outdated (this would be done after). All in all, having this metric (which is then exposed with tags depending on wether priority is none, 0, or non-zero) is very useful to inspect what is going on.

agent/receiver.go
```diff
@@ -86,11 +92,13 @@ func (r *HTTPReceiver) Run() {
 	http.HandleFunc("/v0.1/services", r.httpHandleWithVersion(v01, r.handleServices))
 	http.HandleFunc("/v0.2/traces", r.httpHandleWithVersion(v02, r.handleTraces))
 	http.HandleFunc("/v0.2/services", r.httpHandleWithVersion(v02, r.handleServices))
-
-	// current collector API
 	http.HandleFunc("/v0.3/traces", r.httpHandleWithVersion(v03, r.handleTraces))
 	http.HandleFunc("/v0.3/services", r.httpHandleWithVersion(v03, r.handleServices))
 
+	// current collector API
+	http.HandleFunc("/v0.4/traces", r.httpHandleWithVersion(v04, r.handleTraces))
```
ufoot: This new endpoint has the same input as `/v0.3`, only `/v0.4/traces` returns a JSON with the sampling rates for all services/envs.

agent/stats.go
ufoot: It's indeed used in logs, I backed to the previous version which does not display the priority info.

sampler/prioritysampler_test.go
```diff
@@ -0,0 +1,155 @@
...
+		// Set signature score offset high enough not to kick in during the test.
+		s.Sampler.signatureScoreOffset = 2 * tc.tps
+		s.Sampler.signatureScoreFactor = math.Pow(s.Sampler.signatureScoreSlope, math.Log10(s.Sampler.signatureScoreOffset))
+
+		sampledCount := 0
+		handledCount := 0
+
+		for period := 0; period < initPeriods+periods; period++ {
+			s.Sampler.Backend.DecayScore()
+			s.Sampler.AdjustScoring()
```
ufoot: This really needs to be called, and at some point, having it running now and then gives some fuzziness to the results (so, the tolerance interval is slightly larger than in the previous sampler)

sampler/ratebyservice.go
```diff
@@ -0,0 +1,90 @@
...
+
+func byServiceKey(service, env string) string {
+	return "service:" + service + ",env:" + env
+}
+
+// RateByService stores the sampling rate per service.
+type RateByService struct {
+	timeout time.Duration
+	rates   map[string]timeoutValue
+	mutex   sync.RWMutex
```
ufoot: This one is here to allow exposition of the rate while updating it on the fly.

sampler/score.go
```diff
@@ -21,23 +21,38 @@ func SampleByRate(traceID uint64, sampleRate float64) bool {
 	return true
 }
 
+func capTo1(f float64) float64 {
```
ufoot: dummy function to avoid copy/pasting

agent/info_test.go
```diff
@@ -33,17 +33,20 @@ Trace Agent (v 0.99.0)
...
     Traces received: 0 (0 bytes)
     Spans received: 0
     Services received: 0 (0 bytes)
-    Total data received : 0 bytes
-    
+    Total data received: 0 bytes
+
   ------------------------------
 
+  Rate for 'service:myapp,env:dev': 12.3 %
```
LotharSee: Let's call this `Sample rate`, as it is user exposed that's more understandable.

agent/receiver.go
```diff
@@ -62,12 +67,13 @@ type HTTPReceiver struct {
 }
 
 // NewHTTPReceiver returns a pointer to a new HTTPReceiver
-func NewHTTPReceiver(conf *config.AgentConfig) *HTTPReceiver {
+func NewHTTPReceiver(conf *config.AgentConfig, rates *sampler.RateByService) *HTTPReceiver {
```
LotharSee: It looks like a non-sustainable change of the API: the more we will return things to the Agent, the more complex this will get.

What about making `rates` a part of `conf`?

config/agent.go
LotharSee: Cool, the option is fair and the `true` default is good.

sampler/ratebyservice.go
```diff
@@ -0,0 +1,90 @@
...
+		rate = 1
+	}
+
+	rbs.mutex.Lock()
+	defer rbs.mutex.Unlock()
+
+	if rbs.rates == nil {
+		rbs.rates = make(map[string]timeoutValue, 1)
+	}
+	timeout := rbs.timeout
```
LotharSee: I'm not sure about the need for a timeout at this level of the interface.

It makes the logic more complex, while the only thing we'd expect from `RateByService` is to expose the content of the per-service PrioritySampler. 
So I'd expect the expiration to happen inside the PrioritySampler (and I think it already does).

agent/receiver.go
ufoot: OK, just introduced a `DynamicConfig` object. Indeed, with it, the receiver constructor is less bounded to current impl, it only has to know "there's something dynamic".

config/agent.go
```diff
@@ -38,9 +38,11 @@ type AgentConfig struct {
...
 
 	// Sampler configuration
-	ExtraSampleRate float64
-	PreSampleRate   float64
-	MaxTPS          float64
+	ExtraSampleRate        float64
+	PreSampleRate          float64
+	MaxTPS                 float64
+	PrioritySampling       bool
+	PrioritySamplerTimeout time.Duration
```
LotharSee: This is no longer used, let's fully drop it.

PR #312, opened 2017-08-22
Fix small bug

gabsn:
Add more context in the case we have an issue obfuscating an sql span resource.

### DIFF
--- a/quantizer/sql.go
+++ b/quantizer/sql.go
@@ -11,7 +11,6 @@ import (
 const (
 	sqlQueryTag      = "sql.query"
 	sqlQuantizeError = "agent.parse.error"
-	sqlOriginalQuery = "sql.original.query"
 )
 
 // TokenFilter is a generic interface that a TokenConsumer expects. It defines
@@ -217,7 +216,9 @@ func QuantizeSQL(span model.Span) model.Span {
 			span.Meta = make(map[string]string)
 		}
 		span.Meta[sqlQuantizeError] = "Query not parsed"
-		span.Meta[sqlOriginalQuery] = span.Resource
+		if span.Meta[sqlQueryTag] == "" {
+			span.Meta[sqlQueryTag] = span.Resource
+		}
 		span.Resource = "Non-parsable SQL query"
 		return span
 	}
--- a/quantizer/sql.go
+++ b/quantizer/sql.go
@@ -216,7 +216,7 @@ func QuantizeSQL(span model.Span) model.Span {
 			span.Meta = make(map[string]string)
 		}
 		span.Meta[sqlQuantizeError] = "Query not parsed"
-		if span.Meta[sqlQueryTag] == "" {
+		if _, ok := span.Meta[sqlQueryTag]; !ok {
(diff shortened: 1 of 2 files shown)


quantizer/sql.go
```diff
@@ -208,18 +209,16 @@ func QuantizeSQL(span model.Span) model.Span {
...
 		// users resources. Here we provide more details to debug the problem.
 		log.Debugf("Error parsing the query: `%s`", span.Resource)
-		span.Resource = "Non-parsable SQL query"
-
 		if span.Meta == nil {
 			span.Meta = make(map[string]string)
 		}
-
 		span.Meta[sqlQuantizeError] = "Query not parsed"
+		span.Meta[sqlOriginalQuery] = span.Resource
```
talwai: in general our clients always send the full query text under the `sql.query` metadata https://github.com/DataDog/dd-trace-py/blob/master/ddtrace/ext/sql.py#L10

this metadata key also gets special treatment in the UI. rather than duplicate the info here, let's just check if the `sql.query` key exists, and add it if not

quantizer/sql.go
```diff
@@ -208,18 +208,18 @@ func QuantizeSQL(span model.Span) model.Span {
...
 		// users resources. Here we provide more details to debug the problem.
 		log.Debugf("Error parsing the query: `%s`", span.Resource)
-		span.Resource = "Non-parsable SQL query"
-
 		if span.Meta == nil {
 			span.Meta = make(map[string]string)
 		}
-
 		span.Meta[sqlQuantizeError] = "Query not parsed"
+		if span.Meta[sqlQueryTag] == "" {
```
talwai: hmm it won't ever be set as the empty string. do it like so
```
if _, ok := span.Meta[sqlQueryTag]; !ok {
            span.Meta[sqlQueryTag] = span.Resource
}
```

talwai:
can we add a test?

quantizer/sql.go
gabsn: It will, because the map has to return strings: https://play.golang.org/p/-JzOOGRJaP

gabsn:
We could add a test, if we are able to find a string `s` such as `quantizedString, err := tokenQuantizer.Process(s)` where `err == nil && quantizedString == ""`.

palazzem:
We need a test for that since this part is **really** crucial and important. I think it should be possible to find such string `s` because I assume that if we created this PR is because one client can generate such input. Let me know if you need help on that @gabsn!

Thank you!

quantizer/sql_test.go
```diff
@@ -90,7 +90,6 @@ func TestSQLResourceWithError(t *testing.T) {
 	for _, tc := range testCases {
 		spanQ := Quantize(tc.span)
 		assert.Equal("Non-parsable SQL query", spanQ.Resource)
-		assert.Equal("", spanQ.Meta["sql.query"])
```
talwai: ```assert.Equal(spanQ.Resource, spanQ["sql.query"])```

PR #315, opened 2017-09-20
Add timeout between the agent and the backend

gabsn:
- [x] on staging since **20/09/2017, 11h39**

### DIFF
--- a/agent/endpoint.go
+++ b/agent/endpoint.go
@@ -15,6 +15,9 @@ import (
 	"github.com/DataDog/datadog-trace-agent/watchdog"
 )
 
+// timeout is used by the agent to release pressure on the backend
+var timeout = 10 * time.Second
+
 // apiError stores the error triggered we can't send data to the endpoint.
 // It implements the error interface.
 type apiError struct {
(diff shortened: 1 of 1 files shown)


agent/endpoint.go
```diff
@@ -62,7 +62,9 @@ func NewAPIEndpoint(url, apiKey string) *APIEndpoint {
 	ae := APIEndpoint{
 		apiKey: apiKey,
 		url:    url,
-		client: http.DefaultClient,
+		client: &http.Client{
+			Timeout: 10 * time.Second,
```
talwai: use a constant. Also this setting will be overriden when we use a proxy, so we need to update it here https://github.com/DataDog/datadog-trace-agent/pull/315/files#L83-L87

agent/endpoint.go
```diff
@@ -15,6 +15,9 @@ import (
 	"github.com/DataDog/datadog-trace-agent/watchdog"
 )
 
+// timeout is used by the agent to release pressure on the backend
```
talwai: change the comment 
```
// timeout is the HTTP timeout for POST requests to the Datadog backend
```

PR #323, opened 2017-10-10
Fix bug in Windows rake command

bmermet:
The command to build the windows target was broken because of mistake when concatenating the two parts of the command.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -91,12 +91,12 @@ task :windows do
   ["386", "amd64"].each do |arch|
     case os
     when "windows"
-        set_env = "set \"GOOS=windows\" && set \"GOARCH=#{arch}\""
+        set_env = "set \"GOOS=windows\" && set \"GOARCH=#{arch}\" &&"
     else
         set_env = "GOOS=windows GOARCH=#{arch}"
     end
     go_build("github.com/DataDog/datadog-trace-agent/agent", {
-               :cmd => set_env + "&& go build -a -o trace-agent-windows-#{arch}.exe",
+               :cmd => set_env + " go build -a -o trace-agent-windows-#{arch}.exe",
                :race => ENV['GO_RACE'] == 'true'
              })
   end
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -96,7 +96,7 @@ task :windows do
         set_env = "GOOS=windows GOARCH=#{arch}"
     end
     go_build("github.com/DataDog/datadog-trace-agent/agent", {
-               :cmd => set_env + " go build -a -o trace-agent-windows-#{arch}.exe",
+               :cmd => set_env + "&& go build -a -o trace-agent-windows-#{arch}.exe",
```
palazzem:
```
GOOS=windows GOARCH=#{arch} && go build -a -o trace-agent-windows-#{arch}.exe
```
could be a full command? not sure it works when `os` doesn't evaluate to `windows`.

bmermet: Initially the command worked only on UNIX, I added the windows case a couple of months ago. It does work when the os doesn't evaluate to windows because the go compiler is able to cross compile.

palazzem: What I mean is if the command above is executed that way, because it doesn't work if it's the case:

```bash
$ GOOS=windows env | grep GOOS
# -> GOOS=windows
$ GOOS=windows && env | grep GOOS
# -> returns nothing because the ENV variable is not set since `env` is executed as a second command
```

So my question is if a code path can generate the command I wrote in the previous comment.

bmermet: I'm not sure I understand what you mean :S 
On Windows the command that gets executed is:
```set "GOOS=windows" && set "GOARCH=386"&& go build -a -o trace-agent-windows-386.exe ...```
And on other platforms it is:
```GOOS=windows GOARCH=386&& go build -a -o trace-agent-windows-386.exe ...```

Both commands are working and produce windows binaries, the difference is due to the way environment variables are set in Windows and Unix.

Can you elaborate on what may not be working?

PR #324, opened 2017-10-18
[trace] Moving quantization before truncation of spans

ryplo:
This is to ensure that strings (ex. query strings) stored in certain fields, such as `resources` are still parseable and are quantized properly. 
Tested in staging and prod

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -223,6 +223,7 @@ func (a *Agent) Process(t model.Trace) {
 
 	for i := range t {
 		t[i] = quantizer.Quantize(t[i])
+		t[i].Truncate()
 	}
 
 	pt := processedTrace{
--- a/model/truncator_test.go
+++ b/model/truncator_test.go
@@ -1,0 +1,66 @@
+package model
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTruncateResourcePassThru(t *testing.T) {
+	s := testSpan()
+	before := s.Resource
+	s.Truncate()
+	assert.Equal(t, before, s.Resource)
+}
+
+func TestTruncateLongResource(t *testing.T) {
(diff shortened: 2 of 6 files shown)


palazzem:
@ufoot I think it's safe to ship that huge improvement together with the distributed sampling. If you don't have strong opinion, I'm going to add that one to our 5.18.2 milestone.

agent/agent.go
```diff
@@ -178,6 +178,7 @@ func (a *Agent) Process(t model.Trace) {
 
 	for i := range t {
 		t[i] = quantizer.Quantize(t[i])
+		t[i].Truncate()
```
ufoot: I'm generally speaking in for this, the only thing I'm worried about is `Quantize` eating lots of CPU when the resource is very long. We have a protection against this, the watchdog is going to drop traces at pre-sampling stage if this happens. I think it's fine to ship this as is, but let's keep an eye on performance, not saying we should not do it, just, pay attention to side effects.

model/truncator_test.go
```diff
@@ -0,0 +1,66 @@
+package model
+
```
ufoot: I think we lack a test here, something that would test that say `UPDATE blogpost SET comment='<...something possibly huge, at least longer than 5000 char...>' WHERE blogpost_id=42;` and see how that ends. We should end up with the right quantization, which shows the `WHERE` clause. My understanding is that the purpose of this PR is to treat that case, because before, truncation would have removed the `WHERE` clause before anything else. I might have missed that test, or an equivalent, if that's the case, just point me to it.

ryplo: Discussed with Christian - decided to add a test mimicking agent behaviour to test joined effort of quantizer and truncator.

agent/agent.go
ryplo: Good point! Will keep that in mind

PR #329, opened 2017-10-27
Enable trace agent for windows

derekwbrown:
Implements trace agent as a windows service.


agent/main_windows.go
```diff
@@ -0,0 +1,320 @@
...
+				case svc.Interrogate:
+					changes <- c.CurrentStatus
+					// Testing deadlock from https://code.google.com/p/winsvc/issues/detail?id=4
+					time.Sleep(100 * time.Millisecond)
+					changes <- c.CurrentStatus
+				case svc.Stop, svc.Shutdown:
+					close(exit)
+					return
+				default:
+					elog.Error(1, fmt.Sprintf("unexpected control request #%d", c))
```
bmermet: As @LotharSee mentioned in #330. Should we log receiving unexpected control requests as errors since it's not causing any error in the agent? We don't want the user troubleshooting it if there is nothing wrong.

LotharSee: Yup if they are not actual errors, no reason to scare the user for that.
If we things these make sense to be in the logs for debug/support purpose, then we should make then `info`. Or `warn` if we though our `switch` was exhaustive.

agent/main_windows.go
```diff
@@ -0,0 +1,319 @@
...
+type myservice struct{}
+
+func (m *myservice) Execute(args []string, r <-chan svc.ChangeRequest, changes chan<- svc.Status) (ssec bool, errno uint32) {
+	const cmdsAccepted = svc.AcceptStop | svc.AcceptShutdown
+	changes <- svc.Status{State: svc.StartPending}
+	changes <- svc.Status{State: svc.Running, Accepts: cmdsAccepted}
+
+	exit := make(chan struct{})
+
+	go runAgent(exit)
```
palazzem: With the current implementation if the Agent runs in a goroutine it means that the control loop is the one that ends the process, "killing" all remaining go routines (so the agent). In our case, we start flushing Agent buffers when it receives the `exit` signal, but since the control loop is faster to return, it's not deterministic if buffers are flushed or not. This means that our shutdown may not be graceful.

Probably we should revert to the other way around? https://github.com/DataDog/datadog-trace-agent/pull/329/commits/0feeba7f95f89ff53d9fdf2a4af6dd04baf1b416#diff-0c5f36551698accfa0720833c46f618fR64

agent/main_windows.go
```diff
@@ -0,0 +1,319 @@
...
+		select {
+		case c := <-r:
+			switch c.Cmd {
+			case svc.Interrogate:
+				changes <- c.CurrentStatus
+				// Testing deadlock from https://code.google.com/p/winsvc/issues/detail?id=4
+				time.Sleep(100 * time.Millisecond)
+				changes <- c.CurrentStatus
+			case svc.Stop, svc.Shutdown:
+				elog.Info(0x40000006, ServiceName)
```
bmermet: What is the meaning of these hex values used for logging ?

PR #330, opened 2017-10-30
Graceful shutdown for Windows Service

bmermet:
This PR extracts the logic of providing an exit channel out of the `runAgent` function, that way we can have different exit conditions:
- SIGINT or SIGTERM on unix
- service STOP or SHUTDOWN when it is running as a windows service

### DIFF
--- a/agent/main_windows.go
+++ b/agent/main_windows.go
@@ -11,6 +11,7 @@ import (
 
 	_ "net/http/pprof"
 
+	"github.com/DataDog/datadog-trace-agent/watchdog"
 	log "github.com/cihub/seelog"
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/debug"
@@ -61,7 +62,6 @@ func (m *myservice) Execute(args []string, r <-chan svc.ChangeRequest, changes c
 	exit := make(chan struct{})
 
 	go func() {
-	loop:
 		for {
 			select {
 			case c := <-r:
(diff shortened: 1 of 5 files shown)


agent/main_windows.go
```diff
@@ -59,28 +58,30 @@ func (m *myservice) Execute(args []string, r <-chan svc.ChangeRequest, changes c
...
-				changes <- c.CurrentStatus
-			case svc.Stop, svc.Shutdown:
-				// FIXME need a way to stop!
-				//app.StopAgent()
-				break loop
-			default:
-				elog.Error(1, fmt.Sprintf("unexpected control request #%d", c))
+	exit := make(chan struct{})
+
+	go func() {
```
palazzem: This should not have any implication. Something we're missing @talwai?

agent/main_windows.go
```diff
@@ -59,28 +59,30 @@ func (m *myservice) Execute(args []string, r <-chan svc.ChangeRequest, changes c
...
+				case svc.Interrogate:
+					changes <- c.CurrentStatus
+					// Testing deadlock from https://code.google.com/p/winsvc/issues/detail?id=4
+					time.Sleep(100 * time.Millisecond)
+					changes <- c.CurrentStatus
+				case svc.Stop, svc.Shutdown:
+					close(exit)
+					return
+				default:
+					elog.Error(1, fmt.Sprintf("unexpected control request #%d", c))
```
LotharSee: What could these other command be? Are they all actual errors?

bmermet: These are not error indeed, only signals that we don't handle. I'll add your comment in the bigger PR about making the trace-agent work properly as a Windows Service.

PR #332, opened 2017-10-31
Add commit in default version number

bmermet:
This will make it possible to identify precisely which build is running from the version number.


bmermet:
Not needed since @LotharSee changed the CI.

PR #335, opened 2017-11-10
Change default log path on Windows

bmermet:
The previous default log path assumed that the agent was running on a Unix system. This PR fixes that.

### DIFF
--- a/config/agent_windows.go
+++ b/config/agent_windows.go
@@ -1,0 +1,4 @@
+package config
+
+// DefaultLogFilePath is where the agent will write logs if not overriden in the conf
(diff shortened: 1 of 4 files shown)


config/agent_windows.go
```diff
@@ -0,0 +1,4 @@
+package config
+
+// DefaultLogFilePath is where the agent will write logs if not overriden in the conf
+const DefaultLogFilePath = "c:\\programdata\\datadog\\logs\\trace-agent.log"
```
palazzem: @derekwbrown can you confirm?

derekwbrown: lgtm

PR #336, opened 2017-11-27
agent: introduce a TransactionWriter

talwai:
This PR introduces a transaction writer to the agent. It flushes
a sparse representation of _every_ unsampled trace seen by the agent,
as long as that trace passes some heuristics that deem it a user-generated
"transaction".

Two flags are exposed in the config:
- `UseTransactionAnalyzer`: enable transaction filtering and flushing. default: false
- `AnalyzeWebTransactions`: consider root spans of type `http` to be transactions, and eligible for flushing. default: false

The TransactionWriter exposes two mechanisms for flushing the data
- LogAgentFlusher: flush to tcp://localhost:10520 , where a logs agent is expected to be listening. (default
- IntakeFlusher: flush to the datadog backend over HTTP

Currently the payload is represented in the protobuf message `TransactionPayload`. This is subject to change
and we might want to just use an identical format to that of the existing AgentPayload

We ensure no duplicate spans by only queueing transactions for flush _after_ the agent makes its sampling decision

So the agent sends two streams of data:
- `AgentPayload`s containing sampled traces and Stats
- `TransactionPayloads` containing the transactions extracted from those traces that _do not_ survive sampling

TODO:
 - configure flush endpoint
 - configure analyzed operations by name and service. optionally specify a sample rate for analyzed transactions
 - tests


LotharSee:
Closing as it got replaced with https://github.com/DataDog/datadog-trace-agent/pull/346

PR #337, opened 2017-11-29
[pidfile] add pidfile if so specified

truthbk:
This will help with system-level orchestration.

depends on: `github.com/DataDog/datadog-agent/pkg/pidfile`

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -15,6 +15,7 @@ import (
 	log "github.com/cihub/seelog"
 	_ "net/http/pprof"
 
+	"github.com/DataDog/datadog-agent/pkg/pidfile"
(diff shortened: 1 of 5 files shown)


agent/main.go
```diff
@@ -126,6 +128,20 @@ func runAgent(exit chan struct{}) {
...
 	}
 
+	if !opts.info && opts.pidfilePath != "" {
+		err := pidfile.WritePID(opts.pidfilePath)
+		if err != nil {
+			log.Errorf("Error while writing PID file, exiting: %v", err)
+			os.Exit(1)
+		}
+
+		log.Infof("pid '%d' written to pid file '%s'", os.Getpid(), opts.pidfilePath)
```
clutchski: this should be below the defer.

PR #340, opened 2017-12-05
Replace Trace []Span with []*Span

LotharSee:
Another step before the global introduction of protobuf.

By using references it will simplify the code migration (as generated protobuf structs also include themselves by reference).

That should also avoid various struct copies.

Notes:

- for `model/trace_gen.go`, I simply ran `msgp -file=model/trace.go -marshal=false`. That should be enough as `trace_gen.go` isn't a file we manually modified. I also checked to simply regenerate the file without the `Trace` change and I got exactly the same file as before so the generation method is good. 
- `Quantize` is now in-place.
- The PR is based on the protobuf PR ; I'll merge them together once I've tested them enough.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -64,10 +64,16 @@ PACKAGES = %w(
 )
 
 EXCLUDE_LINT = [
-    'model/services_gen.go',
-    'model/trace_gen.go',
-    'model/span_gen.go',
-    'model/span.pb.go',
+  'model/services_gen.go',
+  'model/trace_gen.go',
+  'model/span_gen.go',
+  'model/span.pb.go',
+]
+
+MSGP_MODELS = [
+  'model/span.pb.go',
+  'model/trace.go',
+  'model/services.go',
 ]
 
 task :default => [:ci]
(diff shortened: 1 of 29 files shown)


Rakefile
```diff
@@ -170,7 +176,17 @@ task :err do
...
 
+desc "Regenerate protobuf files"
 task :protobuf do
   # be sure to have protobuf 3.x and go vendor installed
   sh "protoc -I=model -I=vendor --gogofaster_out=model model/*.proto"
 end
+
+desc "Regenerate messagepack files"
+task :msgp do
+  # Actually, don't use it since we modified manually the generated files for
```
ufoot: Nitpicking, but if we don't use this, a command-like example such as:
```ruby
puts "don't use it since we modified manually the generated files"
puts "to run it manually: msgp -file=model/span.pb.go -marshal=false"
```
would do the job and get a higher probability not to disappear at next refactoring.

model/trace_gen.go
```diff
@@ -4,24 +4,37 @@ package model
...
 	} else {
-		(*z) = make(Trace, zbai)
+		(*z) = make(Trace, xsz)
 	}
-	for zbzg := range *z {
-		err = (*z)[zbzg].DecodeMsg(dc)
-		if err != nil {
-			return
+	for bzg := range *z {
+		if dc.IsNil() {
```
ufoot: OK, so that's the big difference I guess.

PR #341, opened 2017-12-12
Always drain requests before responding

bmermet:
This PR targets a bug we've observed where sometimes the agent will send a HTTP 200 response before all the request's body has been received, which leads to the connection being reset and dropping traces.

### DIFF
--- a/agent/receiver.go
+++ b/agent/receiver.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"sort"
@@ -195,6 +196,7 @@ func (r *HTTPReceiver) replyTraces(v APIVersion, w http.ResponseWriter) {
 // handleTraces knows how to handle a bunch of traces
 func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *http.Request) {
 	if !r.preSampler.Sample(req) {
+		io.Copy(ioutil.Discard, req.Body)
 		HTTPOK(w)
 		return
 	}
--- a/agent/receiver_test.go
+++ b/agent/receiver_test.go
@@ -4,9 +4,12 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"strconv"
+	"sync"
 	"testing"
 	"time"
 
@@ -14,6 +17,7 @@ import (
 	"github.com/DataDog/datadog-trace-agent/fixtures"
 	"github.com/DataDog/datadog-trace-agent/info"
 	"github.com/DataDog/datadog-trace-agent/model"
+	"github.com/DataDog/datadog-trace-agent/sampler"
 	"github.com/stretchr/testify/assert"
 	"github.com/tinylib/msgp/msgp"
 )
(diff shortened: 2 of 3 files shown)


agent/receiver.go
```diff
@@ -195,6 +196,7 @@ func (r *HTTPReceiver) replyTraces(v APIVersion, w http.ResponseWriter) {
 // handleTraces knows how to handle a bunch of traces
 func (r *HTTPReceiver) handleTraces(v APIVersion, w http.ResponseWriter, req *http.Request) {
 	if !r.preSampler.Sample(req) {
+		io.Copy(ioutil.Discard, req.Body)
```
palazzem: 👍

fixtures/trace.go
```diff
@@ -83,13 +83,28 @@ func RandomTrace(maxLevels, maxSpans int) model.Trace {
...
+		// Calculate a trace ID which is predictable (this is why we seed)
+		// but still spreads on a wide spectrum so that, among other things,
+		// sampling algorithms work in a realistic way.
+		traceID := r.Uint64()
+
 		trace := model.Trace{}
 		for j := 0; j < size; j++ {
-			trace = append(trace, GetTestSpan())
+			span := GetTestSpan()
+			if realisticIDs {
```
ufoot: I introduced this because I noticed in some tests, when in debug mode, you'd see many messages in the log saying traces had been dropped because they had spans with the same ID. And, this is legit. It's a nitpick, but not having realistic traces makes tests less realistic and so they might miss interesting real-world cases.

agent/receiver_test.go
```diff
@@ -519,17 +523,80 @@ func TestHandleTraces(t *testing.T) {
 		ts, ok := rs.Stats[info.Tags{Lang: lang}]
 		assert.True(ok)
 		assert.Equal(int64(20), ts.TracesReceived)
-		assert.Equal(int64(57622), ts.TracesBytes)
+		assert.Equal(int64(59222), ts.TracesBytes)
```
ufoot: Slight increase because trace IDs and span IDs are longer. It's still predictable (seeded random) so no flakiness.

agent/receiver_test.go
```diff
@@ -519,17 +523,80 @@ func TestHandleTraces(t *testing.T) {
...
 	}
 	// make sure we have all our languages registered
 	assert.Equal("C#|go|java|python|ruby", receiver.Languages())
 }
 
+// chunkedReader is a reader which forces partial reads, this is required
+// to trigger some network related bugs, such as body not being read fully by server.
+// Without this, all the data could be read/written at once, not triggering the issue.
+type chunkedReader struct {
```
ufoot: That, was the tricky part. Without this it's impossible to reproduce the bug.

agent/receiver_test.go
```diff
@@ -519,17 +523,80 @@ func TestHandleTraces(t *testing.T) {
...
+	defer server.Close()
+	url := server.URL + "/v0.4/traces"
+
+	// Make sure we use share clients, and they are reused.
+	client := &http.Client{Transport: &http.Transport{
+		MaxIdleConnsPerHost: 100,
+	}}
+	for i := 0; i < 3; i++ {
+		wg.Add(1)
+		go func() {
```
ufoot: I'm not totally sure this goroutine thing is totally mandatory, but it's more realistic anyway and does not cost *that* much.

agent/receiver_test.go
```diff
@@ -519,17 +523,80 @@ func TestHandleTraces(t *testing.T) {
...
+
+	receiver := NewHTTPReceiver(conf, dynConf)
+	server := httptest.NewServer(http.HandlerFunc(receiver.httpHandleWithVersion(v04, receiver.handleTraces)))
+
+	defer server.Close()
+	url := server.URL + "/v0.4/traces"
+
+	// Make sure we use share clients, and they are reused.
+	client := &http.Client{Transport: &http.Transport{
+		MaxIdleConnsPerHost: 100,
```
ufoot: Increasing `MaxIdleConnsPerHost` is common practice to ensure the connection is reused.

bmermet:
Great job for finding a way to trigger this bug 👍 

Though I'm not sure that it is exactly the same bug we've been observing in real use cases. The bug triggered in this test disappears if we don't send requests concurrently from several goroutines, while the behavior of `dd-trace-go` is to flush traces only from a single goroutine. On top of that the error message is different, here it is "HTTP/1.x transport connection broken: write tcp 127.0.0.1:60168->127.0.0.1:60161: write: broken pipe" while what we observed in the real world bug was: "HTTP/1.x transport connection broken: write tcp 127.0.0.1:49133->127.0.0.1:8126: write: connection reset by peer".

The two bugs look extremely similar and certainly share a common cause since this patch fixes both. I would like to understand better what the exact cause of this bug is but in the meantime I think this test is good enough to validate the fix and ship it.

fixtures/trace.go
```diff
@@ -83,13 +83,28 @@ func RandomTrace(maxLevels, maxSpans int) model.Trace {
 
 // GetTestTrace returns a []Trace that is composed by ``traceN`` number
 // of traces, each one composed by ``size`` number of spans.
-func GetTestTrace(traceN, size int) model.Traces {
+func GetTestTrace(traceN, size int, realisticIDs bool) model.Traces {
```
bmermet: What are the cases when we do not want to have these "realisticIDs" ?

ufoot: Tests which expect the traceID to be 42 eg : https://github.com/DataDog/datadog-trace-agent/blob/fbc904c22826c37145df1aee40efed8e2839944f/agent/receiver_test.go#L198

PR #343, opened 2017-12-18
Introduce new writers to API v0.2

LotharSee:
Split the Writer into 3 different writers, writing to 3 new endpoints:

- `/traces`, `TraceWriter`: new protobuf endpoint. Only contains traces (and soon transactions).
- `/stats`, `StatsWriter`: similar to the previous AgentPayload but without the traces.
- `/services`, `ServiceWriter`: similar to before. But now a writer on its own.

They are split as they differ in terms of encoding, compression, replay logic, ...

To make that work properly, there is some heavy rewiring.

- The trigger of a flush no longer comes from `agent/agent.go`. Traces and services are sent to their writer as soon as processed. The concentrator generates and flush StatsBuckets itself to the writer. Then writer are free to send it over the wire when/how they want (allowing future different flush conditions / replay features).
- Revamp `Endpoint` API to be much simpler, simply writing bytes with HTTP headers. Also extract the client/proxy configuration.
- New `TracePayload` is our first protobuf payload. For now it isn't compressed, we should benchmarks options before picking a compression.
- Many lines changed around `info` to make it work with the new metrics / writers.

TODOs before being able to merge it:

- [ ] figure out encoding for the trace endpoint
- [ ] heavy real-world testing with compatible server trace api


agent/agent.go
```diff
@@ -60,41 +60,62 @@ type Agent struct {
 // NewAgent returns a new Agent object, ready to be started
 func NewAgent(conf *config.AgentConfig, exit chan struct{}) *Agent {
 	dynConf := config.NewDynamicConfig()
+
+	// inter-component channels
+	rawTraceChan := make(chan model.Trace, 5000) // about 1000 traces/sec for 5 sec
```
AlexJF: Should we take this opportunity to remove some copies here too (chan *model.Trace)?

agent/concentrator.go
```diff
@@ -28,11 +37,52 @@ func NewConcentrator(aggregators []string, bsize int64) *Concentrator {
...
 		buckets:     make(map[int64]*model.StatsRawBucket),
+
+		exit:   make(chan struct{}),
+		exitWG: &sync.WaitGroup{},
 	}
 	sort.Strings(c.aggregators)
 	return &c
 }
 
+// Start starts the writer.
```
AlexJF: starts the concentrator

agent/main.go
```diff
@@ -148,7 +148,7 @@ func runAgent(exit chan struct{}) {
 
 	if opts.info {
 		if err := info.Info(os.Stdout, agentConf); err != nil {
-			// need not display the error, Info should do it already
+			os.Stdout.WriteString(fmt.Sprintf("failed to print info: %s\n", err))
```
AlexJF: Less "should" more action 👍  😄

agent/receiver.go
```diff
@@ -71,8 +71,6 @@ type HTTPReceiver struct {
 func NewHTTPReceiver(conf *config.AgentConfig, dynConf *config.DynamicConfig) *HTTPReceiver {
 	// use buffered channels so that handlers are not waiting on downstream processing
 	return &HTTPReceiver{
-		traces:     make(chan model.Trace, 5000), // about 1000 traces/sec for 5 sec
```
AlexJF: We should be passing these by argument. Technically this is a constructor function so whatever comes out should be working without needing to modify unexported fields (IMO)

agent/sampler.go
```diff
@@ -17,29 +16,28 @@ import (
...
+	lastFlush       time.Time
+
+	// actual implementation of the sampling logic
 	engine sampler.Engine
 }
 
-// NewScoreEngine creates a new empty sampler ready to be started
-func NewScoreEngine(conf *config.AgentConfig) *Sampler {
+// NewScoreSampler creates a new empty sampler ready to be started
+func NewScoreSampler(conf *config.AgentConfig) *Sampler {
```
AlexJF: Again, it kinda feels cleaner to me if we would pass `sampled` as an argument to the constructors.

agent/sampler.go
```diff
@@ -17,29 +16,28 @@ import (
...
-		traceCount:    0,
-		engine:        sampler.NewScoreEngine(conf.ExtraSampleRate, conf.MaxTPS),
+		engine: sampler.NewScoreEngine(conf.ExtraSampleRate, conf.MaxTPS),
 	}
 }
 
-// NewPriorityEngine creates a new empty distributed sampler ready to be started
-func NewPriorityEngine(conf *config.AgentConfig, dynConf *config.DynamicConfig) *Sampler {
+// NewPrioritySampler creates a new empty distributed sampler ready to be started
+func NewPrioritySampler(conf *config.AgentConfig, dynConf *config.DynamicConfig) *Sampler {
```
AlexJF: Ditto

agent/sampler.go
```diff
@@ -49,62 +47,65 @@ func (s *Sampler) Run() {
...
+		switch state := state.(type) {
+		case sampler.InternalState:
+			log.Debugf("%s: inTPS: %f, outTPS: %f, maxTPS: %f, offset: %f, slope: %f, cardinality: %d",
+				engineType, state.InTPS, state.OutTPS, state.MaxTPS, state.Offset, state.Slope, state.Cardinality)
+
+			// publish through expvar
+			switch s.engine.(type) {
+			case *sampler.ScoreEngine:
+				info.UpdateSamplerInfo(info.SamplerInfo{EngineType: engineType, Stats: stats, State: state})
+			case *sampler.PriorityEngine:
```
AlexJF: Would be cleaner as methods of engine. Let method resolution do its work.

model/stats_payload.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"bytes"
+	"compress/gzip"
+	"encoding/json"
+)
+
+// StatsPayload represents the payload to be flushed to the stats endpoint
+type StatsPayload struct {
+	HostName string        `json:"hostnamed"`
```
talwai: typo `hostname` ?

writer/datadog_endpoint.go
```diff
@@ -0,0 +1,66 @@
...
+	req, err := http.NewRequest("POST", url, bytes.NewBuffer(payload))
+
+	// If the request cannot be created, there is no point in trying again later,
+	// it will always yield the same result.
+	if err != nil {
+		// atomic.AddInt64(&ae.stats.TracesPayloadError, 1)
+		return err
+	}
+
+	// Set API key in the header and issue the request
```
talwai: i realize this is copy-paste, but the comment disagrees with the code. We still send the api key as a query param rather than through headers.

Since the code is being touched maybe leave a TODO to _really_ move it to HTTP headers. The new backend endpoints should support this as the main way to auth.

agent/agent.go
LotharSee: `model.Trace` being a slice it isn't such a big deal.

https://github.com/golang/go/wiki/CodeReviewComments#receiver-type
But it falls under the:
> If the receiver is a struct, array or slice and any of its elements is a pointer to something that might be mutating, prefer a pointer receiver, as it will make the intention more clear to the reader.

So it is mostly to make the code more explicit ; I'll leave it as a comment.

AlexJF: True about it just being a slice. Hadn't noticed. Comment was mostly because sampledTraceChan indeed uses `*model.Trace` and it felt a bit weird to have both styles.

LotharSee: Totally ; I'm not sure yet about what is the best there / I'd rather take enough time to make the right decision. Left it as a comment, it can be a later task to review these questions across the entire Agent.

writer/datadog_endpoint.go
LotharSee: Right, I moved it to HTTP headers (the same as https://github.com/DataDog/datadog-agent/blob/master/pkg/forwarder/forwarder.go#L66 )

agent/receiver.go
LotharSee: Very good point! It both makes the code more explicit and avoid sneaky bugs. Updated.

agent/receiver_test.go
```diff
@@ -29,18 +29,35 @@ var headerFields = map[string]string{
...
-	assert := assert.New(t)
+func NewTestReceiverFromConfig(conf *config.AgentConfig) *HTTPReceiver {
+	dynConf := config.NewDynamicConfig()
+
+	rawTraceChan := make(chan model.Trace, 5000)
+	serviceChan := make(chan model.ServicesMetadata, 50)
+	receiver := NewHTTPReceiver(conf, dynConf, rawTraceChan, serviceChan)
 
+	receiver.traces = rawTraceChan
+	receiver.services = serviceChan
```
AlexJF: Redundant with arguments being passed to constructor.

model/trace.proto
```diff
@@ -0,0 +1,12 @@
...
+
+package model;
+
+import "span.proto";
+
+message APITrace {
+	uint64 traceID = 1;
+	repeated Span spans = 2;
+	int64 startTime = 6;
+    int64 endTime = 7;
```
AlexJF: Nit: extra space

LotharSee:
Yes that's the goal: I tried to keep writers as independent/uncoupled as possible ; that way it won't constraint us when we will have more writer-specific code (different buffering, retrying, logging, metrics, ...). After that we will be able to see what's in common and share it! (even if, sometimes, we can't avoid some dupe in Go).

writer/service_writer.go
```diff
@@ -0,0 +1,169 @@
...
+	atomic.StoreInt64(&w.stats.Services, int64(len(serviceBuffer)))
+
+	data, err := model.EncodeServicesPayload(serviceBuffer)
+	if err != nil {
+		log.Errorf("encoding issue: %v", err)
+		return
+	}
+
+	headers := map[string]string{
+		languageHeaderKey: strings.Join(info.Languages(), "|"),
```
LotharSee: Currently updating our backend, as this header will be enough on the stats payload, not needed for service and traces.

PR #344, opened 2017-12-22
Async flushing of writer payloads with jittery exponential retrying

AlexJF:
Includes:
* More modular structure for extra code reuse and cleaner testing.
* Readdition of retriable errors.
* Async flushing of payloads handled by a PayloadSender.
* Addition of exponential jittery backoff timer used by default by the new PayloadSender.

Working on top of: https://github.com/DataDog/datadog-trace-agent/pull/343

### DIFF
--- a/backoff/backoff.go
+++ b/backoff/backoff.go
@@ -1,0 +1,84 @@
+package backoff
+
+import "time"
+
+// Timer represents a timer that implements some backOff strategy that can adapt to number of schedulings.
+type Timer interface {
+	ScheduleRetry(err error) (int, time.Duration)
+	CurrentDelay() time.Duration
+	NumRetries() int
+	ReceiveTick() <-chan time.Time
+	Reset()
+	Stop()
+}
+
+// DelayProvider is a function that takes the current numRetries and last error and returns the delay until next retry.
+type DelayProvider func(numRetries int, err error) time.Duration
+
+// CustomTimer represents a backoff timer configured with a certain DelayProvider.
+type CustomTimer struct {
+	numRetries   int
+	currentDelay time.Duration
+
+	delayProvider DelayProvider
+
+	tickChannel chan time.Time
+	timer       *time.Timer
+}
+
+// NewCustomTimer creates a new custom timer using the provided delay provider.
+func NewCustomTimer(delayProvider DelayProvider) *CustomTimer {
+	return &CustomTimer{
+		delayProvider: delayProvider,
+		tickChannel:   make(chan time.Time),
+	}
+}
+
+// ScheduleRetry schedules the next retry tick according to the delay provider, returning retry num and retry delay.
+func (t *CustomTimer) ScheduleRetry(err error) (int, time.Duration) {
+	t.Stop()
+	t.currentDelay = t.delayProvider(t.numRetries, err)
+
+	t.timer = time.AfterFunc(t.currentDelay, func() {
+		t.tickChannel <- time.Now()
+	})
+
+	t.numRetries++
+
+	return t.numRetries, t.currentDelay
+}
+
+// CurrentDelay returns the delay of the current or last ticked retry.
+func (t *CustomTimer) CurrentDelay() time.Duration {
+	return t.currentDelay
+}
+
+// NumRetries returns the number of tries since this timer was last reset.
+func (t *CustomTimer) NumRetries() int {
+	return t.numRetries
+}
+
+// ReceiveTick returns a channel that will receive a time.Time object as soon as the previously scheduled retry ticks.
+func (t *CustomTimer) ReceiveTick() <-chan time.Time {
+	return t.tickChannel
+}
+
+// Reset stops and resets the number of retries counter of this timer.
+func (t *CustomTimer) Reset() {
+	t.Stop()
+	t.numRetries = 0
+	t.currentDelay = 0
+}
+
+// Stop prevents any current scheduled retry from ticking.
+func (t *CustomTimer) Stop() {
+	if t.timer != nil {
+		t.timer.Stop()
+	}
+}
+
+// Close cleans up the resources used by this timer. It cannot be reused after this call.
+func (t *CustomTimer) Close() {
+	t.Reset()
+	close(t.tickChannel)
+}
--- a/fixtures/trace.go
+++ b/fixtures/trace.go
@@ -69,7 +69,7 @@ func RandomTrace(maxLevels, maxSpans int) model.Trace {
 	t := model.Trace{RandomSpan()}
 
 	prevLevel := t
-	maxDepth := rand.Intn(maxLevels)
+	maxDepth := 1 + rand.Intn(maxLevels)
 
 	for i := 0; i < maxDepth; i++ {
 		if len(prevLevel) > 0 {
--- a/statsd/statsd.go
+++ b/statsd/statsd.go
@@ -7,9 +7,16 @@ import (
 	"github.com/DataDog/datadog-trace-agent/config"
 )
 
+// StatsClient represents a client capable of sending stats to some stat endpoint.
+type StatsClient interface {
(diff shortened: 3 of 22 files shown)


AlexJF:
[trace-agent: retrying payload flush should not be aligned](https://trello.com/c/67EoOfmP/1348-trace-agent-retrying-payload-flush-should-not-be-aligned)

statsd/statsd.go
```diff
@@ -7,9 +7,16 @@ import (
 	"github.com/DataDog/datadog-trace-agent/config"
 )
 
+// StatsClient represents a client capable of sending stats to some stat endpoint.
```
LotharSee: Good idea!

Looks like a small change which could be an easy/quick PR on its own (with all the `TestStatsClient` stuff).

fixtures/trace.go
```diff
@@ -69,7 +69,7 @@ func RandomTrace(maxLevels, maxSpans int) model.Trace {
 	t := model.Trace{RandomSpan()}
 
 	prevLevel := t
-	maxDepth := rand.Intn(maxLevels)
+	maxDepth := 1 + rand.Intn(maxLevels)
```
LotharSee: Nit: to keep `maxLevels` meaningful, prefer a `maxDepth  := max(1, maxLevels)`? Or maybe the actual implementation is such as maxDepth == 1 + maxLevels? (in that case the method comment is misleading).

backoff/backoff.go
```diff
@@ -0,0 +1,81 @@
+package backoff
```
LotharSee: Can you confirm that the `backoff` package isn't used yet?

fixtures/trace.go
AlexJF: `rand.Intn(x)` returns something in `[0, x[`. So for the random trace to actually be in the range `[1, x]` as it seems to be the intention from the method comment (since maxSpans is documented as inclusive, would seem strange to have one be exclusive and the other inclusive), we need to do the +1.

LotharSee: Good catch!

statsd/statsd.go
AlexJF: Good point. Should reduce some load from this PR!

writer/trace_writer.go
```diff
@@ -16,85 +16,224 @@ import (
...
+				continue
+			}
+			log.Trace("Handling new trace %v", trace)
+			if _, err := w.payloadConstructor.Add(trace); err != nil {
+				log.Error("error adding trace to payload", err)
+			}
 		case <-flushTicker.C:
-			w.Flush()
+			log.Debug("Flushing current traces")
+			w.payloadConstructor.ConstructAndSend()
```
LotharSee: The fact that both a ticker in the Writer and a call to `payloadConstructor.Add` can trigger a flush can make comprehension tricky (and not efficient as the tricker can trigger at moments the buffer is almost empty).

We probably want the "construct" decisions to be taken from a single place ; with the current model it seems simpler to have the ticker inside the payloadConstructor itself (and replace it with a `Now() - lastFlush > FlushPeriod` condition).

writer/payload.go
```diff
@@ -0,0 +1,340 @@
...
+// SendErrorHandler represents a callback that is called on failed sending of a payload.
+type SendErrorHandler func(payload *Payload, err error, stats PayloadStats)
+
+// PayloadSender represents an object capable of asynchronously sending payloads to some endpoint.
+type PayloadSender interface {
+	Start()
+	Run()
+	Stop()
+	Send(payload *Payload)
+	OnSendSuccess(handler SendSuccessHandler)
```
LotharSee: It might be more in the spirit of Go to use channels instead of callbacks here.

We could probably have a single channel if we propagate the code/status along the payload + err + stats.
But in practice, all of that is pretty similar (it is mostly about if we want to only communicate by channels or not).

writer/payload.go
```diff
@@ -0,0 +1,340 @@
...
+
+// sendOrQueue sends the provided payload or queues it if this sender is currently queueing payloads.
+func (s *QueuablePayloadSender) sendOrQueue(payload *Payload) error {
+	if payload == nil {
+		return nil
+	}
+
+	var err error
+
+	if !s.queuing {
```
LotharSee: Having the retry decision here might block us if we have writer-specific retry logics (like the support of 413).

An option is to have the Sender communicates back to the writer, in an async channel, the result of each flush (containing payload + source data + response code + info about retries) ; then the writer takes the retry (and potential writer-specific split and changes) decision and push it back to the Sender input (one or more payloads). It can mean a bit more dupe code for the common case of 5xx response, but not much more. Plus it allows us things like not replaying at all under some conditions (for the service metadata, we don't have to flush them all ; the latest is the only we care about).

Another complementary idea is to just remove the 413 case, either by capping the size of the serialized payload (if too big, split as many times as required before flushing) or by considering that the cap on the number of spans/traces is enough (then disable the 413 feature on the server-side).

AlexJF: Delegating all retry logic to writers would not only introduce dupe code, it would also make aligned retrying, retry limitation and total limits (say max bytes waiting for retry) a bit more complicated.

We seem to have 2 kinds of error with different needs:
* Some communication/temporary server issue (these would map to `RetriableError`).
* Invalid input errors like the 413 one.

Technically, the first should be able to be retried transparently by the Sender. No info is needed from the writer and this way sender can abstract all size/num/retry limit from writer.

Writers only need to weight in for the second class of errors since these are the ones the sender cannot possibly retry.

I suggest something along these lines:

* Add `MaxQueuedPayloads` parameter to `QueuablePayloadSender`. This gives us enough flexibility to support all our current cases (set `MaxQueuedPayloads` to 1 for service metadata and you get the only retry latest functionality).
* If something radically different is needed, a writer can always use a custom Sender on a case-by-case scenario while keeping the common cases clean and dupe-free.
* Add Promise-like responses to `PayloadSender.Send`:
  * `response.Then(...)` would run on successful sending.
  * `response.Catch(...)` would run on non-retriable error (either transport related like 413 or when we gave up on retrying at payload sender side). Here is where writer-specific logic should go to IMO. Split the payload, resend it as 2 new ones.

What do you think?

AlexJF: We could alsooffer a third, non-final promise 'resolution':

`response.Retry(...)` allowing a Writer to veto a retry from the sender. It would slightly complicate the sender queue if done asynchronously because right now we are keeping the queue in-order but shouldn't be too difficult to implement and would allow writers to collect total error stats without the current Sender.OnError callback.

writer/payload.go
AlexJF: Recoded with a monitor channel.

writer/payload.go
AlexJF: Added the MaxQueuedPayload parameter I had mentioned before. Promises would add too much complexity. Monitor channel should be sufficient provided that the writer is the only one listening on that channel.

statsd/statsd.go
talwai: are we still splitting this out?

AlexJF: Had forgotten about this sorry. Splitting it out will require some rebase gymnastics which might not make sense now if we are close to merging everything

backoff/backoff.go
```diff
@@ -0,0 +1,84 @@
...
```
gbbr: Would be amazing to add a package level doc (`backoff/doc.go`) explaining in more detail what the package does and what backoff is.

gbbr: Looks like this comment was completely ignored. Why?

AlexJF: Sorry Gabriel, my bad. Completely missed it. I'll add it in a following PR where I'll also add some testing for the transaction code that got added to the trace_writer.

clutchski: @gbbr maybe he wants you to (ahem) backoff 🥁

PR #345, opened 2017-12-27
[dual sampling] using both score and priority sampler for priority 0 traces

ufoot:
In the case of priority 0 traces, send traces to both samplers. This gives
a second chance to priority 0 traces to be picked up by the score sampler.
Also adds stats for -1 sampling priority traces.


config/agent.go
```diff
@@ -324,8 +326,11 @@ APM_CONF:
 	if v, e := conf.GetFloat("trace.sampler", "max_traces_per_second"); e == nil {
 		c.MaxTPS = v
 	}
-	if v := strings.ToLower(conf.GetDefault("trace.sampler", "priority_sampling", "")); v == "yes" || v == "true" {
-		c.PrioritySampling = true
```
ufoot: This probably never worked, the default was `true`, and we were overriding it to `true`...

agent/agent.go
```diff
@@ -174,22 +174,39 @@ func (a *Agent) Process(t model.Trace) {
...
+			samplers = append(samplers, a.ScoreEngine)
+		}
+	} else {
+		// Send traces to one single engine, either Priority or Score
+		if ok && a.PriorityEngine != nil {
+			samplers = []*Sampler{a.PriorityEngine}
+		} else {
+			samplers = []*Sampler{a.ScoreEngine}
+		}
+	}
```
AlexJF: Following looks cleaner IMO (everything related to priority grouped together and ok -> hasPriority):

```
samplers := make([]*Sampler, 0, 2)
priority, hasPriority := root.Metrics[samplingPriorityKey]
if hasPriority && a.PriorityEngine != nil {
  samplers = append(samplers, a.PriorityEngine)

  if a.conf.ScorePriority0Traces && priority == 0 {
    samplers = append(samplers, a.ScoreEngine)
  }
} else {
  samplers = append(samplers, a.ScoreEngine)
}
```

config/agent.go
AlexJF: Good catch!

agent/agent.go
AlexJF: Just realized that to be a 100% match, the a.PriorityEngine should go to a separate if

```
samplers := make([]*Sampler, 0, 2)
priority, hasPriority := root.Metrics[samplingPriorityKey]
if hasPriority {
  if a.PriorityEngine != nil {
    samplers = append(samplers, a.PriorityEngine)
  }

  if a.conf.ScorePriority0Traces && priority == 0 {
    samplers = append(samplers, a.ScoreEngine)
  }
} else {
  samplers = append(samplers, a.ScoreEngine)
}
```

AlexJF: Hmm actually none of them are equivalent to yours now that I look closer :/ Maybe there's no way around it :P

agent/agent.go
```diff
@@ -174,22 +174,39 @@ func (a *Agent) Process(t model.Trace) {
...
-	// We choose the sampler dynamically, depending on trace content,
-	// it has a sampling priority info (wether 0 or 1 or more) we respect
-	// this by using priority sampler. Else, use default score sampler.
-	s := a.ScoreEngine
+	var samplers []*Sampler
+	priority, ok := root.Metrics[samplingPriorityKey]
+	if a.conf.ScorePriority0Traces {
+		// Send traces to possibly several score engines.
+		if ok && a.PriorityEngine != nil {
+			// If Priority is defined, send to priority sampling, regardless of priority value.
```
talwai: Should the count of "priority < 0" traces affect the applied sample rate in the client? Some apps will routinely generate traces which the user will want to ignore via the priority flag e.g. healthchecks, debug endpoints, whatever. Should the frequency of such traces still influence the sample rate?

agent/agent.go
ufoot: Yes, I also spent some time on it initially, the "problem" is that in one case we do an `if ... if ...` and in the other one an `if ... else ...`. Thanks for the thorough review.

agent/agent.go
ufoot: Good point. Indeed, since the client lib can only influence the `0/-1` choice, it totally makes sense it takes decisions only on those ones. But then, I'm thinking we could also exclude the `+2` traces. The problem is quite symmetrical here. In any case, I think this is really sampling logic, so should probably go to sampler package, easier to unit-test and bullet-proof rather than in this "high-level" code.

Also, I'm worried about how we can explain exactly what that `-1` does. Because, it can sound like it's "exclude this from tracing". When, it's really not. The stats are still computed for instance. That's the whole point of this `-1` thing I think, make sure there's no trace sampled, but still trace it. I'm saying this because we had similar issues whit the "disabled tracer" mode, which did not really disabled the tracer, but only the sending of traces to the agent.

I'm going to exclude `-1` and `+2` flagged traces from the count anyway.

sampler/prioritysampler.go
```diff
@@ -91,10 +91,19 @@ func (s *PriorityEngine) Sample(trace model.Trace, root *model.Span, env string)
...
 
+	samplingPriority := root.Metrics[samplingPriorityKey]
+
 	// Regardless of rates, sampling here is based on the metadata set
 	// by the client library. Which, is turn, is based on agent hints,
 	// but the rule of thumb is: respect client choice.
-	sampled := root.Metrics[samplingPriorityKey] > 0
+	sampled := samplingPriority > 0
+
+	if samplingPriority < 0 || samplingPriority > 1 {
```
ufoot: @talwai this is where I'd exclude the traces which should have no impact on sampling ratio. I had a second thought about it and really do think the `priority 2` traces should be excluded too. Not doing so would a) break the ratio for other traces (keeping less of them, when this is really about "decide to have some more") and b) give a false illusion that having 1 and 2 in the same bag does provide some fail-safe rate limiting. It does not. On a service sending 100 traces per second, with a max TPS of 10, the default target ratio would be 0.1. If we select 50% of them with priority set to 2, then we have 50 traces per sec no matter what max TPS is set to. So if we actually include priority 2 traces in the count, it will result in having only the one we selected, but also possibly impact other services, and in any case, there's no capping to 10 TPS. The big picture is: priority 2 traces are something on top of the regular sampling performed by the library, it's not included in it.

config/agent.go
```diff
@@ -171,10 +172,11 @@ func NewDefaultAgentConfig() *AgentConfig {
...
 
-		ExtraSampleRate:  1.0,
-		PreSampleRate:    1.0,
-		MaxTPS:           10,
-		PrioritySampling: true,
+		ExtraSampleRate:      1.0,
+		PreSampleRate:        1.0,
+		MaxTPS:               10,
+		PrioritySampling:     true,
+		ScorePriority0Traces: true,
```
LotharSee: As the goal is to simplify and have both samplers, I'd suggest to just drop both `PrioritySampling` and `ScorePriority0Traces` while always having both logics working. I see no case where these configuration should be used nor exposed to the user.

agent/agent.go
LotharSee: As suggested in the config.go comment, let's just have both mechanisms always on, it should greatly simplify the code here.

config/agent.go
talwai: +1

agent/agent.go
```diff
@@ -174,22 +164,30 @@ func (a *Agent) Process(t model.Trace) {
...
+		// gets the big picture and can set the sampling rates accordingly.
+		samplers = append(samplers, a.PriorityEngine)
+	}
+	if priority == 0 {
+		// Use score engine for traces with no priority or priority set to 0
+		samplers = append(samplers, a.ScoreEngine)
+	}
+
 	priorityPtr := &ts.TracesPriorityNone
-	if a.PriorityEngine != nil {
```
ufoot: No need to check for it now, it's always defined since we removed the `PrioritySampling` option

config/agent.go
ufoot: Useless after all ;)

ufoot:
Companion PRs:

* https://github.com/DataDog/dd-trace-py/pull/391
* https://github.com/DataDog/dd-trace-rb/pull/290
* https://github.com/DataDog/dd-trace-go/pull/137

PR #346, opened 2018-01-03
writer: add transaction stream to trace writer

talwai:
Add a stream for "analyzed transactions" into the trace writer. This stream captures top-level spans for faceted analysis. Transactions can be sampled by rate per service, with `1.0` indicating that we should stream every root span for a service.

Configured by:

```
[trace.analyzed_rate_by_service]
web = 1.0
rpc = 0.2
```

### DIFF
--- a/agent/sampler.go
+++ b/agent/sampler.go
@@ -72,23 +72,22 @@ func (s *Sampler) Add(t processedTrace) {
 	}
 
 	// inspect the WeightedTrace so that we can identify top-level spans
-	// Only top-level spans are eligible to be analyzed
 	for _, span := range t.WeightedTrace {
-		if analyzeRate, ok := s.analyzedRateByService[span.Service]; ok {
-			s.Analyze(span, analyzeRate)
-		}
+		s.Analyze(span)
 	}
 }
 
 // Analyze queues a span for analysis, applying any sample rate specified
 // Only top-level spans are eligible to be analyzed
-func (s *Sampler) Analyze(span *model.WeightedSpan, sampleRate float64) {
+func (s *Sampler) Analyze(span *model.WeightedSpan) {
 	if !span.TopLevel {
 		return
 	}
 
-	if sampler.SampleByRate(span.TraceID, sampleRate) {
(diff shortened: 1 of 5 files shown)


agent/sampler.go
```diff
@@ -61,8 +69,23 @@ func (s *Sampler) Add(t processedTrace) {
 	s.totalTraceCount++
 	if s.engine.Sample(t.Trace, t.Root, t.Env) {
 		s.keptTraceCount++
+		t.Root.Metrics[spanAnalyzedTransaction] = 1
```
AlexJF: Why are we tagging the transaction as having an analyzed span if we only decide whether we analyze it later on? If this is not what is happening, then different const naming or some comment here would clarify it a bit better.

LotharSee: Yeah this is misleading. I think its purpose is to indicate to the log backend (when this span will reach it) that the related trace got sampled. (therefore it can adjust its display vs. the log transaction which doesn't have matching sampled trace).


However that's not the right place to put that as the server sampling could drop this trace.

The logic should be to only analyze in case we didn't sample it ; then let the backend run the analyze on server-side sampled traces.

agent/sampler.go
```diff
@@ -67,24 +65,30 @@ func (s *Sampler) Run() {
...
 		s.keptTraceCount++
-		t.Root.Metrics[spanAnalyzedTransaction] = 1
 		s.sampled <- &t.Trace
 	}
 
-	s.Analyze(t.Root)
+	// inspect the WeightedTrace so that we can identify top-level spans
+	// Only top-level spans are eligible to be analyzed
+	for _, span := range t.WeightedTrace {
+		if analyzeRate, ok := s.analyzedRateByService[span.Service]; ok {
```
AlexJF: LGTM. I'd perhaps move this if and map access inside Analyze, after the `TopLevel` if. That way all "should I analyze this?" decisions are close to each other (and we save one map access per non-root span).

talwai: Yes, good spot

PR #349, opened 2018-01-23
Allow configuration of trace, stats and service writers.

AlexJF:
This PR exposes most of the configurations surrounding the new senders and queues for trace, stats and service writers.

This way, these details can be tuned without need for a recompilation.

### DIFF
--- a/agent/trace-agent.ini
+++ b/agent/trace-agent.ini
@@ -71,3 +71,86 @@ oldest_span_cutoff_seconds=30
 receiver_port=8126
 # how many unique connections to allow during one 30 second lease period
 connection_limit=2000
+
+###################################################
+# Trace writer - combines traces into payloads
+# that are sent to Datadog servers
+###################################################
+[trace.writer.traces]
+# Information about different traces is batched together for performance reasons. This setting controls the maximum
+# amount of spans inside one of these batches. Bigger values lead to fewer but bigger payloads.
+max_spans_per_payload=1000
+# Period with which the agent forces flushes of payloads. In case the previous limit is not reached within this
+# period, a payload with less than max_spans_per_payload will be sent.
+flush_period_seconds=5
+# Period with which the agent will send updated trace writer statistics to DD servers.
+update_info_period_seconds=60
+# For how many seconds should we keep traces queued up in case we are unable to send them to DD servers.
+# A value <= 0 disables this limit
+# Default: 20 minutes
+queue_max_age_seconds=1200
+# Maximum amount of trace data we can have queued up in case we are unable to send them to DD servers.
+# A value <= 0 disables this limit
+# Default: 256MB
+queue_max_bytes=268435456
+# Maximum amount of trace payloads we can have queued up in case we are unable to send them to DD servers.
+# A value <= 0 disables this limit
+# Default: Unlimited
+queue_max_payloads=-1
+# When we are unable to send traces to DD servers, retries will be attempted with various delays. Each retry delay
+# is determined by a full-jitter exponential backoff that performs the following computation:
+# delay = Random(0, Min(MaxDuration, (GrowthBase ^ NumRetries) * Base))
+# The idea is that each consecutive retry for the same payload will, on average, take longer and longer until reaching
+# MaxDuration and will also be desynchronized with retrial processes of other agents.
+# The following parameters enable you to configure MaxDuration, Base and GrowthBase, respectively.
+exp_backoff_max_duration_seconds=120
+exp_backoff_base_milliseconds=200
+exp_backoff_growth_base=2
+
+###################################################
+# Service writer - extracts service info from
+# traces and sends it to Datadog servers.
+###################################################
+[trace.writer.services]
+# Period with which the agent flushes service info to DD servers.
+flush_period_seconds=5
+# Period with which the agent will send updated service writer statistics to DD servers.
+update_info_period_seconds=60
+# When we are unable to send service info to DD servers, retries will be attempted with various delays. Each retry delay
+# is determined by a full-jitter exponential backoff that performs the following computation:
+# delay = Random(0, Min(MaxDuration, (GrowthBase ^ NumRetries) * Base))
+# The idea is that each consecutive retry for the same payload will, on average, take longer and longer until reaching
+# MaxDuration and will also be desynchronized with retrial processes of other agents.
+# The following parameters enable you to configure MaxDuration, Base and GrowthBase, respectively.
+exp_backoff_max_duration_seconds=120
+exp_backoff_base_milliseconds=200
+exp_backoff_growth_base=2
+
+###################################################
+# Stat writer - aggregates stats about traces
+# and sends them to Datadog servers
+###################################################
+[trace.writer.stats]
+# Period with which the agent will send updated stat writer statistics to DD servers.
+update_info_period_seconds=60
+# For how many seconds should we keep stats payloads queued up in case we are unable to send them to DD servers.
+# A value <= 0 disables this limit
+# Default: 20 minutes
+queue_max_age_seconds=1200
+# Maximum amount of stats data we can have queued up in case we are unable to send them to DD servers.
+# A value <= 0 disables this limit
+# Default: 256MB
+queue_max_bytes=268435456
+# Maximum amount of stats payloads we can have queued up in case we are unable to send them to DD servers.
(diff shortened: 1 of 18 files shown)


AlexJF:
[dd-agent: Allow easy configuration of sending queues, max age, max queued payloads, etc...](https://trello.com/c/Gf9VT9SI/27-dd-agent-allow-easy-configuration-of-sending-queues-max-age-max-queued-payloads-etc)

agent/trace-agent.ini
```diff
@@ -71,3 +71,86 @@ oldest_span_cutoff_seconds=30
...
+# that are sent to Datadog servers
+###################################################
+[trace.writer.traces]
+# Information about different traces is batched together for performance reasons. This setting controls the maximum
+# amount of spans inside one of these batches. Bigger values lead to fewer but bigger payloads.
+max_spans_per_payload=1000
+# Period with which the agent forces flushes of payloads. In case the previous limit is not reached within this
+# period, a payload with less than max_spans_per_payload will be sent.
+flush_period_seconds=5
+# Period with which the agent will send updated trace writer statistics to DD servers.
```
LotharSee: update_info is the update of the variables exposed by expvar to the info command (`trace-agent -info`). It doesn't relate to the server.

AlexJF: Oh had missed that part. It's also when we flush the statsd counts/gauges though so it also relates to DD servers.

LotharSee: And I missed this part :P Cool, that's not important.

LotharSee: `As discussed personally, let's remove these various `update_info_period_seconds` from the exposed configuration.

agent/trace-agent.ini
```diff
@@ -71,3 +71,86 @@ oldest_span_cutoff_seconds=30
...
+# A value <= 0 disables this limit
+# Default: Unlimited
+queue_max_payloads=-1
+# When we are unable to send traces to DD servers, retries will be attempted with various delays. Each retry delay
+# is determined by a full-jitter exponential backoff that performs the following computation:
+# delay = Random(0, Min(MaxDuration, (GrowthBase ^ NumRetries) * Base))
+# The idea is that each consecutive retry for the same payload will, on average, take longer and longer until reaching
+# MaxDuration and will also be desynchronized with retrial processes of other agents.
+# The following parameters enable you to configure MaxDuration, Base and GrowthBase, respectively.
+exp_backoff_max_duration_seconds=120
```
LotharSee: These (all the exp_ options) are also very deep and no user will care about adjusting it (it should only be up to Datadog to set sane values). I'd suggest to remove them from the configuration.

agent/trace-agent.ini
```diff
@@ -71,3 +71,86 @@ oldest_span_cutoff_seconds=30
...
+exp_backoff_base_milliseconds=200
+exp_backoff_growth_base=2
+
+###################################################
+# Service writer - extracts service info from
+# traces and sends it to Datadog servers.
+###################################################
+[trace.writer.services]
+# Period with which the agent flushes service info to DD servers.
+flush_period_seconds=5
```
LotharSee: From what I saw, this would flush every 5s even if data didn't change ; which is way too much.
Let's sync with @p-lambert @talwai on how they are changing the reporting logic before converging on its configuration.

AlexJF: Right now, it will flush only if updated:
https://github.com/DataDog/datadog-trace-agent/blob/master/writer/service_writer.go#L144

Update is done by:
https://github.com/DataDog/datadog-trace-agent/blob/master/writer/service_writer.go#L137

But yeah this should definitely be kept in mind if we are to remove the serviceBuffer.

agent/trace-agent.ini
AlexJF: Done

agent/trace-agent.ini
AlexJF: Done

agent/trace-agent.ini
```diff
@@ -71,3 +71,53 @@ oldest_span_cutoff_seconds=30
...
+
+###################################################
+# Stat writer - aggregates stats about traces
+# and sends them to Datadog servers
+###################################################
+[trace.writer.stats]
+# For how many seconds should we keep stats payloads queued up in case we are unable to send them to DD servers.
+# A value <= 0 disables this limit
+# Default: 20 minutes
+queue_max_age_seconds=1200
```
LotharSee: TODO for tomorrow: let's review together a sane value here (one which would work with the backend).

agent/trace-agent.ini
```diff
@@ -71,3 +71,53 @@ oldest_span_cutoff_seconds=30
...
+# Maximum amount of trace data we can have queued up in case we are unable to send them to DD servers.
+# A value <= 0 disables this limit
+# Default: 256MB
+queue_max_bytes=268435456
+# Maximum amount of trace payloads we can have queued up in case we are unable to send them to DD servers.
+# A value <= 0 disables this limit
+# Default: Unlimited
+queue_max_payloads=-1
```
LotharSee: Yet another one I'd suggest to remove, as it is almost another flavor of `queue_max_age_seconds` (mas o menos).
The only one users will really care modifying is `queue_max_bytes`.

agent/trace-agent.ini
LotharSee: Okay let's go with 20 minutes for all endpoints for now!

writer/config/service_writer.go
```diff
@@ -0,0 +1,21 @@
...
+
+// ServiceWriterConfig contains the configuration to customize the behaviour of a ServiceWriter.
+type ServiceWriterConfig struct {
+	FlushPeriod      time.Duration
+	UpdateInfoPeriod time.Duration
+	SenderConfig     QueuablePayloadSenderConf
+}
+
+// DefaultServiceWriterConfig creates a new instance of a ServiceWriterConfig using default values.
+func DefaultServiceWriterConfig() ServiceWriterConfig {
```
LotharSee: Looks like DefaultServiceWriterConfig and DefaultQueuablePayloadSenderConf are only used for tests (as for the app we use the one from the config).

I'd suggest to make it a TestServiceWriterConfig, extracting it from the actual Agent code.

backoff/exponential.go
```diff
@@ -31,6 +29,12 @@ func DefaultExponentialDelayProvider() DelayProvider {
 
 // ExponentialDelayProvider creates a new instance of an ExponentialDelayProvider using the provided config.
 func ExponentialDelayProvider(conf ExponentialConfig) DelayProvider {
+	return ExponentialDelayProviderCustomRandom(conf, rand.New(rand.NewSource(time.Now().UnixNano())))
+}
+
+// ExponentialDelayProviderCustomRandom creates a new instance of ExponentialDelayProvider using the provided config
+// and random number generator.
+func ExponentialDelayProviderCustomRandom(conf ExponentialConfig, rand *rand.Rand) DelayProvider {
```
LotharSee: As `ExponentialDelayProvider` is the only one really used outside, should this one be made private?

writer/config/service_writer.go
AlexJF: It's actually used as the fallback if no such config key is present: 
* Service writer: 
  * https://github.com/DataDog/datadog-trace-agent/pull/349/files#diff-b7db7d3e3d014cfcce344902b2d07a78R192
  * https://github.com/DataDog/datadog-trace-agent/pull/349/files#diff-b7db7d3e3d014cfcce344902b2d07a78R400
* QueuablePayloadSender:
  * https://github.com/DataDog/datadog-trace-agent/pull/349/files#diff-b7db7d3e3d014cfcce344902b2d07a78R447
  * https://github.com/DataDog/datadog-trace-agent/pull/349/files#diff-b48b1e0826f1d12e8a4352c4f3fb700dR19

backoff/exponential.go
AlexJF: Agreed

agent/trace-agent.ini
AlexJF: Will do. Will also remove queue_max_age_seconds and leave only max_bytes. That's the main one of interest as you said.

writer/config/service_writer.go
LotharSee: Very true my bad ; then it is clean that way 👍

PR #352, opened 2018-01-29
Let multiple payloads to be enqueued on `ServiceWriter`

p-lambert:
Since we only send increments as opposed to full snapshots of service metadata, we must support enqueuing multiple payloads for `ServiceWriter`.

Should we also raise the default `MaxAge` configuration param for queued payloads? A single-service payload is less than 100 bytes, and the service number tends to be small as well (in the order of dozens), so I think it should be safe to increase it.

### DIFF
--- a/writer/service_writer.go
+++ b/writer/service_writer.go
@@ -36,11 +36,6 @@ func NewServiceWriter(conf *config.AgentConfig, InServices <-chan model.Services
 		serviceBuffer: model.ServicesMetadata{},
 		BaseWriter: *NewBaseWriter(conf, "/api/v0.2/services", func(endpoint Endpoint) PayloadSender {
 			senderConf := writerConf.SenderConfig
-			// Hardcode service sender configuration as the latest payload supplants older ones. So we don't really
-			// need more than 1 payload in the queue.
-			senderConf.MaxQueuedPayloads = 1
-			senderConf.MaxQueuedBytes = -1
-			senderConf.MaxAge = -1
 			return NewCustomQueuablePayloadSender(endpoint, senderConf)
 		}),
 	}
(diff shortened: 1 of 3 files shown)


writer/service_writer.go
```diff
@@ -36,11 +36,6 @@ func NewServiceWriter(conf *config.AgentConfig, InServices <-chan model.Services
 		serviceBuffer: model.ServicesMetadata{},
 		BaseWriter: *NewBaseWriter(conf, "/api/v0.2/services", func(endpoint Endpoint) PayloadSender {
 			senderConf := writerConf.SenderConfig
-			// Hardcode service sender configuration as the latest payload supplants older ones. So we don't really
-			// need more than 1 payload in the queue.
-			senderConf.MaxQueuedPayloads = 1
-			senderConf.MaxQueuedBytes = -1
-			senderConf.MaxAge = -1
```
p-lambert: I guess we should keep this, right?
```
senderConf.MaxQueuedBytes = -1
senderConf.MaxAge = -1
```

AlexJF:
Also, we might want to add an entry to the `trace-agent.ini` configuration file to configure the queue (like we have for the other writers)

writer/service_writer.go
AlexJF: I wouldn't for MaxQueuedBytes should ideally come from the trace-agent.ini (it defaults to 256MB which should be more than enough for services anyway but just in case lets give the option of tuning this). For MaxAge, we could do it but then I'd prefer if we simply had it as a default in case no MaxAge is configured via trace-agent.ini (it's not documented but it can be done).

p-lambert: Agreed!

PR #356, opened 2018-02-05
Fix transaction sampling by moving it to TransactionSampler

LotharSee:
Dedupe extracted transactions by removing them from them Sampler (we used to need it there as we were tagging transactions of sampled traces) to its own TransactionSampler.

At the same time, add the count of transactions to the info page and metrics reported.


talwai:
Much cleaner indeed! LGTM

PR #357, opened 2018-02-06
Allows Trace Agent to read from Datadog.yaml file when using Agent 6

nmuesch:
This PR allows the Trace Agent to properly read from datadog.yaml when using Agent 6. This specifically uses the already available `--ddconfig` command line argument when starting the Agent to specify which configuration file to use. 

A lot of this was heavily inspired by Conors code for the Process Agent.

Example Datadog Yaml piece:

```
additional_checksd: /etc/datadog-agent/checks.d
api_key: *******************************
apm_enabled: true
app_key: ""
apm_config:
  env: "demo"
  apm_non_local_traffic: true
  ignore_resource: "<Resource_regex>"
  service_writer:
    update_info_period_seconds: 10
    queue: 
      max_age_seconds: 10
      max_bytes: 1024
  stats_writer:
    update_info_period_seconds: 15
    queue:
      max_age_seconds: 15
```

I took the options available from datadog.conf here - https://github.com/DataDog/dd-agent/blob/master/datadog.conf.example#L305-L330

Further work needs to be done to allow Agent 6 to use this by default and will work on that next:
1) Have Agent 6 run the APM Agent with the ddconfig flag pointing to datadog.yaml
2) Have Agent 6 properly import the trace sections from datadog.conf -> datadog.yaml (Sister PR - https://github.com/DataDog/datadog-agent/pull/1207/files)

Additionally, I doubled the `env` field so a user can specify one both in the outer section of the yaml file and in the Trace config separately.


glide.lock
```diff
@@ -72,3 +72,43 @@ testImports:
   subpackages:
```
LotharSee: Glide updates have to go through `glide.yaml` instead of manually modifying the lock file.

config/config_yaml.go
```diff
@@ -0,0 +1,130 @@
...
+		ExtraSampleRate    float64 `yaml:"extra_sample_rate"`
+		MaxTracesPerSecond float64 `yaml:"max_traces_per_second"`
+		TraceIgnore        struct {
+			Ignore map[string][]string `yaml:"resource"`
+		} `yaml:"trace_ignore"`
+		TraceReceiver struct {
+			ReceiverPort    int `yaml:"receiver_port"`
+			ConnectionLimit int `yaml:"connection_limit"`
+		} `yaml:"trace_receiver"`
+	} `yaml:"trace_config"`
```
LotharSee: Is this how we do with other agents? (`process_config`, `log_config`, ...?). Let's try to be consistent.
Do you have a full Agent6 config (including config of these agents)?

nmuesch: Yupp, completely agree. I made this consistent with how the Process Agent works - https://github.com/DataDog/datadog-process-agent/blob/master/config/yaml_config.go#L23

glide.lock
nmuesch: Ahh good catch. Will get this one fixed!

config/config_yaml.go
nmuesch: A more full example here - https://github.com/DataDog/datadog-agent/blob/master/pkg/config/config_template.yaml#L333-L363 I will update this with examples of what the yaml should look like after this is merged.

config/config_yaml.go
```diff
@@ -0,0 +1,126 @@
...
+		ExtraSampleRate    float64             `yaml:"extra_sample_rate"`
+		MaxTracesPerSecond float64             `yaml:"max_traces_per_second"`
+		Ignore             map[string][]string `yaml:"ignore_resource"`
+		ReceiverPort       int                 `yaml:"receiver_port"`
+		ConnectionLimit    int                 `yaml:"connection_limit"`
+	} `yaml:"trace_config"`
+}
+
+// ReadLines reads contents from a file and splits them by new lines.
+func ReadLines(filename string) ([]string, error) {
```
AlexJF: Nit: This sounds like something that should go into a utils package (preferably) or be kept private so as not to pollute the config package namespace.

config/config_yaml.go
```diff
@@ -0,0 +1,126 @@
...
+		if err = yaml.Unmarshal([]byte(strings.Join(lines, "\n")), &yamlConf); err != nil {
+			return nil, fmt.Errorf("parse error: %s", err)
+		}
+		return &yamlConf, nil
+	}
+	return nil, nil
+}
+
+// PathExists returns a boolean indicating if the given path exists on the file system.
+func PathExists(filename string) bool {
```
AlexJF: Nit: This sounds like something that should go into a utils package (preferably) or be kept private so as not to pollute the config package namespace.

config/config_yaml.go
```diff
@@ -0,0 +1,126 @@
...
+
+// NewYamlIfExists returns a new YamlAgentConfig if the given configPath is exists.
+func NewYamlIfExists(configPath string) (*YamlAgentConfig, error) {
+	var yamlConf YamlAgentConfig
+	if PathExists(configPath) {
+		lines, err := ReadLines(configPath)
+		if err != nil {
+			return nil, fmt.Errorf("read error: %s", err)
+		}
+		if err = yaml.Unmarshal([]byte(strings.Join(lines, "\n")), &yamlConf); err != nil {
```
AlexJF: So we're reading, splitting by line, then joining the lines again without doing anything with them? Might as well just pass the file contents directly via `ioutil.ReadFile(configPath)` no?

config/config_yaml.go
```diff
@@ -0,0 +1,126 @@
...
+
+	agentConf.StatsdHost = yc.StatsdHost
+	agentConf.ReceiverHost = yc.ReceiverHost
+
+	return agentConf, nil
+}
+
+// SetupDDAgentConfig initializes the datadog-agent config with a YAML file.
+// This is required for configuration to be available for container listeners.
+func SetupDDAgentConfig(configPath string) error {
```
AlexJF: There are 2 `SetupDDAgentConfig` functions, one here and one in `agent/main.go` and no callers?

config/config_yaml.go
```diff
@@ -0,0 +1,126 @@
...
+
+	TraceAgent struct {
+		Env                string              `yaml:"env"`
+		ExtraSampleRate    float64             `yaml:"extra_sample_rate"`
+		MaxTracesPerSecond float64             `yaml:"max_traces_per_second"`
+		Ignore             map[string][]string `yaml:"ignore_resource"`
+		ReceiverPort       int                 `yaml:"receiver_port"`
+		ConnectionLimit    int                 `yaml:"connection_limit"`
+	} `yaml:"trace_config"`
+}
```
AlexJF: Should we use this opportunity to tackle this [FIXME](https://github.com/DataDog/datadog-trace-agent/blob/142cc36b60bdaf5b8bc481310b2b4318905b14bc/agent/main_nix.go#L14) and thus handle all the configuration options missing from `datadog.conf` like:
* Writer options: https://github.com/DataDog/datadog-trace-agent/blob/d422c3a812362dc4106aff26c0bd3640766f655b/config/agent.go#L54
* Transaction enabling per service: https://github.com/DataDog/datadog-trace-agent/blob/d422c3a812362dc4106aff26c0bd3640766f655b/config/agent.go#L80
* And all the others that are only set from this point onwards: https://github.com/DataDog/datadog-trace-agent/blob/d422c3a812362dc4106aff26c0bd3640766f655b/config/agent.go#L274

Or is the immediate plan to continue to rely on the `legacyConf` (`/etc/datadog/trace-agent.ini`) for most such options and only convert those that were already in `datadog.conf`?

nmuesch: I think this is a good idea and will include shortly.

agent/main.go
```diff
@@ -145,15 +131,18 @@ func runAgent(exit chan struct{}) {
...
 	}
 
-	conf, err := config.NewIfExists(opts.ddConfigFile)
-	if err != nil {
-		log.Errorf("%s: %v", opts.ddConfigFile, err)
-		log.Warnf("ignoring %s", opts.ddConfigFile)
+	if filepath.Ext(opts.ddConfigFile) == ".conf" {
+		conf, err = config.NewIfExists(opts.ddConfigFile)
+		if err != nil {
+			log.Errorf("%s: %v", opts.ddConfigFile, err)
```
AlexJF: Nit: This one should probably be critical as well and match the format of the .yaml version.

agent/main.go
```diff
@@ -145,15 +131,18 @@ func runAgent(exit chan struct{}) {
...
+		}
 	}
 
-	yamlConf, err := config.NewYamlIfExists(opts.ddConfigFile)
-	if err != nil {
-		log.Criticalf("Error reading datadog.yaml: %s", err)
+	if filepath.Ext(opts.ddConfigFile) == ".yaml" {
```
AlexJF: Nit: This should be an `else if`

config/config_yaml.go
AlexJF: I'm fine with a separate PR! :)

config/config_yaml.go
```diff
@@ -54,37 +56,32 @@ func ReadLines(filename string) ([]string, error) {
...
-		lines, err := ReadLines(configPath)
-		if err != nil {
-			return nil, fmt.Errorf("read error: %s", err)
-		}
-		if err = yaml.Unmarshal([]byte(strings.Join(lines, "\n")), &yamlConf); err != nil {
+	if utils.PathExists(configPath) {
+		// lines, err := ReadLines(configPath)
+		// if err != nil {
+		// 	return nil, fmt.Errorf("read error: %s", err)
+		// }
```
AlexJF: Don't forget to remove these :)

config/config_test.go
```diff
@@ -220,7 +220,7 @@ func TestAnalyzedRateByService(t *testing.T) {
 	}, "\n")))
 
 	conf := &File{instance: config, Path: "whatever"}
-	agentConfig, _ := NewAgentConfig(conf, nil)
+	agentConfig, _ := NewAgentConfig(conf, nil, nil)
```
AlexJF: We should add a test testing only yaml and yaml + legacy (maybe not this last one since we plan to remove legacy in another PR).

agent/main.go
```diff
@@ -127,16 +131,23 @@ func runAgent(exit chan struct{}) {
...
+	if filepath.Ext(opts.ddConfigFile) == ".conf" {
+		conf, err = config.NewIfExists(opts.ddConfigFile)
+		if err != nil {
+			log.Errorf("%s: %v", opts.ddConfigFile, err)
+		}
 	}
-	if conf != nil {
-		log.Infof("using configuration from %s", opts.ddConfigFile)
+
+	if filepath.Ext(opts.ddConfigFile) == ".yaml" {
```
AlexJF: Nit: This should be an `else if`.

agent/main.go
```diff
@@ -127,16 +131,23 @@ func runAgent(exit chan struct{}) {
...
 	}
 
-	conf, err := config.NewIfExists(opts.ddConfigFile)
-	if err != nil {
-		log.Errorf("%s: %v", opts.ddConfigFile, err)
-		log.Warnf("ignoring %s", opts.ddConfigFile)
+	if filepath.Ext(opts.ddConfigFile) == ".conf" {
+		conf, err = config.NewIfExists(opts.ddConfigFile)
+		if err != nil {
+			log.Errorf("%s: %v", opts.ddConfigFile, err)
```
AlexJF: Nit: Should log the same way as yaml.

config/config_yaml.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// NewYamlIfExists returns a new YamlAgentConfig if the given configPath is exists.
+func NewYamlIfExists(configPath string) (*YamlAgentConfig, error) {
+	var yamlConf YamlAgentConfig
+	if utils.PathExists(configPath) {
+		// lines, err := ReadLines(configPath)
+		// if err != nil {
+		// 	return nil, fmt.Errorf("read error: %s", err)
+		// }
```
AlexJF: Don't forget to remove this and the above function :)

config/config_yaml.go
```diff
@@ -34,6 +34,38 @@ type YamlAgentConfig struct {
...
 		ConnectionLimit    int                 `yaml:"connection_limit"`
 		NonLocalTraffic    string              `yaml:"trace_non_local_traffic"`
+		
+		//TODO Merge these into config
+		TraceWriter struct {
+			MaxSpansPerPayload int            `yaml:"max_spans_per_payload"`
+			FlushPeriod        int            `yaml:"flush_period_seconds"`
+			UpdateInfoPeriod   int            `yaml:"update_info_period_seconds"`
+			MaxAge             int            `yaml:"queue_max_age_seconds"`
+			MaxQueuedBytes     int64            `yaml:"queue_max_bytes"`
```
AlexJF: Nit: alignment (maybe golint even complains about this? that guy is pretty pedantic 😝)

config/config_yaml.go
```diff
@@ -34,6 +34,38 @@ type YamlAgentConfig struct {
...
+			MaxQueuedPayloads  int            `yaml:"queue_max_payloads"`
+			BackoffDuration    int            `yaml:"exp_backoff_max_duration_seconds"`
+			BackoffBase        int            `yaml:"exp_backoff_base_milliseconds"`
+			BackoffGrowth      int            `yaml:"exp_backoff_growth_base"`
+		} `yaml:"trace_writer"`
+		ServiceWriter struct {
+			FlushPeriod       int             `yaml:"flush_period_seconds"`
+			UpdateInfoPeriod  int             `yaml:"'update_info_period_seconds"`
```
AlexJF: Extra `'`

glide.lock
nmuesch: Fixed

config/config_yaml.go
```diff
@@ -102,5 +134,35 @@ func mergeYamlConfig(agentConf *AgentConfig, yc *YamlAgentConfig) (*AgentConfig,
...
+	agentConf.ServiceWriterConfig.SenderConfig.ExponentialBackoff.GrowthBase = yc.TraceAgent.ServiceWriter.BackoffGrowth
+
+	//Stats Writer
+    agentConf.StatsWriterConfig.UpdateInfoPeriod = utils.GetDuration(yc.TraceAgent.StatsWriter.UpdateInfoPeriod)
+	agentConf.StatsWriterConfig.SenderConfig.MaxAge = utils.GetDuration(yc.TraceAgent.StatsWriter.MaxAge)
+	agentConf.StatsWriterConfig.SenderConfig.MaxQueuedBytes = yc.TraceAgent.StatsWriter.MaxQueuedBytes
+	agentConf.StatsWriterConfig.SenderConfig.MaxQueuedPayloads = yc.TraceAgent.StatsWriter.MaxQueuedPayloads
+	agentConf.StatsWriterConfig.SenderConfig.ExponentialBackoff.MaxDuration = utils.GetDuration(yc.TraceAgent.StatsWriter.BackoffDuration)
+	agentConf.StatsWriterConfig.SenderConfig.ExponentialBackoff.Base = utils.GetDuration(yc.TraceAgent.StatsWriter.BackoffBase)
+	agentConf.StatsWriterConfig.SenderConfig.ExponentialBackoff.GrowthBase = yc.TraceAgent.StatsWriter.BackoffGrowth
```
AlexJF: ^ Wow that's a mouthful 😄 Maybe we can abstract away the filling of SenderConfig stuff like how it's done around here: https://github.com/DataDog/datadog-trace-agent/blob/858083c4a9b5de23a3824d8f2fdd6a60576e837f/config/agent.go#L446 ?

Would make any future changes to SenderConfig more maintainable.

nmuesch: I agree that would be a more maintainable option in the future. Though I'm not seeing a great way to accomplish that in go. Since these are nested structs, I can't access them via a "section" like is done in that function where you are reading sections from a file. Any thoughts there?

nmuesch:
In the sake of time and whats most useful, I think this may be good. I implemented the functions to set the agentConfig more dynamically, but that introduces a little more hierarchy into the yaml structure. I think this is OK as its less commonly used fields, but I've documented them above in the PR description. 

Let me know what you think and thanks for the feedback 🙇

config/config_yaml.go
```diff
@@ -0,0 +1,207 @@
...
+	agentConf.Enabled = yc.TraceAgent.Enabled
+	agentConf.DefaultEnv = yc.DefaultEnv
+
+	agentConf.ReceiverPort = yc.TraceAgent.ReceiverPort
+	agentConf.ExtraSampleRate = yc.TraceAgent.ExtraSampleRate
+	agentConf.MaxTPS = yc.TraceAgent.MaxTracesPerSecond
+
+	agentConf.Ignore["resource"] = strings.Split(yc.TraceAgent.Ignore, ",")
+
+	agentConf.ConnectionLimit = yc.TraceAgent.ConnectionLimit
```
AlexJF: What happens if the user doesn't specify a connection limit in the YAML? Seems like default values are overridden? (same applies to all other settings where we do a direct assignment without checking if the yc.TraceAgent part actually contains data).

config/config_yaml.go
```diff
@@ -0,0 +1,207 @@
...
+
+	c.SenderConfig = readQueueablePayloadSenderConfigYaml(yc.QueueablePayloadSender)
+
+	return c
+}
+
+func readQueueablePayloadSenderConfigYaml(yc queueablePayloadSender) writerconfig.QueuablePayloadSenderConf {
+	c := writerconfig.DefaultQueuablePayloadSenderConf()
+
+	if yc.MaxAge > 0 {
```
AlexJF: `MaxAge`, `MaxQueuedBytes` and `MaxQueuedPayloads` actually treat <= 0 as infinite so this check should probably be `!= 0` so the user can specify `-1` in the YAML to say they don't want to delete old payloads (I assume `0` is the value that comes in `yc` if the user didn't specify anything right?).

utils/config_file.go
```diff
@@ -0,0 +1,19 @@
...
+// PathExists returns a boolean indicating if the given path exists on the file system.
+func PathExists(filename string) bool {
+	if _, err := os.Stat(filename); err == nil {
+		return true
+	}
+	return false
+}
+
+// GetDuration returns the duration of the provided value
+func GetDuration(value int) time.Duration {
```
AlexJF: Since this assumes `value` is seconds this should either be documented or made explicit by calling the function `DurationFromSeconds` or something like that.

config/config_yaml.go
```diff
@@ -0,0 +1,207 @@
...
+	ExtraSampleRate    float64 `yaml:"extra_sample_rate"`
+	MaxTracesPerSecond float64 `yaml:"max_traces_per_second"`
+	Ignore             string  `yaml:"ignore_resource"`
+	ReceiverPort       int     `yaml:"receiver_port"`
+	ConnectionLimit    int     `yaml:"connection_limit"`
+	NonLocalTraffic    string  `yaml:"trace_non_local_traffic"`
+
+	TraceWriter   traceWriter   `yaml:"trace_writer"`
+	ServiceWriter serviceWriter `yaml:"service_writer"`
+	StatsWriter   statsWriter   `yaml:"stats_writer"`
```
AlexJF: Cheers for adding the writer configs, had assumed they would come in a separate PR though :)

There's one last config that should definitely be exposed in the YAML as well and that is tied to our transaction analytics: https://github.com/DataDog/datadog-trace-agent/blob/858083c4a9b5de23a3824d8f2fdd6a60576e837f/config/agent.go#L80

Example config:

```
[trace.analyzed_rate_by_service]
web = 0.9
db = 1
```

I'm totally fine with this appearing in a different PR so we can close this one quickly but I think it's pretty important that this ships with this release as well (along with all the others missing but since we're tight on time, I can accept that we keep the legacy `/etc/datadog/trace-agent.ini` config for the more obscure settings).

config/config_yaml.go
AlexJF: Oh I got what you were saying. It's because you wanted the config to be flatter so the payload sender stuff was directly under trace_writer and so on. 

I think I actually prefer how it was before. Either that or change the config so it looks like

```
apm_config:
  env: "demo"
  trace_non_local_traffic: true
  ignore_resource: "<Resource_regex>"
  trace_writer:
    update_info_period_seconds: 10
    queue: 
      max_bytes: 1024
```

There's still a weird nesting going on there for the queue where most of the times `max_bytes` should be the only interesting option. That's what pushing me back to what you had before:

```
apm_config:
  env: "demo"
  trace_non_local_traffic: true
  ignore_resource: "<Resource_regex>"
  trace_writer:
    update_info_period_seconds: 10
    queue_max_bytes: 1024
```

What do you think?

config/config_yaml.go
```diff
@@ -0,0 +1,229 @@
...
+
+//YamlAgentConfig is the Primary Object we retrieve from Datadog.yaml
+type YamlAgentConfig struct {
+	APIKey   string `yaml:"api_key"`
+	HostName string `yaml:"hostname"`
+
+	StatsdHost   string `yaml:"bind_host"`
+	ReceiverHost string ""
+
+	StatsdPort int    `yaml:"StatsdPort"`
```
LotharSee: Looking at https://github.com/DataDog/datadog-agent/blob/b5d2187bfe3f3a8457bebdde9e17a2d7af9ef8c6/pkg/config/config_template.yaml it seems to be the `dogstatsd_port` key isntead.

config/config_yaml.go
```diff
@@ -0,0 +1,229 @@
...
+
+	AnalyzedRateByService map[string]float64 `yaml:"analyzed_rate_by_service"`
+}
+
+//YamlAgentConfig is the Primary Object we retrieve from Datadog.yaml
+type YamlAgentConfig struct {
+	APIKey   string `yaml:"api_key"`
+	HostName string `yaml:"hostname"`
+
+	StatsdHost   string `yaml:"bind_host"`
```
LotharSee: This is now ill-defined. We need to distinguish the dest port for our dogstatsd metrics from which interface the trace-agent will listen to.

config/config_yaml.go
```diff
@@ -0,0 +1,229 @@
...
+	ReceiverHost string ""
+
+	StatsdPort int    `yaml:"StatsdPort"`
+	LogLevel   string `yaml:"log_level"`
+
+	DefaultEnv string `yaml:"env"`
```
LotharSee: This is at the `TraceAgent` level.

config/config_test.go
nmuesch: Looks like Jaime added one :)

LotharSee:
Got extended and replaced by https://github.com/DataDog/datadog-trace-agent/pull/359

PR #362, opened 2018-02-15
Make Backend a MemoryBackend of a Backend interface

LotharSee:
Make `Backend` an interface.
It should make the code more explicit about what it exposes, and will allow a better reusability/flexibility.

### DIFF
--- a/sampler/backend.go
+++ b/sampler/backend.go
@@ -1,159 +1,28 @@
 package sampler
 
-import (
-	"sync"
-	"time"
-)
+// Backend stores and count traces and signatures ingested by a sampler
+type Backend interface {
+	// Run runs the blocking execution of the backend main loop
+	Run()
+	// Stop stops the backend main loop
+	Stop()
 
-// Backend storing any state required to run the sampling algorithms.
-//
-// Current implementation is only based on counters with polynomial decay.
-// Its bias with steady counts is 1 * decayFactor.
-// The stored scores represent approximation of the real count values (with a countScaleFactor factor).
-type Backend struct {
-	// Score per signature
-	scores map[Signature]float64
-	// Score of all traces (equals the sum of all signature scores)
-	totalScore float64
-	// Score of sampled traces
-	sampledScore float64
-	mu           sync.RWMutex
+	// CountSample counts that 1 trace is going through the sampler
+	CountSample()
+	// CountSignature counts that 1 signature is going through the sampler
+	CountSignature(signature Signature)
 
-	// Every decayPeriod, decay the score
-	// Lower value is more reactive, but forgets quicker
-	decayPeriod time.Duration
-	// At every decay tick, how much we reduce/divide the score
-	// Lower value is more reactive, but forgets quicker
-	decayFactor float64
-	// Factor to apply to move from the score to the representing number of traces per second.
-	// By definition of the decay formula: countScaleFactor = (decayFactor / (decayFactor - 1)) * decayPeriod
-	// It also represents by how much a spike is smoothed: if we instantly receive N times the same signature,
-	// its immediate count will be increased by N / countScaleFactor.
-	countScaleFactor float64
+	// GetTotalScore returns the TPS of all traces ingested
+	GetTotalScore() float64
+	// GetSampledScore returns the TPS of all traces sampled
+	GetSampledScore() float64
+	// GetUpperSampledScore is similar to GetSampledScore, but with the upper approximation
+	GetUpperSampledScore() float64
+	// GetSignatureScore returns the TPS of traces ingested of a given signature
+	GetSignatureScore(signature Signature) float64
+	// GetAllSignatureScores returns the TPS of traces ingested for all signatures
+	GetAllSignatureScores() map[Signature]float64
 
-	exit chan struct{}
-}
-
-// NewBackend returns an initialized Backend
-func NewBackend(decayPeriod time.Duration) *Backend {
-	// With this factor, any past trace counts for less than 50% after 6*decayPeriod and >1% after 39*decayPeriod
-	// We can keep it hardcoded, but having `decayPeriod` configurable should be enough?
-	decayFactor := 1.125 // 9/8
-
-	return &Backend{
-		scores:           make(map[Signature]float64),
-		sampledScore:     0,
-		decayPeriod:      decayPeriod,
-		decayFactor:      decayFactor,
-		countScaleFactor: (decayFactor / (decayFactor - 1)) * decayPeriod.Seconds(),
-		exit:             make(chan struct{}),
-	}
-}
-
-// Run runs and block on the Sampler main loop
-func (b *Backend) Run() {
-	t := time.NewTicker(b.decayPeriod)
-	defer t.Stop()
-
-	for {
-		select {
-		case <-t.C:
-			b.DecayScore()
-		case <-b.exit:
-			return
-		}
-	}
-}
-
-// Stop stops the main Run loop
-func (b *Backend) Stop() {
-	close(b.exit)
-}
-
-// CountSignature counts an incoming signature
-func (b *Backend) CountSignature(signature Signature) {
-	b.mu.Lock()
-	b.scores[signature]++
-	b.totalScore++
-	b.mu.Unlock()
-}
-
-// CountSample counts a trace sampled by the sampler
-func (b *Backend) CountSample() {
-	b.mu.Lock()
-	b.sampledScore++
-	b.mu.Unlock()
-}
-
-// GetSignatureScore returns the score of a signature.
-// It is normalized to represent a number of signatures per second.
-func (b *Backend) GetSignatureScore(signature Signature) float64 {
-	b.mu.RLock()
-	score := b.scores[signature] / b.countScaleFactor
-	b.mu.RUnlock()
-
-	return score
-}
-
-// GetAllSignatureScores returns the scores for all signatures.
-// It is normalized to represent a number of signatures per second.
-func (b *Backend) GetAllSignatureScores() map[Signature]float64 {
-	b.mu.RLock()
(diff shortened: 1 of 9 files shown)


sampler/backend.go
```diff
@@ -1,159 +1,28 @@
...
+	// Stop stops the backend main loop
+	Stop()
+
+	// CountSample counts that 1 trace is going through the sampler
+	CountSample()
+	// CountSignature counts that 1 signature is going through the sampler
+	CountSignature(signature Signature)
+
+	// GetTotalScore returns the TPS of all traces ingested
+	GetTotalScore() float64
```
gbbr: Would be awesome if we could rid ourselves of the `Get*` prefix for these methods. In "[_Effective Go_](https://golang.org/doc/effective_go.html#Getters)" you can see in the corresponding section that it is advised to not use this prefix in Go. Unless there is a good reason for it, would be great to remove it.

sampler/backend.go
```diff
@@ -1,159 +1,28 @@
...
+	// GetSampledScore returns the TPS of all traces sampled
+	GetSampledScore() float64
+	// GetUpperSampledScore is similar to GetSampledScore, but with the upper approximation
+	GetUpperSampledScore() float64
+	// GetSignatureScore returns the TPS of traces ingested of a given signature
+	GetSignatureScore(signature Signature) float64
+	// GetAllSignatureScores returns the TPS of traces ingested for all signatures
+	GetAllSignatureScores() map[Signature]float64
+
+	// get the number of different signatures seen
```
gbbr: Please start the documentation name with the name of the method, e.g.:

> GetCardinality returns the number of distinct signatures seen.

sampler/backend.go
```diff
@@ -1,159 +1,28 @@
...
+	// CountSignature counts that 1 signature is going through the sampler
+	CountSignature(signature Signature)
+
+	// GetTotalScore returns the TPS of all traces ingested
+	GetTotalScore() float64
+	// GetSampledScore returns the TPS of all traces sampled
+	GetSampledScore() float64
+	// GetUpperSampledScore is similar to GetSampledScore, but with the upper approximation
+	GetUpperSampledScore() float64
+	// GetSignatureScore returns the TPS of traces ingested of a given signature
```
gbbr: Unless the TPS concept is explained elsewhere in another doc, I think it would be great to say it "traces per second" instead of TPS here.

sampler/backend.go
```diff
@@ -1,159 +1,28 @@
...
+// Backend stores and count traces and signatures ingested by a sampler
+type Backend interface {
+	// Run runs the blocking execution of the backend main loop
+	Run()
+	// Stop stops the backend main loop
+	Stop()
+
+	// CountSample counts that 1 trace is going through the sampler
+	CountSample()
+	// CountSignature counts that 1 signature is going through the sampler
```
gbbr: In some places throughout the definition of this interface you've left a blank line between the method and the documentation of the next method. Please try and keep consistency. As a personal preference I prefer to have the blank line, but you can chose whichever version you like as long as it's consistent (ideally throughout the package but at the minimum on the interface level).

sampler/backend.go
```diff
@@ -1,159 +1,28 @@
...
-			b.scores[sig] /= b.decayFactor
-		} else {
-			// When the score is too small, we can optimize by simply dropping the entry
-			delete(b.scores, sig)
-		}
-	}
-	b.totalScore /= b.decayFactor
-	b.sampledScore /= b.decayFactor
-	b.mu.Unlock()
+// Backend stores and count traces and signatures ingested by a sampler
```
gbbr: s/count/counts

Would be awesome if we could end each sentence with a period.

sampler/memory_backend.go
```diff
@@ -0,0 +1,155 @@
...
+	"time"
+)
+
+// MemoryBackend storing any state required to run the sampling algorithms.
+//
+// Current implementation is only based on counters with polynomial decay.
+// Its bias with steady counts is 1 * decayFactor.
+// The stored scores represent approximation of the real count values (with a countScaleFactor factor).
+type MemoryBackend struct {
+	// Score per signature
```
gbbr: Just as you did for the struct doc, it would be nice if the field comments would start with the identifier, e.g.:
```go
// scores maps signatures to scores
```

sampler/memory_backend.go
```diff
@@ -0,0 +1,155 @@
...
+	scores map[Signature]float64
+	// Score of all traces (equals the sum of all signature scores)
```
gbbr: e.g.
```go
// totalScore holds the score sum of all traces (equals the sum of all signature scores)
```

sampler/memory_backend.go
```diff
@@ -0,0 +1,155 @@
...
+	countScaleFactor float64
+
+	exit chan struct{}
+}
+
+// NewMemoryBackend returns an initialized Backend
+func NewMemoryBackend(decayPeriod time.Duration, decayFactor float64) *MemoryBackend {
+	return &MemoryBackend{
+		scores:           make(map[Signature]float64),
+		sampledScore:     0,
```
gbbr: Numeric fields start with a default value of `0`, so you can delete this line.

sampler/memory_backend.go
```diff
@@ -0,0 +1,155 @@
...
+// Stop stops the main Run loop
+func (b *MemoryBackend) Stop() {
+	close(b.exit)
+}
+
+// CountSignature counts an incoming signature
+func (b *MemoryBackend) CountSignature(signature Signature) {
+	b.mu.Lock()
+	b.scores[signature]++
+	b.totalScore++
```
gbbr: You could use `sync/atomic` both here and in the method below.

sampler/memory_backend.go
```diff
@@ -0,0 +1,155 @@
...
+// CountSample counts a trace sampled by the sampler
+func (b *MemoryBackend) CountSample() {
+	b.mu.Lock()
+	b.sampledScore++
+	b.mu.Unlock()
+}
+
+// GetSignatureScore returns the score of a signature.
+// It is normalized to represent a number of signatures per second.
+func (b *MemoryBackend) GetSignatureScore(signature Signature) float64 {
```
gbbr: This can be simplified without the local variable, but I don't feel strongly about it:
```go
b.mu.RLock()
defer b.mu.RUnlock()

return b.scores[signature] / b.countScaleFactor
```

sampler/backend.go
```diff
@@ -1,159 +1,28 @@
...
+	// GetTotalScore returns the TPS of all traces ingested
+	GetTotalScore() float64
+	// GetSampledScore returns the TPS of all traces sampled
+	GetSampledScore() float64
+	// GetUpperSampledScore is similar to GetSampledScore, but with the upper approximation
+	GetUpperSampledScore() float64
+	// GetSignatureScore returns the TPS of traces ingested of a given signature
+	GetSignatureScore(signature Signature) float64
+	// GetAllSignatureScores returns the TPS of traces ingested for all signatures
+	GetAllSignatureScores() map[Signature]float64
```
gbbr: It think it would be no problem nor would it make it more confusing to remove "All" from this name.

sampler/memory_backend.go
```diff
@@ -0,0 +1,155 @@
...
+
+// GetUpperSampledScore returns a certain upper bound of the global count of all sampled traces.
+func (b *MemoryBackend) GetUpperSampledScore() float64 {
+	// Overestimate the real score with the high limit of the backend bias.
+	return b.GetSampledScore() * b.decayFactor
+}
+
+// GetCardinality returns the number of different signatures seen recently.
+func (b *MemoryBackend) GetCardinality() int64 {
+	b.mu.Lock()
```
gbbr: I think this was meant to be a "read lock".

sampler/memory_backend.go
gbbr: You may do the same in all the below methods to get rid of the extra local variable.

sampler/memory_backend.go
```diff
@@ -0,0 +1,155 @@
...
+
+	return cardinality
+}
+
+// decayScore applies the decay to the rolling counters
+func (b *MemoryBackend) decayScore() {
+	b.mu.Lock()
+	for sig := range b.scores {
+		score := b.scores[sig]
+		if score > b.decayFactor*minSignatureScoreOffset {
```
gbbr: Since you only use `score` here, on this condition, it might be easier as:
```go
if score := b.scores[sig]; score > b.decayFactor*minSignatureScoreOffset {
```
...or even not having it at all

sampler/backend.go
LotharSee: Absolutely and that's my plan ; I don't wanted to mix goals so I wanted to keep it for a future PR.

sampler/backend.go
LotharSee: Right, they were grouped by "nature", but that's not explicit and a bit arbitrary.

sampler/backend.go
gbbr: If you want, I am happy to make these changes for you within this PR.

sampler/memory_backend.go
LotharSee: Right, the use of `atomic` could be considered for most of the unique counters of the MemoryBackend. I don't want to mix purposes in this PR but that's certainly something which could be addressed (with benchmarks?) in future work.

sampler/memory_backend.go
LotharSee: Good catch!

sampler/memory_backend.go
LotharSee: I have no strong opinion either.
At some places we aren't using `defer`, and from what I recall it isn't to micro-optimize but mostly because they are trivial situations that don't require it.
I'll be fine if we change it in a way or another in the future.

PR #363, opened 2018-02-16
Refactor sampling logic to remove dupe sampled traces.

AlexJF:
Before, when using multiple samplers, the same trace could end up being
sent twice.

With this new logic. All involved samplers will be asked if the trace should be kept or not but the samplers themselves won't add it to the output. The agent will look at the decision of all samplers and if at least one said to keep the trace then the trace is added once to the output channel.

Also, signature sampler should be asked for ALL traces so that it has a non-biased vision of the distributions.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -50,6 +50,8 @@ type Agent struct {
 	ServiceExtractor   *TraceServiceExtractor
 	ServiceMapper      *ServiceMapper
 
+	sampledTraceChan chan *model.Trace
+
 	// config
 	conf    *config.AgentConfig
 	dynConf *config.DynamicConfig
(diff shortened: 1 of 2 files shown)


agent/agent.go
```diff
@@ -50,6 +50,8 @@ type Agent struct {
 	ServiceExtractor   *TraceServiceExtractor
 	ServiceMapper      *ServiceMapper
 
+	sampledTraceChan chan *model.Trace
```
gbbr: I am not sure if sending a pointer down a channel counts as thread safe

AlexJF: At this point, pt is treated as read-only so that gives you thread-safety. Yeah it's an implicit contract but it's one we usually implicitly assume for the sake of performance. It's true that at this time we could be copying it by value for no extra cost since it's a slice anyway. But in the future this could change to a more complicated structure and sending it by pointer would save unnecessary copies.

PR #365, opened 2018-02-19
agent: move to cmd/trace-agent

gbbr:
The general approach in Go projects is to have the binary (main package)
in the `cmd` folder under the name that the binary will result in. This
PR makes that change and updates the build files.

`agent` gets renamed to `cmd/trace-agent`

We do the same in our very own `datadog-agent`


gbbr:
I have confirmed with croissant that this is true. `rake build` is used. Link to code [here](https://github.com/DataDog/datadog-agent/blob/master/omnibus/config/software/datadog-trace-agent.rb#L52-L53).

PR #366, opened 2018-02-19
build: upgrade to CircleCI 2.0

gbbr:
Reduces CI time from ~5 min to ~1 min.


gbbr:
@LotharSee fixed version to 1.8 as per suggestion. I've also enabled the tests to run in parallel, gaining another minute build time.

Will take your bonus suggestion into consideration as well, to update our build chain, that will also bring a lot of benefits.

PR #368, opened 2018-02-20
utils: remove package

gbbr:
The `utils` package held a small file with two helper methods that were only used in `config`. It is also not an appropriate (descriptive) name for a package.


config/merge_yaml.go
```diff
@@ -274,3 +274,8 @@ func readExponentialBackoffConfigYaml(yc queueablePayloadSender) backoff.Exponen
 
 	return c
 }
+
+// getDuration returns the duration of the provided value in seconds
+func getDuration(seconds int) time.Duration {
```
vlad-mh: Do we really need this?

gbbr: Not sure. It's this versus typing `time.Duration(n) * time.Second` every time, above.

PR #370, opened 2018-02-20
Integration+Bench tests around trace handling pipeline.

AlexJF:
As part of ongoing benchmark work, added 2 integration+benchmark tests exercising the whole trace handling pipeline. Each processes the same number of traces, sending it to a fake trace-api goroutine and tracks the time required for all those traces to be acknowledged as well as the total number of bytes sent. Difference between them is that one has transactions disabled whereas the other writes 1 transaction per trace.

Didn't write them as normal go benchmarks because that would require a non-negligible refactor of the agent to allow starting and stopping several agents per process as the test runner attempts to determine the right N.

### DIFF
--- a/cmd/trace-agent/agent_test.go
+++ b/cmd/trace-agent/agent_test.go
@@ -2,9 +2,13 @@ package main
 
 import (
 	"fmt"
+	"math/rand"
+	"net"
 	"net/http"
 	"runtime"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


AlexJF:
A series of local runs on my PC with 5000 traces, each with 20 spans report:

### No transactions
* Time: avg 3.94s (0.20s std deviation)
* Size: avg 14.65MB (~0.004MB std deviation)

### Transactions
* Time: avg. 4.13s (~0.07s std deviation)
* Size: avg. 15.66MB (~0.0053MB std deviation)

gbbr:
Why aren't we using [actual benchmarks](https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go)? That would allow us to compare stats moving from change to another in the future, using tools like [benchstat](https://godoc.org/golang.org/x/perf/cmd/benchstat).

cmd/trace-agent/agent_test.go
```diff
@@ -161,3 +165,148 @@ func formatTrace(t model.Trace) model.Trace {
...
+	srv.Close()
+}
+
+// findFreePort returns a free tcp port or panics
+func findFreePort() int {
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		panic(err)
+	}
+	defer listener.Close()
```
gbbr: Just to be on the safe side, might be better to use this instead of the type cast:

```go
	_, port, err := net.SplitHostPort(listener.Addr().String())
	if err != nil {
		return "", err
	}
	return port, err
```

gbbr: Or at the very least to check if the cast is ok

AlexJF:
I did comment on the reason in the PR. Anyway since this was not immediately merged, I'm doing the aforementioned refactor and will rewrite them as benchmarks and update this PR. As for the type cast part, this is a test helper so I'd prefer to have it panic directly instead of adding extra complexity by having to check return types and so on. If we plan on using this outside of tests then yes, I agree.

gbbr:
Sorry @AlexJF I've missed that paragraph. I don't mind having them like this either, but then perhaps they should be a separate program that is not part of the test-suite, if benchmarks are causing trouble.

One idea is to create a new folder called `testutil/benchmarks` and have this stuff in there. We could also move `fixtures` into `testutil/fixtures`.

Just a suggestion. You can do as you feel would be best.

cmd/trace-agent/agent_test.go
```diff
@@ -161,3 +165,148 @@ func formatTrace(t model.Trace) model.Trace {
...
+	return listener.Addr().(*net.TCPAddr).Port
```
gbbr: What I mean here is simply:
```go
if addr, ok := listener.Addr().(*net.TCPAddr); ok {
    return addr.Port
}
return some_other_default_port_as_fallback
```

cmd/trace-agent/agent_test.go
```diff
@@ -161,3 +165,148 @@ func formatTrace(t model.Trace) model.Trace {
...
+	rand.Seed(1)
+
+	// Create test traces
+	traces := make([]model.Trace, 0, numTraces)
+	for i := 0; i < numTraces; i++ {
+		trace := fixtures.RandomFixedSizeTrace(numSpansPerTrace)
+
+		// Panic if the number generator failed to preserve number of spans
+		if len(trace) != numSpansPerTrace {
+			panic(trace)
```
gbbr: Isn't it better to just fail?

cmd/trace-agent/agent_test.go
```diff
@@ -161,3 +165,148 @@ func formatTrace(t model.Trace) model.Trace {
...
+	// Start the agent with a waitgroup signaling when it stops
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		agent.Run()
+		wg.Done()
+	}()
+
+	// Create a mock http server tracking number of trace requests received and total trace request bytes.
+	totalBytesReceived := int64(0)
```
gbbr:
```go
var totalBytesReceived int64
```
Has the exact same output and is more idiomatic (e.g. no casting)

AlexJF:
Don't worry about it, the refactor I mentioned had already started before your comment, I just forgot to put the notice on this PR :) 

It's an important step in adding real benchmarks that can use all the goodies you linked to so it's definitely something that needs to be addressed.

cmd/trace-agent/agent_test.go
```diff
@@ -161,3 +165,148 @@ func formatTrace(t model.Trace) model.Trace {
...
+	// Keep track of received trace payloads in a waitgroup so we can wait for this at the end
+	reqWg := sync.WaitGroup{}
```
gbbr: Please change this to
```go
var reqWg sync.WaitGroup
```
That's the idiomatic way to define a waitgroup.

cmd/trace-agent/agent_test.go
AlexJF: Hmm yeah I guess I can add that for peace of mind 😄

gbbr:
Let's close this for now, it seems to have stalled. We can re-open it later.

PR #372, opened 2018-02-21
 {cmd,statsd,writer}: fix race conditions in tests

gbbr:
Running the tests with `go test -race /....` reveals several race conditions. This PR fixes them and enables the flag. It slows CI down a bit but adds more thread-safety.

### DIFF
--- a/cmd/trace-agent/service_mapper.go
+++ b/cmd/trace-agent/service_mapper.go
@@ -12,10 +12,12 @@ import (
 // ServiceMapper provides a cache layer over model.ServicesMetadata pipeline
 // Used in conjunction with ServiceWriter: in-> ServiceMapper out-> ServiceWriter
 type ServiceMapper struct {
-	in    <-chan model.ServicesMetadata
-	out   chan<- model.ServicesMetadata
-	exit  chan bool
-	done  sync.WaitGroup
+	in   <-chan model.ServicesMetadata
+	out  chan<- model.ServicesMetadata
+	exit chan bool
+	done sync.WaitGroup
+
+	mu    sync.RWMutex // guards cache
 	cache model.ServicesMetadata
 }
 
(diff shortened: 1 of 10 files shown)


cmd/trace-agent/service_mapper.go
```diff
@@ -85,11 +89,21 @@ func (s *ServiceMapper) update(metadata model.ServicesMetadata) {
...
+func (s *ServiceMapper) cacheSize() int {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return len(s.cache)
+}
+
 func (s *ServiceMapper) shouldAdd(service string, metadata model.ServicesMetadata) bool {
+	s.mu.RLock()
 	cacheEntry, ok := s.cache[service]
+	s.mu.RUnlock()
```
AlexJF: I don't really like adding all these locks just for the tests. Locks may have a significant impact in performance and in production this code would be perfectly thread-safe since all changes/accesses to cache are done by the same coroutine.

gbbr: I agree. But I think we should definitely re-enable the `-race` flag for tests. We could, of course see how we can improve the tests so that these locks are not necessary in subsequent PR imo.

AlexJF: Sure but I'd just remove the intermediate length checks from the test and just do a final one after the Stop (that should not race I think?) instead of adding the locks.

cmd/trace-agent/service_mapper.go
```diff
@@ -56,7 +58,9 @@ func (s *ServiceMapper) Run() {
 		case metadata := <-s.in:
 			s.update(metadata)
 		case <-telemetryTicker.C:
+			s.mu.RLock()
 			log.Infof("total number of tracked services: %d", len(s.cache))
```
LotharSee: Want to use `cacheSize` here instead?

gbbr: I was going to take @AlexJF suggestion from above and remove the locks, and the checks. This structure is already thread-safe (considering nobody accesses private variables within the same package).

LotharSee: Right very good point. +1 with adjusting test to run with the same thread-safety as the actual code.
Also, maybe add comments in the code when we do the assumption that a given struct will be executed in a thread-safe environment? Might be too much though.

gbbr:
Removed locks and redundant checks. A separate unit test can be made separately for the cache if needed. PTAL

gbbr:
@LotharSee happy?

PR #378, opened 2018-02-22
build: s/glide/dep

gbbr:
`dep init` automatically converted the glide files. Double-checked commits for each dependency.


glide.yaml
```diff
@@ -1,45 +0,0 @@
...
-  version: v2.17.01
-  subpackages:
-  - cpu
-- package: golang.org/x/sys
-  subpackages:
-  - windows/svc
-  - windows/svc/debug
-  - windows/svc/eventlog
-  - windows/svc/mgr
-- package: github.com/golang/protobuf
```
AlexJF: Shouldn't this package along with datadog-agent and golang.org/x/sys appear in Gopkg.toml in addition to the lock file?

gbbr: Based on my understanding of what the [`Gopkg.toml`](https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#constraint) file represents, I think that it only holds a set of "hand-written" rules, in this case constraints to certain tags that `dep init` has found in the glide configuration. I think the repos you've mentioned didn't have these constraints, they've only had commits set.

I would assume that if we ever want to update (using something like `dep ensure -update`) the `toml` file will block us from going past the tag in the constraint.

More info [here](https://github.com/golang/dep/blob/master/docs/Gopkg.toml.md#constraint)

AlexJF: Ok the actual explanation for this I guess comes from https://github.com/golang/dep/blob/master/docs/glossary.md#direct-dependency (emphasis mine):

> A project's direct dependencies are those that it imports from one or more of its packages, **or** includes in its required list in Gopkg.toml.

I guess dep automatically adds implicit `required` entries in Gopkg.toml for detected external imports.

Seems a bit strange to allow hidden dependencies like this though, at least coming from almost any other language.

gbbr: Thanks for reviewing Alex. Not sure what to say about your strange finding though :)

PR #379, opened 2018-02-22
vendor: commit dependencies into repository

gbbr:
### Pros

* Faster builds (CI builds 22 seconds)
* Less dependencies (dep is no longer a dependency, unless needed in development, for updates)
* Easier deployments (we don't need to download, install and run dep on each OS, every time)
* Hard copies of each library (no longer relying on network, e.g. site is down making repo unavailable)

### Cons

* Larger repository size (+41MB)
* Requires new contribution rules (each vendor update should be isolated into it's own PR or commit to avoid messy diffs)

⚠️ Note that this PR does not remove `dep`, it just commits the `vendor` folder to simplify builds, deployments and setting up development.


gbbr:
Just rebased. All good. Thanks for looking.

PR #383, opened 2018-02-26
[hostname] collect directly from agent6 if possible

truthbk:
If we're on agent6 let's try to grab the hostname using the agent's facilities.

Note: I'd like to improve the tests, but mocking Exec is a bit of a pain, so I'll do that in a separate PR.

### DIFF
--- a/config/agent.go
+++ b/config/agent.go
@@ -72,6 +72,9 @@ type AgentConfig struct {
 
 	// transaction analytics
 	AnalyzedRateByService map[string]float64
+
+	// infrastructure agent binary
+	DDAgentBin string // DDAgentBin will be "" for Agent5 scenarios
 }
 
 // NewDefaultAgentConfig returns a configuration with the default values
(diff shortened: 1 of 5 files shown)


config/agent.go
```diff
@@ -117,16 +120,27 @@ func NewDefaultAgentConfig() *AgentConfig {
...
+	}
 
 	var stdout, stderr bytes.Buffer
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
+	// Copying all environment variables to child process
+	// Windows: Required, so the child process can load DLLs, etc.
+	// Linux:   Optional, but will make use of DD_HOSTNAME and DOCKER_DD_AGENT if they exist
+	osEnv := os.Environ()
+	cmd.Env = append(cmd.Env, osEnv...)
```
AlexJF: If `osEnv` contains an invalid `PYTHONPATH` the agent 5 command will use that one instead of `defaultDDAgentPyEnv`. We should probably append `cmd.Env` to `osEnv`?

PR #384, opened 2018-02-26
Compute sublayers for non root toplevel spans

bmermet:
Until now sublayers were only computed for the root span of each trace. We now compute it for all non leaf top level spans of the trace. The purpose being to fix some gaps in the data currently shown in the UI.

This is implemented by extracting all relevant subtraces in each trace and computing the sublayers for each of theses subtraces individually.

One consequence of this change is that now the complexity of computing sublayers is no longer `O(s)` with `s` the number of spans, but `O(s * t)` with `s` the number of spans and `t` the number of top level spans. Which may have a very significant performance impact on some pathological traces.

It has been deployed on staging and I haven't noticed any significant change on the CPU footprint from  trace-agent.

### DIFF
--- a/model/trace.go
+++ b/model/trace.go
@@ -122,3 +122,89 @@ func (t Trace) APITrace() *APITrace {
 		EndTime:   end,
 	}
 }
+
+// Subtrace represents the combination of a root and the trace consisting of all its descendant spans
+type Subtrace struct {
+	Root  *Span
+	Trace Trace
+}
+
+// spaAndAncestors is an internal type used by ExtractTopLevelSubtraces
+type spanAndAncestors struct {
+	Span      *Span
+	Ancestors []*Span
+}
+
+// element and queue implement a very basic FIFO used to do a BFS on a trace
+type element struct {
+	SpanAndAncestors *spanAndAncestors
+	Next             *element
+}
+
+type queue struct {
+	first *element
+	last  *element
+}
+
+func (q *queue) Push(s *spanAndAncestors) {
+	e := &element{s, nil}
+	if q.first == nil {
+		q.first = e
+		q.last = e
+		return
+	}
+	q.last.Next = e
+	q.last = e
+}
+
+func (q *queue) Pop() *spanAndAncestors {
+	if q.first == nil {
+		return nil
+	}
+	r := q.first.SpanAndAncestors
+	q.first = q.first.Next
+	return r
+}
+
+// ExtractTopLevelSubtraces extracts all subtraces rooted in a toplevel span
+// ComputeTopLevel should be called before
+func (t Trace) ExtractTopLevelSubtraces(root *Span) []Subtrace {
+	if root == nil {
+		return []Subtrace{}
+	}
+	childrenMap := t.ChildrenMap()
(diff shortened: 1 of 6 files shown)


Gopkg.lock
```diff
@@ -25,7 +25,8 @@
 [[projects]]
```
LotharSee: Gabriel updated the way we package vendors in https://github.com/DataDog/datadog-trace-agent/pull/379

So it will need a bit of change to also commit the vendor changes. I'd suggest to make the `testify` upgrade an independent preliminary PR (as its diff will be noisy).

model/trace.go
```diff
@@ -122,3 +122,87 @@ func (t Trace) APITrace() *APITrace {
 		EndTime:   end,
 	}
 }
+
+// Subtrace represents the combination of a root and the trace consisting of all its descendant spans
```
gbbr: You missed a word here.
> Subtrace represents the combination of a root **span** and the trace consisting of all its descendant spans

model/trace.go
```diff
@@ -122,3 +122,87 @@ func (t Trace) APITrace() *APITrace {
...
+type Subtrace struct {
+	Root  *Span
+	Trace Trace
+}
+
+// spaAndAncestors is an internal type used by ExtractTopLevelSubtraces
```
gbbr: Here there is also a typo. It would be nice to explain (for devs) what this type is, in the comment, as opposed to writing that it is internal, which is obvious from the fact that it's unexported.

model/trace.go
```diff
@@ -122,3 +122,87 @@ func (t Trace) APITrace() *APITrace {
...
+func (s *stack) Pop() *spanAndAncestors {
+	if s.head == nil {
+		return nil
+	}
+	value := s.head.SpanAndAncestors
+	s.head = s.head.Next
+	return value
+}
+
+// ExtractTopLevelSubtraces extracts all subtraces rooted in a toplevel span
```
gbbr: Please add a comma at the end of this line and a period at the end of the second.

model/trace.go
```diff
@@ -122,3 +122,87 @@ func (t Trace) APITrace() *APITrace {
...
+func (t Trace) ExtractTopLevelSubtraces(root *Span) []Subtrace {
+	if root == nil {
+		return []Subtrace{}
+	}
+	childrenMap := t.ChildrenMap()
+	subtraces := []Subtrace{}
+
+	visited := make(map[*Span]bool, len(t))
+	subtracesMap := make(map[*Span][]*Span)
+	next := stack{}
```
gbbr: I don't feel very strongly about this but in Go it's generally nicer as:
```
var next stack
```
Which has the same outcome.

model/trace.go
bmermet: ✅

model/trace.go
bmermet: ✅

model/trace.go
bmermet: ✅

model/trace.go
bmermet: That looks nicer indeed

PR #385, opened 2018-02-26
[hostname][test] make hostname logic testable

truthbk:
The hostname logic is pretty important, but also very simple, so I'm not convinced this PR is necessary. The goal of the PR is to make the `hostname` detection logic testable by allowing us to mock the `exec.Cmd` calls in the same way the go sources test it (see: https://golang.org/src/os/exec/exec_test.go). 

Making the logic testable implies making some changes to code that may not be justified, so feel free to close the PR or make any suggestions. I just wanted to make sure the changes earlier commited in #383 could be tested if so desired.

### DIFF
--- a/config/agent.go
+++ b/config/agent.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"bytes"
 	"errors"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


gbbr:
Closing for now. Please re-open if you are still interested to merge this and I am happy to get it in. There are conflicts now.

PR #387, opened 2018-02-28
Do not compute stats for non top-level spans

bmermet:
This PR removes stats for non top-level spans since these stats are not surfaced in the UI.

This will however have an effect on resources which are both top-level and non top-level depending on the context. Since only the stats of the top-level ones will be computed and displayed.


cmd/trace-agent/concentrator.go
```diff
@@ -90,6 +90,10 @@ func (c *Concentrator) Add(t processedTrace) {
 	c.mu.Lock()
 
 	for _, s := range t.WeightedTrace {
+		// We do not compute stats for non top level spans since this is not surfaced in the UI
```
ufoot: :+1:

PR #393, opened 2018-03-05
cmd/trace-agent: remove exit channels in favor of http.Server and context.Context

gbbr:
In this change:

* Removes `exit` channel from StoppableListener and instead just uses [`(*http.Server).Shutdown`](https://golang.org/pkg/net/http/#Server.Shutdown) from the standard library (introduced in Go 1.8)
* Removes `exit` channel from the agent and instead uses the more idiomatic cancellation context.
* Adds the Windows build to CI.

### DIFF
--- a/cmd/trace-agent/agent_test.go
+++ b/cmd/trace-agent/agent_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"runtime"
@@ -31,17 +32,14 @@ func TestWatchdog(t *testing.T) {
 	defaultMux := http.DefaultServeMux
 	http.DefaultServeMux = http.NewServeMux()
 
-	exit := make(chan struct{})
-	agent := NewAgent(conf, exit)
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	agent := NewAgent(ctx, conf)
 
 	defer func() {
-		close(agent.exit)
+		cancelFunc()
 		// We need to manually close the receiver as the Run() func
 		// should have been broken and interrupted by the watchdog panic
-		close(agent.Receiver.exit)
-		// we need to wait more than on second (time for StoppableListener.Accept
-		// to acknowledge the connection has been closed)
-		time.Sleep(2 * time.Second)
+		agent.Receiver.Stop()
 		http.DefaultServeMux = defaultMux
 	}()
 
@@ -129,8 +127,9 @@ func BenchmarkAgentTraceProcessingWithWorstCaseFiltering(b *testing.B) {
 }
 
 func runTraceProcessingBenchmark(b *testing.B, c *config.AgentConfig) {
-	exit := make(chan struct{})
-	agent := NewAgent(c, exit)
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+	agent := NewAgent(ctx, c)
 	log.UseLogger(log.Disabled)
 
 	b.ResetTimer()
@@ -143,8 +142,9 @@ func runTraceProcessingBenchmark(b *testing.B, c *config.AgentConfig) {
 func BenchmarkWatchdog(b *testing.B) {
 	conf := config.NewDefaultAgentConfig()
 	conf.APIKey = "apikey_2"
-	exit := make(chan struct{})
-	agent := NewAgent(conf, exit)
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+	agent := NewAgent(ctx, conf)
 
 	b.ResetTimer()
 	b.ReportAllocs()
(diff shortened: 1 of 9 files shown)


cmd/trace-agent/listener.go
```diff
@@ -9,28 +9,27 @@ import (
 	log "github.com/cihub/seelog"
 )
 
-// StoppableListener wraps a regular TCPListener with an exit channel so we can exit cleanly from the Serve() loop of our HTTP server
-type StoppableListener struct {
-	exit      chan struct{}
+// RateLimitedListener wraps a regular TCPListener with an exit channel so we can exit cleanly from the Serve() loop of our HTTP server
```
bmermet: You can remove reference to the exit channel, since it is no longer true.

cmd/trace-agent/main.go
```diff
@@ -24,14 +25,14 @@ import (
 )
 
 // handleSignal closes a channel to exit cleanly from routines
-func handleSignal(exit chan struct{}) {
+func handleSignal(onCancel func()) {
```
bmermet: I don't like the name`onCancel`, it makes me think it's triggered by a cancel event when it's the other way around. `onSignal` or `cancelFunc` would seem like more natural names to me.

cmd/trace-agent/receiver.go
```diff
@@ -128,8 +127,7 @@ func (r *HTTPReceiver) Listen(addr, logExtra string) error {
 		return fmt.Errorf("cannot listen on %s: %v", addr, err)
 	}
 
-	stoppableListener, err := NewStoppableListener(listener, r.exit,
-		r.conf.ConnectionLimit)
+	stoppableListener, err := NewRateLimitedListener(listener, r.conf.ConnectionLimit)
```
bmermet: s/stoppableListener/rateLimitedListener/

cmd/trace-agent/receiver.go
```diff
@@ -152,12 +150,18 @@ func (r *HTTPReceiver) Listen(addr, logExtra string) error {
...
+		r.server.Serve(stoppableListener)
 	}()
 
 	return nil
 }
 
+func (r *HTTPReceiver) Stop() error {
+	expiry := time.Now().Add(20 * time.Second) // give it 20 seconds
+	ctx, _ := context.WithDeadline(context.Background(), expiry)
+	return r.server.Shutdown(ctx)
```
bmermet: Seems like a much better way to Stop existing connections than before 👍

cmd/trace-agent/listener.go
gbbr: 👍

cmd/trace-agent/main.go
gbbr: Agreed. Changed to `onSignal`.

cmd/trace-agent/receiver.go
gbbr: Ooopsie...

cmd/trace-agent/receiver_test.go
```diff
@@ -63,7 +63,7 @@ func TestReceiverRequestBodyLength(t *testing.T) {
 	go receiver.Run()
 
 	defer func() {
-		close(receiver.exit)
+		receiver.Stop()
 		// we need to wait more than on second (time for StoppableListener.Accept
 		// to acknowledge the connection has been closed)
 		time.Sleep(2 * time.Second)
```
bmermet: Sorry, I didn't spot that on my first read. But is this Sleep still needed?

cmd/trace-agent/agent_test.go
```diff
@@ -31,14 +32,14 @@ func TestWatchdog(t *testing.T) {
...
+	agent := NewAgent(ctx, conf)
 
 	defer func() {
-		close(agent.exit)
+		cancelFunc()
 		// We need to manually close the receiver as the Run() func
 		// should have been broken and interrupted by the watchdog panic
-		close(agent.Receiver.exit)
+		agent.Receiver.Stop()
 		// we need to wait more than on second (time for StoppableListener.Accept
```
bmermet: Same thing here.

cmd/trace-agent/receiver_test.go
gbbr: I don't know to be honest. The tests pass on both `master` and this branch without it. Do you reckon I should remove it? We could always add it back if this test turns flaky because of it.

gbbr:
Removed the `time.Sleep` calls. Tests passed without them before, on `master` too. I think it's safe to not have them now. Perhaps we can address that later if the test turns out to be flaky.

cmd/trace-agent/receiver_test.go
bmermet: I think that approach makes sense, we're always better off with less code :)

PR #395, opened 2018-03-09
writer: inherit and use "skip_ssl_validation" setting from Agent v{5,6}

gbbr:
This should permit use-cases of running the agent behind a proxy such as `haproxy`, using the `skip_ssl_validation: true` setting in the config file.

### DIFF
--- a/config/merge_ini.go
+++ b/config/merge_ini.go
@@ -67,6 +67,13 @@ func mergeIniConfig(c *AgentConfig, conf *File) error {
 		} else if pURL != nil {
 			c.ProxyURL = pURL
 		}
+
+		switch m.Key("skip_ssl_validation").MustString("") {
+		case "yes", "true", "1":
+			c.SkipSSLValidation = true
+		case "no", "false", "0":
+			c.SkipSSLValidation = false
+		}
 	}
 
 	// [trace.api] section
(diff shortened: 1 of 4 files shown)


config/merge_yaml.go
```diff
@@ -18,10 +18,11 @@ import (
...
 type YamlAgentConfig struct {
-	APIKey   string `yaml:"api_key"`
-	HostName string `yaml:"hostname"`
-	LogLevel string `yaml:"log_level"`
-	Proxy    proxy  `yaml:"proxy"`
+	APIKey            string `yaml:"api_key"`
+	HostName          string `yaml:"hostname"`
+	LogLevel          string `yaml:"log_level"`
+	Proxy             proxy  `yaml:"proxy"`
+	SkipSSLValidation string `yaml:"skip_ssl_validation"`
```
LotharSee: You should make it a `*bool` (like `traceAgent.Enabled`) so that you don't have to test manually yes, true, 1, ...

gbbr:
@LotharSee done. PTAL.

config/merge_ini.go
```diff
@@ -67,6 +67,13 @@ func mergeIniConfig(c *AgentConfig, conf *File) error {
 		} else if pURL != nil {
 			c.ProxyURL = pURL
 		}
+
+		switch m.Key("skip_ssl_validation").MustString("") {
```
gbbr: This package also has a `Bool` method, but I wasn't sure whether to use it or not. I chose not to, because it wasn't used elsewhere either.

PR #398, opened 2018-03-14
filters: add a span tag replace filters

gbbr:
This change adds a new setting (only Agent 6 for now) which allows to
set regular expressions in order to parse tag values. This helps hide
sensitive information from the UI. A new configuration example is:

```
apm_config:
    enabled: true
    endpoint: <redacted>
    # allows the same tag multiple times
    # the placeholder should be a user defined string
    # this list of regex applies to all Spans of a trace
    # the format is a list of sequences where:
    # [ tag_name, search_regex, placeholder ]
    replace_tags:
      - ["*", "user_id/([^/]*)", "?"] # applies to all filters and tags
      - ["resource.name", "foo", "bar"] # applies to resources
      - ["http.url", "token/(.*)", "?"]
      - ["http.url", "guid", "[REDACTED]"]
      - ["custom.tag", "something", "something_else"]
```


gbbr:
Switching this to @LotharSee's suggestion to have keyed name values for each item.

gbbr:
@LotharSee I've changed this. Would be great if you could take a look.

config/merge_yaml.go
```diff
@@ -288,6 +306,38 @@ func readExponentialBackoffConfigYaml(yc queueablePayloadSender) backoff.Exponen
...
+		re, err := regexp.Compile(r)
+		if err != nil {
+			return md, fmt.Errorf("key %q: %s", key, err)
+		}
+		md[key][re] = str
+	}
+	return md, nil
+}
+
+func ParseReplaceRulesFromString(rules [][3]string) (map[string]map[*regexp.Regexp]string, error) {
```
LotharSee: We should move this function from here. Either to the test package where it is used, or by converting the tests to the new format.

gbbr: Done

filters/tag.go
```diff
@@ -0,0 +1,47 @@
+package filters
+
+import (
+	"regexp"
+
+	"github.com/DataDog/datadog-trace-agent/config"
+	"github.com/DataDog/datadog-trace-agent/model"
+)
+
+var _ Filter = (*tagReplacer)(nil)
```
LotharSee: What is the purpose of this line?

gbbr: It is a common pattern that shows that the structure implements the Filter interface, as well as tests that failing at compile time.

For the latter it may many times not be needed because other code might be already using this structure as that interface, in which case compiling will fail anyway. But I have made it a habit to do this because in many cases that doesn't happen. I don't feel strongly about it but I think it's a good thing to have.

It is used a lot in the standard library too. Check [this example](https://github.com/golang/go/blob/8f6e8f9efd433fc67ff1e6e1caa5bc8edcad38f6/src/math/big/intconv.go#L50).

config/agent.go
```diff
@@ -71,6 +72,14 @@ type AgentConfig struct {
...
 	Ignore map[string][]string
 
+	// ReplaceTags is used to filter out sensitive information from tag values.
+	// It maps tag keys to a set of replacements.
+	// TODO(x): Introduce into Agent5 ini config. Currently only supported in 6.
+	ReplaceTags map[string]map[*regexp.Regexp]string `json:"-"`
+
+	// RawReplaceTags holds the unparsed replace rules. They are here for expvar.
+	RawReplaceTags []*replaceRule `json:"ReplaceTags"`
+
```
LotharSee: What about using a structure to wrap these two?
Something similar to `replaceRule` could also hold the its built Regexp then a list of it would be enough to be stored in the config. It allows both to have have a single struct and place per rule, and to have something simpler than `map[string]map[*regexp.Regexp]string`.
As `tagReplacer.Keep` is iterating over the tags anyway, it doesn't change much to keep only a list.

That's a nitpick though.

gbbr: Yeah I wasn't too pleased about this either, I think I've iterated a bit too much over the code and it became messy. I also "copied" the PR from our internal ticket a little bit. I like your suggestion. I'm gonna try it out.

gbbr: Actually what you've suggested in here worked, and I've added. PTAL. And thanks, it helps to have a second pair of eyes. I wasn't very happy with the initial version but I think now it's looking better.

config/merge_yaml.go
```diff
@@ -288,6 +306,25 @@ func readExponentialBackoffConfigYaml(yc queueablePayloadSender) backoff.Exponen
 	return c
 }
 
+// parseReplaceRules takes a set of replace rules, validates and parses them,
+// returning the corresponding map containing compiled regular expressions.
+func parseReplaceRules(rules []*ReplaceRule) error {
```
gbbr: I will rename this to "compileReplaceRules" and change the docs too after your review. I don't want to do that while you are potentially looking :)

gbbr:
FYI the following config:
```
# ...
apm_config:
    # ...
    replace_tags:
         - name: "http.url"
           pattern: "guid"
           repl: "[REDACTED]"
         - name: "http.url"
           pattern: "(token/)([^/]*)"
           repl: "${1}[HIDDEN]"
         - name: "custom.tag"
           pattern: "something"
           repl: "something_else"
```
Tested on staging:
screen shot 2018-03-14 at 2 48 05 pm

gbbr:
Thanks for your help!

PR #401, opened 2018-03-19
filter: simplify filter interface

gbbr:


filters/resource.go
```diff
@@ -15,7 +15,8 @@ type resourceFilter struct {
 }
 
 // Keep returns true if Span.Resource doesn't match any of the filter's rules
-func (f *resourceFilter) Keep(root *model.Span, trace *model.Trace) bool {
+func (f *resourceFilter) Keep(trace *model.Trace) bool {
+	root := trace.GetRoot()
```
LotharSee: `GetRoot` is an expensive operation, that's why most of our code is actually passing around explicitly the `root` next to its trace (to avoid computing `GetRoot` all the time we need it), or using the processedTrace struct.

gbbr:
Good to know, thanks for explaining @LotharSee, should've trusted your judgement from the start 😄

gbbr:
And thanks for fixing this for me.

PR #402, opened 2018-03-19
config: add a poller for server-sent config

talwai:
This implements a Poller that periodically speaks to a Datadog config endpoint, and updates in-memory config for the Agent when necessary.

- `Poller` relies on an incrementing index when determining when to propagate changes to other pieces of the agent.

- As a proof of concept, the `TransactionSampler` is configured to listen for updates on `AnalyzedRateByService`

- `config/server_config.go` describes the structure of the server response. Extra fields in the response will be ignored when unmarshalling, so additional config can be overloaded into the response and handled in future agent versions.

- Optionally `Poller` will persist config in a user-specified path (default: `/var/run/datadog/trace_agent_config.gob` on *nix, but this needs to be discussed further). This should handle cold starts and also cases where the config server is unavailable for long periods


poller/poller_test.go
```diff
@@ -0,0 +1,64 @@
...
+	assert.NotNil(url)
+	assert.Nil(err)
+
+	p := &Poller{
+		defaultInterval, "", http.DefaultClient,
+		url.String(), make(chan *config.ServerConfig), "apikey_2", 0,
+	}
+
+	go func() {
+		for {
```
gbbr: I think this works with `for...range` too

poller/poller_test.go
```diff
@@ -0,0 +1,64 @@
...
+			select {
+			case conf := <-p.updates:
+				assert.Equal(conf.ModifyIndex, int64(1000))
+				assert.Equal(len(conf.AnalyzedRateByService), 1)
+				assert.Equal(conf.AnalyzedRateByService["web"], 1.0)
+
+				done <- struct{}{}
```
gbbr: Nit but `close(done)` is nicer IMO.

poller/poller.go
```diff
@@ -0,0 +1,170 @@
...
+			reportPersistFailed(err, []string{})
+		}
+	}
+
+	p.index = s.ModifyIndex
+	return nil
+}
+
+// Updates returns the channel to subscribe to for configuration updates
+func (p *Poller) Updates() chan *config.ServerConfig {
```
gbbr: Since this is exported I think it definitely is useful to specify the returned channel as receive only:
```go
func (p *Poller) Updates() <-chan *config.ServerConfig {
```

config/server_config.go
```diff
@@ -0,0 +1,30 @@
+package config
+
+import (
+	"encoding/gob"
+	"os"
+)
+
+// ServerConfig contains configuration options sent down by the Datadog API
+type ServerConfig struct {
```
gbbr: I would tend to recommend calling this `Server` instead of `ServerConfig` to avoid stutter, given that it'll be used as `config.Server`, which sounds better than `config.ServerConfig`.

poller/poller.go
```diff
@@ -0,0 +1,170 @@
...
+// Poller polls Datadog for configuration updates
+type Poller struct {
+	interval    time.Duration
+	persistPath string
+	client      *http.Client
+	endpoint    string
+	updates     chan *config.ServerConfig
+	apiKey      string
+
+	index int64
```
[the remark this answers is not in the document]
LotharSee: We should also probably use a more descriptive name than index. Maybe `version`?

config/server_config.go
```diff
@@ -0,0 +1,30 @@
...
+	ModifyIndex int64 `json:modify_index,omitempty`
+
+	AnalyzedRateByService map[string]float64 `json:analyzed_rate_by_service,omitempty`
+}
```
LotharSee: Suggestion: adding the config polling frequency itself to the config? It would allow the config backend to auto-adjust/protect itself.

poller/poller.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+	}
+
+	p.index = s.ModifyIndex
+	return nil
+}
+
+// Updates returns the channel to subscribe to for configuration updates
+func (p *Poller) Updates() chan *config.ServerConfig {
+	return p.updates
```
LotharSee: This implementation won't work with multiple components listening to config updates,  only one will get the update.

poller/poller.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	var s config.ServerConfig
+	err = json.NewDecoder(response.Body).Decode(&s)
+	if err != nil {
+		tags := []string{"error:decoding_error"}
+		return pollingError{err, tags}
+	}
+
+	if s.ModifyIndex > p.index {
```
LotharSee: We probably want the server-returned version to be a header, so that we don't need extra parsing / it is consistent with the request.

Also, we should consider that the server will return a `304` when the server-side version is the same. It gives us a fast lane (which will represent 99.9% of the time) for that function.

poller/poller.go
```diff
@@ -0,0 +1,170 @@
...
+		return pollingError{err, tags}
+	}
+
+	defer response.Body.Close()
+
+	tags := []string{"status_code:" + strconv.Itoa(response.StatusCode)}
+	switch response.StatusCode {
+	case http.StatusOK:
+		reportPollingSuccess(tags)
+	case http.StatusNotModified:
```
LotharSee: +1

gbbr:
I think we can close this.

PR #405, opened 2018-03-27
config: support overriding agent endpoint via env

LeoCavaille:
For internal use.

### DIFF
--- a/config/merge_env.go
+++ b/config/merge_env.go
@@ -26,6 +26,10 @@ func mergeEnv(c *AgentConfig) {
 		c.HostName = v
 	}
 
+	if v := os.Getenv("DD_APM_DD_URL"); v != "" {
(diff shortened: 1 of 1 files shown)


config/merge_env.go
```diff
@@ -26,6 +26,10 @@ func mergeEnv(c *AgentConfig) {
 		c.HostName = v
 	}
 
+	if v := os.Getenv("DD_TRACE_AGENT_ENDPOINT"); v != "" {
```
LotharSee: To kind of align it with its config option `apm_config.apm_dd_url`, good let's move to `DD_APM_DD_URL`. I'm not in love with that but that's the closest we will get.

LeoCavaille: fixed in aa67dd0

PR #406, opened 2018-04-09
Fix performance issue in quantile.WeighSummary

bmermet:
`rand.Seed(7337)` was called on every call to `probabilisticRound` which was both incorrect and  very slow since it was equivalent to having a single hard coded float to do all the roundings and calls to `rand.Seed` are very expensive.

### DIFF
--- a/quantile/weighted.go
+++ b/quantile/weighted.go
@@ -5,19 +5,27 @@ import (
 	"math/rand"
 )
 
+var randomFloats []float64
+
+func init() {
+	// generate a list of guaranteed random numbers for the probabilistic round
+	randomFloats = make([]float64, 0, 100)
+	rand.Seed(7337)
+	for i := 0; i < 100; i++ {
+		randomFloats = append(randomFloats, rand.Float64())
+	}
+}
+
 // WeightedSliceSummary associates a weight to a slice summary.
 type WeightedSliceSummary struct {
 	Weight float64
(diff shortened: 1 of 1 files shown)


quantile/weighted.go
```diff
@@ -5,19 +5,27 @@ import (
...
 
+var randomFloats []float64
+
+func init() {
+	// generate a list of guaranteed random numbers for the probabilistic round
+	randomFloats = make([]float64, 0, 100)
+	rand.Seed(7337)
+	for i := 0; i < 100; i++ {
+		randomFloats = append(randomFloats, rand.Float64())
+	}
```
LotharSee: May this be more efficient?

```
randomFloats = make([]float64, 100)
for i := 0; i < 100; i++ {
	randomFloats[i] = rand.Float64()
}
```

quantile/weighted.go
```diff
@@ -5,19 +5,27 @@ import (
 	"math/rand"
 )
 
+var randomFloats []float64
+
+func init() {
+	// generate a list of guaranteed random numbers for the probabilistic round
+	randomFloats = make([]float64, 100)
+	rand.Seed(7337)
```
LotharSee: Nitpick: to not affect the global `rand` seed, we could use `r:= New(NewSource(7337)) / r.Float64()`.

PR #407, opened 2018-04-09
quantizer: reduce cardinality of select/set

gbbr:


tylerbenson:
Makes me very grateful that we're doing this in the agent, rather than the clients.  👍

PR #409, opened 2018-04-11
build: Makefile defaults acting up due to extra space

truthbk:
The extra space was being included in the vars messing up the calls to `windmc` and `winres` when a  `TRACE_AGENT_VERSION` that didn't match semver `maj.min.patch` was being passed (this scenario could happen from omnibus).

Also, if we typically default to `0.99.0` if nothing was set, we should also default to `0.99.0` if a weird version (ie. branch name) is passed to the agent.


gbbr:
Closing for now. Please re-open if you are still interested to merge this and I am happy to do so.

PR #410, opened 2018-04-11
quantizer: reduce cardinality of select/set

gbbr:


gbbr:
One more thing that could help on this is to abort the sorting when we hit open braces (`(`), which means that a nested query follows. We can't handle that complexity right now.

palazzem:
We'll keep this PR on-hold, but leaving it here as a reference for a bit. Thanks @gbbr for the work!

PR #420, opened 2018-05-07
Add dedicated error sampling

bmermet:
This PR add a dedicated error sampler to the trace-agent. The goal of this is to keep more error traces (which are the most interesting ones).

With this PR the `trace-agent` will maintain two different score samplers that will each have their own maximum TPS. One for traces containing errors and one for other traces. That way the TPS of errors that the agent sends to the API will be independent from the TPS of traces containing no errors.

### DIFF
--- a/cmd/trace-agent/agent.go
+++ b/cmd/trace-agent/agent.go
@@ -43,6 +43,7 @@ type Agent struct {
 	Concentrator       *Concentrator
 	Filters            []filters.Filter
 	ScoreSampler       *Sampler
+	ErrorsScoreSampler *Sampler
 	PrioritySampler    *Sampler
 	TransactionSampler *TransactionSampler
 	TraceWriter        *writer.TraceWriter
@@ -86,6 +87,7 @@ func NewAgent(ctx context.Context, conf *config.AgentConfig) *Agent {
 	f := filters.Setup(conf)
 
 	ss := NewScoreSampler(conf)
+	ess := NewErrorsSampler(conf)
 	ps := NewPrioritySampler(conf, dynConf)
 	ts := NewTransactionSampler(conf, analyzedTransactionChan)
 	se := NewTraceServiceExtractor(serviceChan)
@@ -99,6 +101,7 @@ func NewAgent(ctx context.Context, conf *config.AgentConfig) *Agent {
 		Concentrator:       c,
 		Filters:            f,
 		ScoreSampler:       ss,
+		ErrorsScoreSampler: ess,
 		PrioritySampler:    ps,
 		TransactionSampler: ts,
 		TraceWriter:        tw,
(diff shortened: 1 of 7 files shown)


cmd/trace-agent/agent.go
```diff
@@ -320,3 +329,12 @@ func (a *Agent) watchdog() {
 
 	info.UpdatePreSampler(*a.Receiver.preSampler.Stats())
 }
+
+func traceContainsError(trace model.Trace) bool {
```
gbbr: There's a lot of functions which iterate over the entire trace (e.g. `GetRoot`, `ChildrenMap`, etc.). We should possibly only do this once when we receive the trace and compute various stats about the trace such as if it has errors, it's root, the children map and other things we need.

This is just a remark, no action needed here. Possibly a worthy improvement to be done in another PR or during a potential refactor.

cmd/trace-agent/sampler.go
```diff
@@ -31,6 +31,13 @@ func NewScoreSampler(conf *config.AgentConfig) *Sampler {
 	}
 }
 
+// NewErrorsSampler creates a new empty error sampler ready to be started
```
gbbr: Nit. Can we end this sentence with a period? Also, it would be helpful (at least to me) to know what an "errors sampler" is. Maybe this is the right place to write about that?

info/info.go
```diff
@@ -155,7 +156,7 @@ func publishSamplerInfo() interface{} {
 	return samplerInfo
 }
 
-// UpdatePrioritySamplerInfo updates internal stats about priority sampking
+// UpdatePrioritySamplerInfo updates internal stats about priority sampling
```
gbbr: Please put a period at the end of this sentence, and the next one further down.

sampler/coresampler.go
```diff
@@ -33,6 +33,18 @@ const (
 	defaultSignatureScoreSlope  float64       = 3
 )
 
+// EngineType represent the type of a sampler engine
+type EngineType int
```
gbbr: I like this a lot 👍 Gives for a much cleaner, easier to understand and read API.

sampler/scoresampler.go
```diff
@@ -20,13 +20,25 @@ import (
...
 	s := &ScoreEngine{
-		Sampler: newSampler(extraRate, maxTPS),
+		Sampler:    newSampler(extraRate, maxTPS),
+		engineType: NormalScoreEngineType,
+	}
+
+	return s
+}
+
+// NewErrorEngine returns an initialized Sampler
```
gbbr: I think we can do a bit better with this comment. 😄

cmd/trace-agent/agent.go
bmermet: Agreed, we should definitely think of consolidating all this logic to iterate only once on the trace. I'll keep that for another PR though.

sampler/scoresampler.go
bmermet: Indeed :)

PR #425, opened 2018-05-24
Change agent config for transactions

bmermet:
This change updates the configuration to support the new format for transactions. The deprecated format is still temporarily supported but will be overridden if the new one is provided. This change makes transaction filtering based both on service name and operation name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,7 @@ VERSION_MINOR := $(if $(VERSION_MINOR),$(VERSION_MINOR), 0)
 VERSION_PATCH := $(if $(VERSION_PATCH),$(VERSION_PATCH), 0)
 
 install:
-	# prepares all dependencies by running the 'deps' task, generating
-	# versioning information and installing the binary.
+	# generate versioning information and installing the binary.
 	go generate ./info
 	go install ./cmd/trace-agent
 
--- a/README.md
+++ b/README.md
@@ -75,13 +75,19 @@ it manually.
 
 ## Development
 
-Pre-requisites:
-- `go` 1.9+
+First, make sure Go 1.9+ is installed. You can do this by following the steps on the [official website](https://golang.org/dl/).
 
-
-Build and run from source:
-- Run `make install` to install the `trace-agent` binary in $GOPATH/bin
-- You may then run it with `trace-agent --config PATH_TO_YOUR_DATADOG_CONFIG_FILE`
+Then, download (or update) the latest source by running:
+```bash
+go get -u github.com/DataDog/datadog-trace-agent/...
+```
+Navigate to the download location and install the `trace-agent` binary in $GOPATH/bin:
+```bash
+cd $GOPATH/src/github.com/DataDog/datadog-trace-agent
+make install
+```
+You may now run the agent using `trace-agent --config /etc/dd-agent/datadog.conf`, considering
+that $GOPATH/bin has been added to your $PATH.
 
 
 ## Testing
--- a/info/info.go
+++ b/info/info.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 	"os"
 	"regexp"
-	"strconv"
 	"strings"
 	"sync"
 	"text/template"
@@ -332,7 +331,7 @@ func Info(w io.Writer, conf *config.AgentConfig) error {
 	if host == "0.0.0.0" {
 		host = "127.0.0.1" // [FIXME:christian] not fool-proof
 	}
-	url := "http://localhost:" + strconv.Itoa(conf.ReceiverPort) + "/debug/vars"
+	url := fmt.Sprintf("http://%s:%d/debug/vars", conf.ReceiverHost, conf.ReceiverPort)
 	client := http.Client{Timeout: 3 * time.Second}
 	resp, err := client.Get(url)
 	if err != nil {
--- a/model/normalizer.go
+++ b/model/normalizer.go
@@ -49,11 +49,11 @@ func (s *Span) Normalize() error {
 		return fmt.Errorf("span.normalize: `Name` too long (max %d chars): %s", MaxNameLen, s.Name)
 	}
 	// name shall comply with Datadog metric name normalization
-	var ok bool
-	s.Name, ok = normMetricNameParse(s.Name)
+	name, ok := normMetricNameParse(s.Name)
 	if !ok {
 		return fmt.Errorf("span.normalize: invalid `Name`: %s", s.Name)
 	}
+	s.Name = name
 
 	// Resource
 	if s.Resource == "" {
--- a/model/normalizer_test.go
+++ b/model/normalizer_test.go
@@ -65,6 +65,20 @@ func TestNormalizeName(t *testing.T) {
 	}
 }
 
+func TestNormalizeNameFailure(t *testing.T) {
+	invalidNames := []string{
+		"",   // Empty.
+		"/",  // No alphanumerics.
+		"//", // Still no alphanumerics.
+		strings.Repeat("x", MaxNameLen+1), // Too long.
+	}
+	s := testSpan()
+	for _, v := range invalidNames {
+		s.Name = v
+		assert.Error(t, s.Normalize())
+	}
+}
+
 func TestNormalizeResourcePassThru(t *testing.T) {
 	s := testSpan()
 	before := s.Resource


config/agent.go
```diff
@@ -119,8 +120,9 @@ func NewDefaultAgentConfig() *AgentConfig {
 		MaxConnections:   200, // in practice, rarely goes over 20
 		WatchdogInterval: time.Minute,
 
-		Ignore:                make(map[string][]string),
-		AnalyzedRateByService: make(map[string]float64),
+		Ignore: make(map[string][]string),
```
AlexJF: Not sure if this triggers the linter or not but I guess it should be aligned with the ones below?

config/merge_ini.go
```diff
@@ -339,3 +364,11 @@ func readProxyURL(m *ini.Section) (*url.URL, error) {
 
 	return url.Parse(path)
 }
+
+func parseAnalyzedSpanFormat(name string) (string, string, error) {
+	splits := strings.Split(name, "/")
+	if len(splits) != 2 {
```
AlexJF: According to this: https://github.com/DataDog/datadog-trace-agent/blob/231444ce1d42a433eebeef5f2a26f0c777b04d6f/model/tags.go#L276, a service name can include a '/'as long as it doesn't start with one. Should we change this logic to allow multiple parts but consider all but the last as belonging to the service name?

bmermet: Ooops, good catch ! Thank you :)
It might be simpler to just use a different character, like a "|" so that there is no ambiguity.

LotharSee: Good catch, I'd suggest to simply go with a symbol forbidden both in the service and operation name.

config/agent.go
bmermet: I'm not sure why, but my gofmt keeps aligning it like this.

cmd/trace-agent/transaction_sampler.go
```diff
@@ -7,37 +7,106 @@ import (
 )
 
 // TransactionSampler extracts and samples analyzed spans
```
gbbr: Can we explain how this works? I for one, without domain knowledge, have no idea what this does. It would be great if newcomers or other contributors could understand how this works by reading the documentation.

cmd/trace-agent/transaction_sampler.go
```diff
@@ -7,37 +7,106 @@ import (
...
+	// inspect the WeightedTrace so that we can identify top-level spans
+	for _, span := range t.WeightedTrace {
+		if s.shouldAnalyze(span) {
+			s.analyzed <- span.Span
+		}
+	}
+}
+
+func (s *transactionSampler) shouldAnalyze(span *model.WeightedSpan) bool {
+	if !span.TopLevel {
```
LotharSee: The new config format should exclude this condition, only relying on the explicit configuration. That'd also safely allow analyzing non-top-level spans if wanted.

config/merge_yaml.go
```diff
@@ -221,8 +222,23 @@ func mergeYamlConfig(agentConf *AgentConfig, yc *YamlAgentConfig) error {
 	agentConf.StatsWriterConfig = readStatsWriterConfigYaml(yc.TraceAgent.StatsWriter)
 	agentConf.TraceWriterConfig = readTraceWriterConfigYaml(yc.TraceAgent.TraceWriter)
 
-	// undocumented
-	agentConf.AnalyzedRateByService = yc.TraceAgent.AnalyzedRateByService
+	// undocumented deprecated
+	agentConf.AnalyzedRateByServiceLegacy = yc.TraceAgent.AnalyzedRateByServiceLegacy
```
LotharSee: If the deprecated config format is used, let's log a warning asking to move to the new one?

cmd/trace-agent/transaction_sampler.go
```diff
@@ -7,37 +7,106 @@ import (
...
+	if len(conf.AnalyzedSpansByService) > 0 {
+		return newTransactionSampler(conf.AnalyzedSpansByService, analyzed)
+	}
+	if len(conf.AnalyzedRateByServiceLegacy) > 0 {
+		return newLegacyTransactionSampler(conf.AnalyzedRateByServiceLegacy, analyzed)
+	}
+	return &disabledTransactionSampler{}
+}
+
+type disabledTransactionSampler struct {
```
gbbr: Can we remove the new line here and on the empty functions below?

cmd/trace-agent/transaction_sampler.go
```diff
@@ -7,37 +7,106 @@ import (
 )
 
 // TransactionSampler extracts and samples analyzed spans
-type TransactionSampler struct {
+type TransactionSampler interface {
+	// Enabled tells if the transaction analysis is enabled
+	Enabled() bool
+	// Add extracts analyzed spans and send them to its `analyzed` channel
```
gbbr: I find it a bit annoying that I am not able to follow and understand the code in the agent. Is there any chance we could write documentation in such a way that it is clear what these things do? For example:
```
// Add extracts analyzed spans and send them to its `analyzed` channel
```
This documentation says nothing about the method signature. Where does it extract them from? Does it return them? What `analyzed` channel?

cmd/trace-agent/transaction_sampler.go
```diff
@@ -7,37 +7,106 @@ import (
...
+	Add(processedTrace)
```
gbbr: My suggestion is to remove the `Enabled` method from this interface and instead just manage it internally. If the sampler is disabled, simply make the call a no-op. This method is only used in one place and is not needed IMO.

On the other hand, I think that now that we have an interface here, it should make sense on its own, outside of context. We could do this with a simple change:
```go
// transactionSampler finds interesting spans in a trace based on implementation-specific criteria
// and returns them via an output.
type transactionSampler interface {
    // process processes the given trace and returns the span of interest via the out channel.
    process(trace processedTrace, out chan<- *model.Span)
}
```

cmd/trace-agent/transaction_sampler.go
bmermet: That makes sense. However we'll also need to update the backend to make sure we support that case properly. As it is done now, non-root spans will all be considered as "uncomplete traces".

cmd/trace-agent/transaction_sampler.go
gbbr: > The benefit of the Enabled method is that checking it before calling Add allows to avoid creating a goroutine when it is not enabled (for most people).

Yes, you're right about this. I have considered this as well. But maybe we can create the goroutine inside the (what is now) `Add` method instead, when the sampler is enabled. How about that?

> Also for the naming, Add is consistent with the method of the Sampler type which does a similar job.

I'm not sure I see the value of that. Why do we need them to be consistent? Also, I don't get the feeling that we are adding something to something else here, but more like we are iterating over the trace and returning spans selectively.

What I'm trying to say is that if we made this into an interface, it shouldn't be only because we want it to be "switchable" and it should make sense on it's own too. I'm not so strongly opinionated about the naming, and I'd be ok with keeping it if you feel strongly about it, but I do think we need to add some sort of output or return, otherwise it's too confusing as it is now.

WDYT about the two points above?

gbbr: P.S. I like the new documentation, by reading that, perhaps even `Extract` could be an alternative to `Add`.

cmd/trace-agent/transaction_sampler.go
```diff
@@ -6,38 +6,103 @@ import (
...
-// TransactionSampler extracts and samples analyzed spans
-type TransactionSampler struct {
+// TransactionSampler filters and samples interesting spans in a trace based
+// on the criteria and rates set in the config and returns them via an output
+// channel.
+type TransactionSampler interface {
+	// Enabled tells if the transaction analysis is enabled
+	Enabled() bool
+	// Add extracts matching spans from the given trace and returns them via the output channel.
+	Add(processedTrace)
```
gbbr: _Copying over the comment as it got lost due to recent changes_

> The benefit of the Enabled method is that checking it before calling Add allows to avoid creating a goroutine when it is not enabled (for most people).

Yes, you're right about this. I have considered that as well. But maybe we can create the goroutine inside the (what is now) `Add` method instead, when the sampler is enabled. How about that?

> Also for the naming, Add is consistent with the method of the Sampler type which does a similar job.

I'm not sure I see the value of that. Why do we need them to be consistent? Also, I don't get the feeling that we are adding something to something else here, but more like we are iterating over the trace and returning spans selectively. If you read the documentation that you wrote (which I find great), the word "Add" doesn't really have its place in there IMHO. Perhaps "Extract" would be a good alternative?

What I'm trying to say is that if we made this into an interface, it shouldn't be only because we want it to be "switchable" and it should make sense on it's own too. I'm not so strongly opinionated about the naming, and I'd be ok with keeping it if you feel strongly about it, but I do think we need to add some sort of output or return, otherwise it's too confusing as it is now. 

WDYT about the two points above?

bmermet: > Yes, you're right about this. I have considered that as well. But maybe we can create the goroutine inside the (what is now) Add method instead, when the sampler is enabled. How about that?

I don't like that approach so much for two main reasons:
- It is not visible from the interface that the work is going to happen in another goroutine (in fact different implementations might do it differently), which makes it harder to reason about the code.
- It makes the code harder to test since you have to add synchronisation to do simple asserts.

If you think strongly about this `Enabled` method, I'd rather go in another direction among:
- Keep this `enabled` boolean information in the `Agent` type itself and keep the if statement similar to what it is now. (this has the downside of making the `Agent` type even more bloated than it is now)
- Just create a goroutine for each trace processed (but I'm not sure of the performance implications this would have).


On your other point, I agree that adding the `out` channel as parameter or returning a result makes it clearer. I'll change that. And about the naming, I agree that `Extracts` better reflects what it does and keeping it consistent with other samplers is not that valuable.

gbbr: > I don't like that approach so much for two main reasons: [...]

I agree with this. What if we were to just move the `defer` call into the `Extract` method, and instead just have:

```go
go a.TransactionSampler.Extract(pt, out)
```
I think it's pretty safe to say that the cost of that goroutine will be negligible in case of a disabled transaction sampler. But perhaps I'm wrong. Do you think this will be any better? If you think not, then let's keep the `Enabled` method. I agree that bloating the agent struct further is not the best option.

cmd/trace-agent/transaction_sampler.go
```diff
@@ -6,38 +6,99 @@ import (
...
-	analyzed              chan *model.Span
-	analyzedRateByService map[string]float64
+// TransactionSampler filters and samples interesting spans in a trace based
+// on the criteria and rates set in the config and returns them via an output
+// channel.
+type TransactionSampler interface {
+	// Enabled tells if the transaction analysis is enabled
+	Enabled() bool
+	// Extracts extracts matching spans from the given trace and returns them via the `out` channel.
+	Extract(t processedTrace, out chan *model.Span)
```
gbbr: Can we please change this method signature to make it explicit that the `out` channel is send-only?
```go
Extract(t processedTrace, out chan<- *model.Span)
```

cmd/trace-agent/transaction_sampler.go
gbbr: I have a better, slightly dirty idea, which can help us get rid of the `Enabled` method and also not start the goroutine in the case when the sampler is disabled:
```go
if _, ok := a.TransactionSampler.(*disabledTransactionSampler); !ok {
    // ...
}
```
WDYT?

cmd/trace-agent/agent.go
```diff
@@ -297,7 +299,7 @@ func (a *Agent) Process(t model.Trace) {
 	if a.TransactionSampler.Enabled() {
```
gbbr: _copying this over since it got lost again_

I have a better, slightly dirty idea, which can help us get rid of the `Enabled` method and also not start the goroutine in the case when the sampler is disabled:
```go
if _, ok := a.TransactionSampler.(*disabledTransactionSampler); !ok {
    // ...
}
```
(a second alternative solution would be to simply not initialize the interface, thus only having to check for `nil`. But in that case the very need for an interface becomes arguable)

I'm fine with any of the above as long as we manage to remove the `Enabled` method.

WDYT?

bmermet: I find neither solutions great :(

I don't really see the value of doing a cast to find out the underlying type of the interface vs making it explicit with the `Enabled` method.

Having a `nil` type that we need to check before every use seems quite dangerous to me.

Actually I think another way to solve that problem could simply be to run the `Extract` method in the same goroutine as the other samplers. I don't think there is any benefit to run these operations concurrently.

bmermet: I've submitted a change showing that approach, what do you think?

gbbr: I can't really say I have anything against your arguments. As for your latest change, I'm happy that the method is now gone, but do you think it's ok that the transaction sampler now runs sequentially in the same goroutine as the other samplers? If yes, then it's all good as far as I'm concerned.

PR #426, opened 2018-05-28
README: clarify install instructions.

### LINKED ISSUE #415: Missing information in trace agent build
Dont we need to do a go get -d ./... before running make build to install trace agent for installing the dependencies in our go path ?
Also the make install needs to be run with the $GOPATH only right?

This is for mac

gbbr:
Closes #415

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,13 +75,19 @@ it manually.
 
 ## Development
 
-Pre-requisites:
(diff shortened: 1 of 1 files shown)


gbbr:
Thanks @ufoot! You're right, I've added the *NIX location.

PR #429, opened 2018-06-01
cmd/trace-agent: add a metric when the agent starts.

vlad-mh:
**main: add a metric when the agent starts.**

This metrics allows to track the restarts of the trace-agent,
independently from whatever supervisor it's using.

### DIFF
--- a/cmd/trace-agent/main.go
+++ b/cmd/trace-agent/main.go
@@ -203,6 +203,11 @@ func runAgent(ctx context.Context) {
 		die("cannot configure dogstatsd: %v", err)
 	}
 
+	// count the number of times the agent started
+	statsd.Client.Count("datadog.trace_agent.started", 1, []string{
+		"version:" + info.Version,
+	}, 1)
+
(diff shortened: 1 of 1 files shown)


vlad-mh:
Just a rebase on master.

cmd/trace-agent/main.go
```diff
@@ -203,6 +203,11 @@ func runAgent(ctx context.Context) {
 		die("cannot configure dogstatsd: %v", err)
 	}
 
+	// Track the application start
```
gbbr: This comment is a bit odd, can you please change it? "Track the application start" sounds like you're tracking something about the start, but maybe we can be more accurate.
```go
// count the number of times the agent started
```
Also, I find that throughout the project, and commnunity-wide, inline comments are usually lower-cased. I don't have a strong preference as long as we are consistent throughout the project. If the comment starts with an uppercase letter however, I feel it's a phrase and a bit odd when it doesn't end with a dot.

PR #430, opened 2018-06-01
cmd/trace-agent: improvements to trace/transaction writing pipeline.

AlexJF:
This PR reworks the logic around trace and transaction writing to fix some identified issues.

# Trace splitting

Traces are no longer arbitrarily split when writing. This introduced some issues later on in the server side processing pipeline, specially around priority sampling.

Instead, `MaxSpansPerPayload` is now a soft-limit. Whenever possible, payloads will respect said limit. In case a single trace + its transactions have more spans than that limit, then we still attempt to send that trace (on its own) to Datadog's API. There's a chance that the API will reject it if it goes over the input size limit but, if that happens, that means the trace is several MB big which is not normal anyway so the issue should be handled at the client side.

# Traces and transactions are written atomically

Before, traces and transactions were written asynchronously which could result in scenarios where a trace would be flushed to the API without some of its transactions which would only be flushed later on  in a consecutive payload. This introduced some issues during server-side processing where we might mistakenly not tag some transactions as having been sampled because we did not know about its associated trace.

With this change, traces and their respective transactions are treated "atomically" so that it can never happen that such a split occurs.

### DIFF
--- a/cmd/trace-agent/agent.go
+++ b/cmd/trace-agent/agent.go
@@ -53,8 +53,7 @@ type Agent struct {
 	ServiceExtractor   *TraceServiceExtractor
 	ServiceMapper      *ServiceMapper
 
-	sampledTraceChan        chan *model.Trace
-	analyzedTransactionChan chan *model.Span
+	sampledTraceChan chan *writer.SampledTrace
 
 	// config
 	conf    *config.AgentConfig
@@ -73,8 +72,7 @@ func NewAgent(ctx context.Context, conf *config.AgentConfig) *Agent {
 
 	// inter-component channels
 	rawTraceChan := make(chan model.Trace, 5000) // about 1000 traces/sec for 5 sec, TODO: move to *model.Trace
-	sampledTraceChan := make(chan *model.Trace)
-	analyzedTransactionChan := make(chan *model.Span)
+	sampledTraceChan := make(chan *writer.SampledTrace)
 	statsChan := make(chan []model.StatsBucket)
 	serviceChan := make(chan model.ServicesMetadata, 50)
 	filteredServiceChan := make(chan model.ServicesMetadata, 50)
@@ -94,29 +92,28 @@ func NewAgent(ctx context.Context, conf *config.AgentConfig) *Agent {
 	ts := NewTransactionSampler(conf)
 	se := NewTraceServiceExtractor(serviceChan)
 	sm := NewServiceMapper(serviceChan, filteredServiceChan)
-	tw := writer.NewTraceWriter(conf, sampledTraceChan, analyzedTransactionChan)
+	tw := writer.NewTraceWriter(conf, sampledTraceChan)
 	sw := writer.NewStatsWriter(conf, statsChan)
 	svcW := writer.NewServiceWriter(conf, filteredServiceChan)
 
 	return &Agent{
-		Receiver:                r,
-		Concentrator:            c,
-		Filters:                 f,
-		ScoreSampler:            ss,
-		ErrorsScoreSampler:      ess,
-		PrioritySampler:         ps,
-		TransactionSampler:      ts,
-		TraceWriter:             tw,
-		StatsWriter:             sw,
-		ServiceWriter:           svcW,
-		ServiceExtractor:        se,
-		ServiceMapper:           sm,
-		sampledTraceChan:        sampledTraceChan,
-		analyzedTransactionChan: analyzedTransactionChan,
-		conf:    conf,
-		dynConf: dynConf,
-		ctx:     ctx,
-		die:     die,
+		Receiver:           r,
+		Concentrator:       c,
+		Filters:            f,
+		ScoreSampler:       ss,
+		ErrorsScoreSampler: ess,
+		PrioritySampler:    ps,
+		TransactionSampler: ts,
+		TraceWriter:        tw,
+		StatsWriter:        sw,
+		ServiceWriter:      svcW,
+		ServiceExtractor:   se,
+		ServiceMapper:      sm,
+		sampledTraceChan:   sampledTraceChan,
+		conf:               conf,
+		dynConf:            dynConf,
+		ctx:                ctx,
+		die:                die,
 	}
 }
 
@@ -295,12 +292,17 @@ func (a *Agent) Process(t model.Trace) {
 			sampled = s.Add(pt) || sampled
 		}
 
+		var sampledTrace writer.SampledTrace
+
 		if sampled {
-			a.sampledTraceChan <- &pt.Trace
+			sampledTrace.Trace = &pt.Trace
 		}
 
-		// Transactions extraction
-		a.TransactionSampler.Extract(pt, a.analyzedTransactionChan)
+		sampledTrace.Transactions = a.TransactionSampler.Extract(pt)
+
+		if !sampledTrace.Empty() {
+			a.sampledTraceChan <- &sampledTrace
+		}
 	}()
 }
 
--- a/cmd/trace-agent/transaction_sampler.go
+++ b/cmd/trace-agent/transaction_sampler.go
@@ -6,12 +6,10 @@ import (
 	"github.com/DataDog/datadog-trace-agent/sampler"
 )
 
-// TransactionSampler filters and samples interesting spans in a trace based
-// on the criteria and rates set in the config and returns them via an output
-// channel.
+// TransactionSampler filters and samples interesting spans in a trace based on implementation specific criteria.
 type TransactionSampler interface {
-	// Extracts extracts matching spans from the given trace and returns them via the `out` channel.
-	Extract(t processedTrace, out chan<- *model.Span)
+	// Extract extracts matching spans from the given trace and returns them.
+	Extract(t processedTrace) []*model.Span
 }
 
 // NewTransactionSampler creates a new empty transaction sampler
@@ -27,7 +25,9 @@ func NewTransactionSampler(conf *config.AgentConfig) TransactionSampler {
 
 type disabledTransactionSampler struct{}
 
-func (s *disabledTransactionSampler) Extract(t processedTrace, out chan<- *model.Span) {}
+func (s *disabledTransactionSampler) Extract(t processedTrace) []*model.Span {
+	return nil
+}
 
 type transactionSampler struct {
 	analyzedSpansByService map[string]map[string]float64
@@ -39,14 +39,18 @@ func newTransactionSampler(analyzedSpansByService map[string]map[string]float64)
 	}
 }
 
-// Add extracts analyzed spans and send them to its `analyzed` channel
-func (s *transactionSampler) Extract(t processedTrace, out chan<- *model.Span) {
+// Extract extracts analyzed spans and returns them as a slice
+func (s *transactionSampler) Extract(t processedTrace) []*model.Span {
+	var transactions []*model.Span
+
 	// inspect the WeightedTrace so that we can identify top-level spans
 	for _, span := range t.WeightedTrace {
 		if s.shouldAnalyze(span) {
-			out <- span.Span
+			transactions = append(transactions, span.Span)
 		}
 	}
+
+	return transactions
 }
 
 func (s *transactionSampler) shouldAnalyze(span *model.WeightedSpan) bool {
@@ -70,14 +74,18 @@ func newLegacyTransactionSampler(analyzedRateByService map[string]float64) *lega
 	}
 }
 
-// Add extracts analyzed spans and send them to the `out` channel
-func (s *legacyTransactionSampler) Extract(t processedTrace, out chan<- *model.Span) {
+// Extract extracts analyzed spans and returns them as a slice
+func (s *legacyTransactionSampler) Extract(t processedTrace) []*model.Span {
+	var transactions []*model.Span
+
 	// inspect the WeightedTrace so that we can identify top-level spans
 	for _, span := range t.WeightedTrace {
 		if s.shouldAnalyze(span) {
-			out <- span.Span
+			transactions = append(transactions, span.Span)
 		}
 	}
+
+	return transactions
 }
 
 // shouldAnalyze tells if a span should be considered as analyzed
--- a/writer/trace_writer.go
+++ b/writer/trace_writer.go
@@ -17,14 +17,24 @@ import (
 	"github.com/golang/protobuf/proto"
 )
 
+// SampledTrace represents the result of a trace sample operation.
+type SampledTrace struct {
+	Trace        *model.Trace
+	Transactions []*model.Span
+}
+
+// Empty returns true if this SampledTrace has no data.
+func (s *SampledTrace) Empty() bool {
+	return s.Trace == nil && len(s.Transactions) == 0
+}
+
 // TraceWriter ingests sampled traces and flushes them to the API.
 type TraceWriter struct {
-	stats          info.TraceWriterInfo
-	hostName       string
-	env            string
-	conf           writerconfig.TraceWriterConfig
-	InTraces       <-chan *model.Trace
-	InTransactions <-chan *model.Span
+	stats    info.TraceWriterInfo
+	hostName string
+	env      string
+	conf     writerconfig.TraceWriterConfig
+	in       <-chan *SampledTrace
 
 	traces        []*model.APITrace
 	transactions  []*model.Span
@@ -34,7 +44,7 @@ type TraceWriter struct {
 }
 
 // NewTraceWriter returns a new writer for traces.
-func NewTraceWriter(conf *config.AgentConfig, InTraces <-chan *model.Trace, InTransactions <-chan *model.Span) *TraceWriter {
+func NewTraceWriter(conf *config.AgentConfig, in <-chan *SampledTrace) *TraceWriter {
 	writerConf := conf.TraceWriterConfig
 	log.Infof("Trace writer initializing with config: %+v", writerConf)
 
@@ -46,8 +56,7 @@ func NewTraceWriter(conf *config.AgentConfig, InTraces <-chan *model.Trace, InTr
 		traces:       []*model.APITrace{},
 		transactions: []*model.Span{},
 
-		InTraces:       InTraces,
-		InTransactions: InTransactions,
+		in: in,
 
 		BaseWriter: *NewBaseWriter(conf, "/api/v0.2/traces", func(endpoint Endpoint) PayloadSender {
 			return NewCustomQueuablePayloadSender(endpoint, writerConf.SenderConfig)
@@ -109,15 +118,8 @@ func (w *TraceWriter) Run() {
 
 	for {
 		select {
-		case trace := <-w.InTraces:
-			if trace == nil {
-				continue
-			}
-			w.handleTrace(trace)
-		case transaction := <-w.InTransactions:
-			// no need for lock for now as flush is sequential
-			// TODO: async flush/retry
-			w.transactions = append(w.transactions, transaction)
+		case sampledTrace := <-w.in:
+			w.handleSampledTrace(sampledTrace)
 		case <-flushTicker.C:
 			log.Debug("Flushing current traces")
 			w.flush()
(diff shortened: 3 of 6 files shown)


writer/trace_writer.go
```diff
@@ -139,52 +136,77 @@ func (w *TraceWriter) Stop() {
...
-		splitIndex := len(*trace) - spanOverflow
-		log.Debugf("Splitting trace at index %d", splitIndex)
-		// Set the spans of the split trace to the ones over the split index
-		splitTrace = (*trace)[splitIndex:]
-		// Set the spans of the original to the ones below the split index so it ends up with a non-overflowing amount
-		// of traces
-		truncatedTrace := (*trace)[:splitIndex]
-		trace = &truncatedTrace
+		// If current number of spans in buffer reached the limit, flush
+		w.flushDueToMaxSpansPerPayload()
```
LotharSee: If we receive a large trace with `len(trace) > MaxSpansPerPayload` when the buffer is empty, we will call `flushDueToMaxSpansPerPayload` for nothing.
Technically it will do nothing (as `flush` test the buffer size) but we will get a wrong `w.stats.FlushMaxSpans` stat.

Actually a non-interesting nitpick?

writer/trace_writer.go
```diff
@@ -139,52 +136,77 @@ func (w *TraceWriter) Stop() {
...
+	}
+
+	w.handleTrace(payload.Trace)
+	w.handleTransactions(payload.Transactions)
+
+	if spansToAdd > w.conf.MaxSpansPerPayload {
+		// If what we just added already goes over the limit, report this but lets carry on and flush
+		atomic.AddInt64(&w.stats.SingleMaxSpans, 1)
+		log.Warnf("Found single trace with more spans than limit per payload of %d spans", w.conf.MaxSpansPerPayload)
```
LotharSee: This could become noisy which not being actionable nor technically bad. Should we remove it?

cmd/trace-agent/agent.go
```diff
@@ -295,12 +292,16 @@ func (a *Agent) Process(t model.Trace) {
...
 		if sampled {
-			a.sampledTraceChan <- &pt.Trace
+			writerPayload.Trace = &pt.Trace
 		}
 
 		// Transactions extraction
-		a.TransactionSampler.Extract(pt, a.analyzedTransactionChan)
+		writerPayload.Transactions = a.TransactionSampler.Extract(pt)
+
+		a.traceChan <- &writerPayload
```
bmermet: If the trace is not sampled and no transactions are analyzed, writerPayload will be empty. We might just as well not send it in the channel

writer/trace_writer.go
```diff
@@ -274,6 +298,8 @@ func (w *TraceWriter) updateInfo() {
 	w.statsClient.Count("datadog.trace_agent.trace_writer.bytes", int64(twInfo.Bytes), nil, 1)
 	w.statsClient.Count("datadog.trace_agent.trace_writer.retries", int64(twInfo.Retries), nil, 1)
 	w.statsClient.Count("datadog.trace_agent.trace_writer.errors", int64(twInfo.Errors), nil, 1)
+	w.statsClient.Count("datadog.trace_agent.trace_writer.flush_max_spans", int64(twInfo.FlushMaxSpans), nil, 1)
```
LotharSee: How relevant / useful are these new metrics? If we don't picture good use-cases ; it'd save some resources to not report them.

writer/trace_writer.go
```diff
@@ -17,14 +17,19 @@ import (
 	"github.com/golang/protobuf/proto"
 )
 
+// TraceWriterPayload represents the result of a trace sample operation that should be written by a TraceWriter.
+type TraceWriterPayload struct {
```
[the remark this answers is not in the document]
AlexJF: Yeah good call. Went with `TraceSample`.

writer/trace_writer.go
AlexJF: Nice catch

writer/trace_writer.go
AlexJF: Removed, we can rely on the metric to notice these.

cmd/trace-agent/agent.go
AlexJF: Yeah, lets save some channel contention for sure 😄

writer/trace_writer.go
AlexJF: I added them because I missed them while debugging the recent issues (try to understand just how many splits were being done). But yeah now that we fixed splitting, `flush_max_spans` probably won't be that useful.

`single_max_spans` might still be useful though. Having a non-0 value is a hint that something is wrong as such traces are probably not very useful. And when trace payloads get rejected by the API it can give us an idea if traces just have too many spans (this would be non-0) or if they have few but very heavy spans (this would be 0). Don't feel strongly about it though.

cmd/trace-agent/agent.go
```diff
@@ -295,12 +292,19 @@ func (a *Agent) Process(t model.Trace) {
 			sampled = s.Add(pt) || sampled
 		}
 
+		traceSample := writer.TraceSample{}
```
gbbr: Let's make this:
```go
var traceSample writer.TraceSample
```
or
```go
traceSample := writer.TraceSample{
    Transactions: a.TransactionSampler.Extract(pt),
}
```
Whichever you prefer.

cmd/trace-agent/agent.go
```diff
@@ -295,12 +292,19 @@ func (a *Agent) Process(t model.Trace) {
...
-			a.sampledTraceChan <- &pt.Trace
+			traceSample.Trace = &pt.Trace
 		}
 
 		// Transactions extraction
-		a.TransactionSampler.Extract(pt, a.analyzedTransactionChan)
+		traceSample.Transactions = a.TransactionSampler.Extract(pt)
+
+		// Only send if there's something interesting
+		if traceSample.Trace != nil || traceSample.Transactions != nil {
```
gbbr: I wonder if it would be nicer to have a method on `TraceSample` called `Empty` and making this:
```go
if !traceSample.Empty() {
   // ...
}
```

cmd/trace-agent/transaction_sampler.go
```diff
@@ -7,11 +7,10 @@ import (
 )
 
 // TransactionSampler filters and samples interesting spans in a trace based
-// on the criteria and rates set in the config and returns them via an output
-// channel.
+// on the criteria and rates set in the config and adds them to a TraceWriterPayload.
 type TransactionSampler interface {
-	// Extracts extracts matching spans from the given trace and returns them via the `out` channel.
-	Extract(t processedTrace, out chan<- *model.Span)
+	// Extracts extracts matching spans from the given trace and returns them via a slice
```
gbbr: There is a typo here (s/Extracts/Extract). I think this should suffice:
```go
// Extract extracts matching spans from the given trace and returns them.
```

writer/trace_writer.go
```diff
@@ -17,14 +17,19 @@ import (
 	"github.com/golang/protobuf/proto"
 )
 
+// TraceSample represents the result of a trace sample operation.
+type TraceSample struct {
```
gbbr: This name makes me think that this structure holds only a sample (part) of a trace and could be misleading. I did see @LotharSee's comment about the naming, but I'm thinking maybe `SampledTrace` would be a more specific name. What do you think?

```go
// A SampledTrace holds the result of a trace sampling operation.
type SampledTrace struct {
     // Trace specifies the trace that was sampled.
    Trace *model.Trace

    // Transactions specifies the spans which were extracted by the TransactionSampler.
    Transactions []*model.Span
}
```

writer/trace_writer.go
```diff
@@ -17,14 +17,19 @@ import (
...
-	hostName       string
-	env            string
-	conf           writerconfig.TraceWriterConfig
-	InTraces       <-chan *model.Trace
-	InTransactions <-chan *model.Span
+	stats     info.TraceWriterInfo
+	hostName  string
+	env       string
+	conf      writerconfig.TraceWriterConfig
+	InSamples <-chan *TraceSample
```
gbbr: I'd tend to think the same here, the word "sample" makes me think of only a part of the whole. This is the main input channel of the writer and it should be more than sufficient to call it `in`. It also doesn't need to be an exported field since it isn't used anywhere outside this package. I've heard the expression "the 'in' channel" used many times, and since we only have one I think it would be clear enough.

writer/trace_writer.go
```diff
@@ -34,7 +39,7 @@ type TraceWriter struct {
 }
 
 // NewTraceWriter returns a new writer for traces.
-func NewTraceWriter(conf *config.AgentConfig, InTraces <-chan *model.Trace, InTransactions <-chan *model.Span) *TraceWriter {
+func NewTraceWriter(conf *config.AgentConfig, InSamples <-chan *TraceSample) *TraceWriter {
```
gbbr: IMHO it would nicer to call this channel `in`. It's the main input channel of the writer and it's clear what it's sending. There is also no need to capitalise variable names.

writer/trace_writer.go
```diff
@@ -139,52 +136,77 @@ func (w *TraceWriter) Stop() {
 	w.BaseWriter.Stop()
 }
 
-func (w *TraceWriter) handleTrace(trace *model.Trace) {
-	if len(*trace) == 0 {
-		log.Debugf("Ignoring 0-length trace")
+func (w *TraceWriter) handleTraceSample(sample *TraceSample) {
+	if sample == nil {
```
gbbr: We don't really need this do we?

writer/trace_writer.go
```diff
@@ -139,52 +136,77 @@ func (w *TraceWriter) Stop() {
...
 		return
 	}
 
-	log.Tracef("Handling new trace with %d spans: %v", len(*trace), trace)
+	trace := sample.Trace
+	transactions := sample.Transactions
+
+	if trace == nil && transactions == nil {
```
gbbr: We could use `sample.Empty()` here again if you follow my suggestion.

writer/trace_writer.go
```diff
@@ -139,52 +136,77 @@ func (w *TraceWriter) Stop() {
...
 
-	if w.spansInBuffer == w.conf.MaxSpansPerPayload {
-		log.Debugf("Flushing because we reached max per payload")
-		// If current number of spans in buffer reached the limit, flush
-		w.flush()
-	} else if w.spansInBuffer > w.conf.MaxSpansPerPayload {
-		// Should never happen due to overflow detection above but just in case
-		panic("Number of spans in buffer went over the limit")
+func (w *TraceWriter) handleTransactions(transactions []*model.Span) {
+	if transactions == nil {
```
gbbr:
```go
if len(transactions) == 0 {
```
Will have the same effect.

But then again, you can remove the entire `if` statement because the `for` loop will not iterate and the addition will be with `0`.

writer/trace_writer.go
```diff
@@ -139,52 +136,77 @@ func (w *TraceWriter) Stop() {
...
+
+	if transactions != nil {
+		spansToAdd += len(transactions)
+	}
+
+	// If we have stuff pending and adding the new data would overflow max spans per payload, force a flush
+	if w.spansInBuffer > 0 {
+		spanOverflow := w.spansInBuffer + spansToAdd - w.conf.MaxSpansPerPayload
+
+		if spanOverflow > 0 {
```
gbbr: This is a bit too "smart", it would be much clearer to do:
```go
if w.spansInBuffer + spansToAdd > w.conf.MaxSpansPerPayload {
```
Better keep it simple.

writer/trace_writer.go
```diff
@@ -139,52 +136,77 @@ func (w *TraceWriter) Stop() {
...
+
+	if trace != nil {
+		spansToAdd += len(*trace)
+	}
+
+	if transactions != nil {
+		spansToAdd += len(transactions)
+	}
+
+	// If we have stuff pending and adding the new data would overflow max spans per payload, force a flush
```
gbbr: Please put this comment inside the `if` statement body. It would also be nice if we wouldn't use the word "stuff" 😄 Something like:
```go
// check if the new data would overflow
```

writer/trace_writer.go
```diff
@@ -139,52 +136,77 @@ func (w *TraceWriter) Stop() {
...
+		w.transactions = append(w.transactions, transaction)
 	}
+
+	w.spansInBuffer += len(transactions)
+}
+
+func (w *TraceWriter) flushDueToMaxSpansPerPayload() {
+	atomic.AddInt64(&w.stats.FlushMaxSpans, 1)
+	log.Debugf("Flushing because we reached max per payload")
+	// If current number of spans in buffer reached the limit, flush
```
gbbr: I think this comment might be leftover from previous code, can you please remove it?

writer/trace_writer.go
```diff
@@ -139,52 +136,77 @@ func (w *TraceWriter) Stop() {
...
+	// If we have stuff pending and adding the new data would overflow max spans per payload, force a flush
+	if w.spansInBuffer > 0 {
+		spanOverflow := w.spansInBuffer + spansToAdd - w.conf.MaxSpansPerPayload
+
+		if spanOverflow > 0 {
+			w.flushDueToMaxSpansPerPayload()
+		}
+	}
+
+	w.handleTrace(sample.Trace)
```
gbbr: I know you didn't name this but `handleTrace` is not descriptive enough. What if we would call them `appendTrace` and `appendTransactions`, since that's what they're doing?

writer/trace_writer.go
gbbr: Sorry for being so picky, I normally don't care so much. It's just because it's open source.

writer/trace_writer.go
```diff
@@ -139,52 +141,65 @@ func (w *TraceWriter) Stop() {
...
+	if sampledTrace.Empty() {
+		log.Debug("Ignoring empty trace writer payload")
 		return
 	}
 
-	log.Tracef("Handling new trace with %d spans: %v", len(*trace), trace)
+	trace := sampledTrace.Trace
+	transactions := sampledTrace.Transactions
+
+	var spansToAdd int
```
gbbr: Would you mind renaming this to `count` or `n`? It's pretty obvious what we are counting here, given the context. I know you don't like short names but I feel that in general in Go it is preferred like this and I would love it if we were to stay consistent with not only the rest of the project but the Go community in general, as much as possible.

I think `n` is the generally used variable for counting items, both in computer science and in mathematics and it will be clear to everyone.

writer/trace_writer_test.go
```diff
@@ -88,18 +106,27 @@ func TestTraceWriter_BigTraceHandling(t *testing.T) {
 		fixtures.RandomTrace(3, 1),
 	}
 
-	// When sending those 6 traces
+	var testTransactions []model.Span
+
+	// When sending those 7 traces, all with transactions for the root span
```
gbbr: I think the grammar here might be a bit off or I'm too tired. I don't understand what happens when sending "_those 7 traces_".

Rather than making a story that has to be read throughout the test and then pieced together, I'd tend to think it's more preferred and easy to make sense of it we describe the action that happens, as opposed to giving only a piece of the story which is out of context without the rest, i.e.:

```go
// send the test traces down the trace channel with the root span attached as a transaction
```

writer/trace_writer_test.go
```diff
@@ -112,9 +139,26 @@ func TestTraceWriter_BigTraceHandling(t *testing.T) {
 		numSpans += len(trace)
 	}
 
-	expectedNumPayloads := int(math.Ceil(float64(numSpans) / float64(traceWriter.conf.MaxSpansPerPayload)))
+	var expectedNumPayloads int
+	var pendingSpans int
+
+	for _, trace := range testTraces {
```
gbbr: This test is super complicated. I'm not exactly sure what's going on here. It's pretty hard to understand. You are counting the number of spans in a trace and adding them to the `pendingSpans` variable, which looks completely incorrect because of the `len(trace) + 1`. Why `+ 1`? When looking closer at the code, it's perhaps because of transaction you add in the root above? Either way, it's very confusing and it's already a pretty big test.

I think that we need to spend a bit more time here to simplify this.

writer/trace_writer_test.go
```diff
@@ -25,7 +25,7 @@ func TestTraceWriter_TraceHandling(t *testing.T) {
 	assert := assert.New(t)
 
 	// Given a trace writer, its incoming channel and the endpoint that receives the payloads
-	traceWriter, traceChannel, testEndpoint, _ := testTraceWriter()
+	traceWriter, payloadChannel, testEndpoint, _ := testTraceWriter()
```
gbbr: This is called a `traceChannel` in the other test. Please choose one, whichever you prefer.

writer/trace_writer_test.go
```diff
@@ -124,48 +168,57 @@ func TestTraceWriter_UpdateInfoHandling(t *testing.T) {
...
 
 	traceWriter.Start()
 
-	expectedNumPayloads := int64(0)
-	expectedNumSpans := int64(0)
-	expectedNumTraces := int64(0)
-	expectedNumBytes := int64(0)
-	expectedNumErrors := int64(0)
-	expectedMinNumRetries := int64(0)
+	var expectedNumPayloads int64
```
gbbr: Let's group these please, like elsewhere in the project:
```go
var (
    expectedNumPayloads int64
    expectedNumSpans int64
    expectedNumTraces int64
    expectedNumBytes int64
    expectedNumErrors int64
    expectedMinNumRetries int64
)
```

writer/trace_writer_test.go
gbbr: I would like to suggest to merge both these tests into one single test (let's say `TestTraceWriter`) and use [sub-tests](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks) for each case, one called "normal" and one called "large", or whatever you prefer. I'd also like to suggest removing all these `for` loops and instead adding a new function to the `fixtures` package which generates `SampledTrace`s for you. We could call it `RandomSampledTrace` to make it analogue to the already existing `RandomTrace`.

Additionally, you could declare another method on the `SampledTrace` type that returns the total length (spans). Since this method won't be used in the code, you can just declare it in the test file as it will work just fine.

The goal here is to considerably shorten these tests and remove unnecessary duplication.

AlexJF: Simplified a little bit. Can't do much further than this without starting to lose coverage though 😞

writer/trace_writer_test.go
AlexJF: Done

cmd/trace-agent/agent.go
```diff
@@ -295,12 +292,19 @@ func (a *Agent) Process(t model.Trace) {
...
 		}
 
+		var sampledTrace writer.SampledTrace
+
 		if sampled {
-			a.sampledTraceChan <- &pt.Trace
+			sampledTrace.Trace = &pt.Trace
 		}
 
 		// Transactions extraction
```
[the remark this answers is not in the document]
AlexJF: Done

writer/trace_writer.go
```diff
@@ -17,14 +17,24 @@ import (
...
 
+// SampledTrace represents the result of a trace sample operation.
+type SampledTrace struct {
+	Trace        *model.Trace
+	Transactions []*model.Span
+}
+
+// Empty returns true if this TraceSample has no data.
+func (s *SampledTrace) Empty() bool {
+	return s == nil || (s.Trace == nil && len(s.Transactions) == 0)
```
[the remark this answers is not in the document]
gbbr: You're right! This is interesting and I have seen this used in our Go tracer as well (in the old version). Although, I have never encountered this out there in the wild. I would assume the reason is probably that if you ever end up passing a `nil` pointer, it's probably a sign of bad design. I mean other than some struct's defaults, why would you? If it ever _has_ to be `nil` then I think an explicit check is better than this. That's my assumption anyway, but I'm happy to be proven wrong.

As for `handleSampledTrace`, I think it should never receive `nil` and you could definitely remove the check, but I don't mind keeping it if you will . You already check for that in the `Process` method of the agent. If someone in the future passes `nil` as a value, they might have malicious intentions and a `panic` is ok I'd say 😈

AlexJF:
Squashed and rebased on master

PR #431, opened 2018-06-05
model/normalizer: Enforce valid utf-8 during normalization.

AlexJF:
An attempt at enforcing valid utf8 strings for span contents after normalization.

It works but has a non-negligible impact which I couldn't avoid by writing the same code with several variations (my own loop, etc...). Most of the time and allocations are actually inside the Rune handling code of the stdlib it seems.

Impact on my MBP:

Impact with no invalid utf-8 strings:
```
$ benchstat old.txt new-valid.txt
name             old time/op    new time/op    delta
Normalization-4     549ns ± 8%     671ns ± 0%  +22.13%  (p=0.016 n=5+4)

name             old alloc/op   new alloc/op   delta
Normalization-4      792B ± 0%      792B ± 0%     ~     (all equal)

name             old allocs/op  new allocs/op  delta
Normalization-4      9.00 ± 0%      9.00 ± 0%     ~     (all equal)
```

Impact with invalid utf-8 strings in Resource, Type and Meta:
```
$ benchstat old.txt new-invalid.txt
name             old time/op    new time/op    delta
Normalization-4     549ns ± 8%    1344ns ± 0%  +144.68%  (p=0.008 n=5+5)

name             old alloc/op   new alloc/op   delta
Normalization-4      792B ± 0%     1296B ± 0%   +63.64%  (p=0.008 n=5+5)

name             old allocs/op  new allocs/op  delta
Normalization-4      9.00 ± 0%     17.00 ± 0%   +88.89%  (p=0.008 n=5+5)
```

### DIFF
--- a/model/normalizer.go
+++ b/model/normalizer.go
@@ -1,10 +1,13 @@
 package model
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
+	"unicode/utf8"
 
 	log "github.com/cihub/seelog"
 )
@@ -56,6 +59,7 @@ func (s *Span) Normalize() error {
 	s.Name = name
 
 	// Resource
+	s.Resource = toUTF8(s.Resource)
 	if s.Resource == "" {
 		return errors.New("span.normalize: empty `Resource`")
 	}
@@ -98,8 +102,20 @@ func (s *Span) Normalize() error {
 	// ParentID set on the client side, no way of checking
 
 	// Type
+	s.Type = toUTF8(s.Type)
 	if len(s.Type) > MaxTypeLen {
 		return fmt.Errorf("span.normalize: `Type` too long (max %d chars): %s", MaxTypeLen, s.Type)
+	}
+
+	for k, v := range s.Meta {
+		utf8K := toUTF8(k)
+
+		if k != utf8K {
+			delete(s.Meta, k)
+			k = utf8K
+		}
+
+		s.Meta[k] = toUTF8(v)
 	}
 
 	// Environment
@@ -225,4 +241,29 @@ func normMetricNameParse(name string) (string, bool) {
 	}
 
 	return string(res), true
+}
+
+// toUTF8 forces the string to utf-8 by replacing illegal character sequences with the utf-8 replacement character.
+func toUTF8(s string) string {
+	if utf8.ValidString(s) {
+		// if string is already valid utf8, return it as-is. Checking validity is cheaper than blindly rewriting.
+		return s
+	}
+
+	in := strings.NewReader(s)
+	var out bytes.Buffer
+	out.Grow(len(s))
+
+	for {
+		r, _, err := in.ReadRune()
+		if err != nil {
+			// note: by contract, if `in` contains non-valid utf-8, no error is returned. Rather the utf-8 replacement
+			// character is returned. Therefore, the only error should usually be io.EOF indicating end of string.
+			// If any other error is returned by chance, we quit as well, outputting whatever part of the string we
+			// had already constructed.
+			return out.String()
+		}
+
+		out.WriteRune(r)
+	}
 }
--- a/model/normalizer_test.go
+++ b/model/normalizer_test.go
@@ -290,3 +290,81 @@ func TestIsValidStatusCode(t *testing.T) {
 	assert.False(isValidStatusCode("600"))
 	assert.False(isValidStatusCode("Invalid status code"))
 }
+
+func TestNormalizeInvalidUTF8(t *testing.T) {
+	invalidUTF8 := "test\x99\x8f"
+
+	t.Run("service", func(t *testing.T) {
+		assert := assert.New(t)
+
+		span := testSpan()
+		span.Service = invalidUTF8
+
+		err := span.Normalize()
+
+		assert.Nil(err)
+		assert.Equal("test", span.Service)
+	})
+
+	t.Run("resource", func(t *testing.T) {
+		assert := assert.New(t)
+
+		span := testSpan()
+		span.Resource = invalidUTF8
+
+		err := span.Normalize()
+
+		assert.Nil(err)
+		assert.Equal("test��", span.Resource)
+	})
+
+	t.Run("name", func(t *testing.T) {
+		assert := assert.New(t)
+
+		span := testSpan()
+		span.Name = invalidUTF8
+
+		err := span.Normalize()
+
+		assert.Nil(err)
+		assert.Equal("test", span.Name)
+	})
+
+	t.Run("type", func(t *testing.T) {
+		assert := assert.New(t)
+
+		span := testSpan()
+		span.Type = invalidUTF8
+
+		err := span.Normalize()
+
+		assert.Nil(err)
+		assert.Equal("test��", span.Type)
+	})
+
+	t.Run("meta", func(t *testing.T) {
+		assert := assert.New(t)
+
+		span := testSpan()
+		span.Meta = map[string]string{
+			invalidUTF8: "test1",
+			"test2":     invalidUTF8,
+		}
+
+		err := span.Normalize()
+
+		assert.Nil(err)
+		assert.EqualValues(map[string]string{
+			"test��": "test1",
+			"test2":  "test��",
+		}, span.Meta)
+	})
+}
+
+func BenchmarkNormalization(b *testing.B) {
+	b.ReportAllocs()
+
+	for i := 0; i < b.N; i += 1 {
+		testSpan().Normalize()
+	}
+}


model/normalizer.go
```diff
@@ -226,3 +266,7 @@ func normMetricNameParse(name string) (string, bool) {
 
 	return string(res), true
 }
+
+func forceUtf8(s string) (string, error) {
```
gbbr: I'd say it might be (insignificantly) more efficient to have only one decoder per function call. Maybe it would be best to inline this, but I don't feel strongly about it.

If you decide to keep it here, can we just call it `toUTF8` instead? I feel it would be much clearer. Force seems like a word that we would use with an operation that has the potential to fail but we are ignoring the error. I don't think this is the case.

model/normalizer_test.go
```diff
@@ -290,3 +290,91 @@ func TestIsValidStatusCode(t *testing.T) {
 	assert.False(isValidStatusCode("600"))
 	assert.False(isValidStatusCode("Invalid status code"))
 }
+
+func TestNormalizeInvalidUTF8(t *testing.T) {
+	invalidUtf8 := "test\x99\x8f"
```
gbbr: I like how you called it `UTF8` in the test function title, can we do the same everywhere else. `s/Utf8/UTF8`?

AlexJF:
Done, didn't know about this benchstat thing, useful :)

model/normalizer.go
```diff
@@ -226,3 +258,22 @@ func normMetricNameParse(name string) (string, bool) {
...
 	return string(res), true
 }
+
+func toUTF8(s string) (string, error) {
+	in := bytes.NewBuffer([]byte(s))
+	res := bytes.NewBuffer(make([]byte, 0, len(s)))
+
+	for {
+		r, _, err := in.ReadRune()
+
```
gbbr: Would you terribly mind removing all the empty lines? As far as I see, in general the practice is to have a statement, and then immediately after it's error handling, so I'd love it if it would be possible to change this as:
```go
r, _, err := in.ReadRune()
if err != nil {
	return "", err
}
if err == io.EOF {
	return res.String(), nil
}
```
Also, I've recently discovered [`strings.Builder`](https://golang.org/pkg/strings/#Builder) which I think should be more efficient.

model/normalizer.go
```diff
@@ -101,6 +108,31 @@ func (s *Span) Normalize() error {
 	if len(s.Type) > MaxTypeLen {
 		return fmt.Errorf("span.normalize: `Type` too long (max %d chars): %s", MaxTypeLen, s.Type)
 	}
+	utf8Type, err := toUTF8(s.Type)
+	if err != nil {
+		return fmt.Errorf("span.normalize: error converting type to utf-8: %s", err.Error())
+	}
+	s.Type = utf8Type
+
+	// Meta
```
gbbr: Can we please drop this comment?

model/normalizer.go
AlexJF: Oh great tip regarding the strings.Builder (also replaced the input with a strings.Reader). The new benchstat (compare it with the one in the PR description, using bytes.Buffer):

```
$ benchstat old.txt new.txt
name             old time/op    new time/op    delta
Normalization-4     549ns ± 8%    1097ns ± 3%  +99.67%  (p=0.008 n=5+5)

name             old alloc/op   new alloc/op   delta
Normalization-4      792B ± 0%      848B ± 0%   +7.07%  (p=0.008 n=5+5)

name             old allocs/op  new allocs/op  delta
Normalization-4      9.00 ± 0%     15.00 ± 0%  +66.67%  (p=0.008 n=5+5)
```

model/normalizer.go
AlexJF: Done

AlexJF:
Oh `strings.Builder` apparently requires go 1.10. Do we need compatibility with 1.9 or can we update the CI?

gbbr:
It's definitely not a problem to update CI. AFAIK we are aiming to have the same version in CI as we do in [Agent 6](https://github.com/DataDog/datadog-agent).

@derekwbrown or @masci can you guys please confirm that it's ok for us to have a requirement for go1.10? (tldr; to use `strings.Builder`)

masci:
Ping @olivielpeau for review

AlexJF:
My 2 cents: 

We can't really control what gets sent to us and given that traces are encoded via protobuf and this was merged (https://github.com/golang/protobuf/pull/499), then any non-utf8 traces would be immediately dropped at the agent during serialization time when/if we upgrade to latest versions.

This might be acceptable behaviour or not, that's more of a product decision. I'll just comment that, from a user's standpoint, it seems much more friendly to let the traces go through but clearly mark strings with invalid utf8 with `�`. But as this PR shows, this comes with a certain cost.

Another thing we could do is have this utf8 normalization happen later on in the pipeline, after sampling. That could reduce the performance impact but would mean we'll have traces flying through our pipeline with invalid utf8. This might not be a problem currently but if we introduce more parts later on it's something to keep in mind.

model/normalizer.go
```diff
@@ -59,6 +61,11 @@ func (s *Span) Normalize() error {
 	if s.Resource == "" {
 		return errors.New("span.normalize: empty `Resource`")
 	}
+	utf8Resource, err := toUTF8(s.Resource)
+	if err != nil {
+		return fmt.Errorf("span.normalize: error converting resoutce to utf-8: %s", err.Error())
```
LotharSee: typo at `resoutce`.

model/normalizer.go
```diff
@@ -101,6 +108,30 @@ func (s *Span) Normalize() error {
...
+			k = utf8K
+		}
+
+		utf8V, err := toUTF8(v)
+		if err != nil {
+			return fmt.Errorf("span.normalize: error converting meta value to utf-8: %s", err.Error())
+		}
+
+		s.Meta[k] = utf8V
+	}
```
LotharSee: Afair when `Span.Normalize` returns an error, we drop the full span / trace (no stats). It means it can break significantly the product just for a small unicode issue on an meta attribute. And that might be more frequent that we'd hope (let's say one put an uncontroller variable in a meta which might end up with binary content).

Could we instead replace non-unicode strings with a placeholder (like we do for failing SQL quantization)? It would apply to resource, type and meta too.

AlexJF: This is what this PR does (check the tests). The error handling here is just to be extra sure but looking at the code for `ReadRune()` and `WriteRune()` no errors are returned except for`EOF` which is handled properly.

AlexJF: Added a comment to clarify `toUTF8`'s behaviour.

olivielpeau:
@gbbr Please don't rely on go 1.10 features yet, we're working on using go 1.10 for the 6.3 release but if we run into instabilities during QA we may have to revert to go 1.9

model/normalizer.go
LotharSee: Hmm, sneaky. And how could we be 100% sure that `in.ReadRune()` won't return a non `io.EOF` error? 
As if it happens, we are back to the situation where `Normalize` return an error -> trace dropped.

When reading it isn't much obvious. If we think we can have a "safe" toUTF8 then we should provide an implementation not returning error. And if errors can still happen, `Normalize` should handle them in a safe manner (non-dropping replacement).

What do you think?

model/normalizer.go
```diff
@@ -59,6 +61,11 @@ func (s *Span) Normalize() error {
 	if s.Resource == "" {
 		return errors.New("span.normalize: empty `Resource`")
 	}
+	utf8Resource, err := toUTF8(s.Resource)
+	if err != nil {
+		return fmt.Errorf("span.normalize: error converting resource to utf-8: %s", err.Error())
+	}
+	s.Resource = utf8Resource
```
LotharSee: We might want to test `s.Resource == ""` after the utf8 replacement? (I don't know if they actually are cases were it'd be false before the replacement / true after ; but I guess it is more logic to test for emptiness at the end anyway).

model/normalizer.go
```diff
@@ -59,6 +61,11 @@ func (s *Span) Normalize() error {
 	if s.Resource == "" {
 		return errors.New("span.normalize: empty `Resource`")
 	}
+	utf8Resource, err := toUTF8(s.Resource)
```
bmermet: With this implementation we are effectively allocating twice each string processed by the agent. With the assumption that these UTF-8 errors should be very rare, wouldn't it make sense to first detect if a string is invalid utf-8 before doing the substitution?

AlexJF: Yeah good point. I was worried that checking UTF-8 validation would itself be almost as expensive but just ran some checks and it seems to only bump 22.36% of time:

```
$ benchstat old.txt new-valid.txt
name             old time/op    new time/op    delta
Normalization-4     549ns ± 8%     672ns ± 1%  +22.36%  (p=0.008 n=5+5)

name             old alloc/op   new alloc/op   delta
Normalization-4      792B ± 0%      792B ± 0%     ~     (all equal)

name             old allocs/op  new allocs/op  delta
Normalization-4      9.00 ± 0%      9.00 ± 0%     ~     (all equal)
```

So I might just end up doing that :+1:

model/normalizer.go
```diff
@@ -259,18 +243,24 @@ func normMetricNameParse(name string) (string, bool) {
...
-func toUTF8(s string) (string, error) {
-	in := strings.NewReader(s)
-	out := strings.Builder{}
+func toUTF8(s string) string {
+	if utf8.ValidString(s) {
+		// if string is already valid utf8, return it as-is. Checking validity is cheaper than blindly rewriting.
+		return s
+	}
+
+	in := bytes.NewBufferString(s)
```
gbbr: We don't really need a buffer here, `strings.NewReader` should be good enough.

model/normalizer.go
```diff
@@ -259,18 +243,24 @@ func normMetricNameParse(name string) (string, bool) {
...
+	out := bytes.Buffer{}
```
gbbr: Would you mind changing this to:
```go
var out bytes.Buffer
```
I honestly find your version more uniform given the immediate declaration above, but I'd like to stay consistent.

gbbr:
All good. Please make sure to "Squash & Merge" when ready.

PR #432, opened 2018-06-05
writer: split stats writer payloads based on number of entries.

vlad-mh:
Adds a splitting logic in the stats writer in order never to exceed a pre-configured maximum number of "entries" in a stats payload.

Doing so ensures we don't hit the maximum payload size of the Datadog trace intake API.

The first commit is just a cleanup of the stats writer code, and some added comments.

The second commit adds the configuration value `max_entries_per_payload`. This value should be tweaked very cautiously. Setting it too low will have the agent send much more payloads than it should when flushing, which will impact performances and will hit the backend significantly more. Setting the value too high, and the payloads might end up being rejected due to their size.

The third commit is the actual splitting logic + unit tests.

### DIFF
--- a/writer/config/stats_writer.go
+++ b/writer/config/stats_writer.go
@@ -2,10 +2,12 @@ package config
 
 import "time"
 
-// maxEntriesPerPayload is the maximum number of entries in a stat payload. A
-// value of 0 disables splitting all together. For now, the feature is disabled
-// by default.
-const maxEntriesPerPayload = 0
+// maxEntriesPerPayload is the maximum number of entries in a stat payload. An
+// entry has an average size of 125 bytes in a compressed payload. The current
+// Datadog intake API limits a compressed payload to ~3MB (24,000 entries), but
+// let's have the default ensure we don't have paylods > 1.5 MB (12,000
+// entries).
+const maxEntriesPerPayload = 12000
 
 // StatsWriterConfig contains the configuration to customize the behaviour of a TraceWriter.
 type StatsWriterConfig struct {
--- a/writer/config/stats_writer.go
+++ b/writer/config/stats_writer.go
@@ -2,16 +2,23 @@ package config
 
 import "time"
 
+// maxEntriesPerPayload is the maximum number of entries in a stat payload. A
+// value of 0 disables splitting all together. For now, the feature is disabled
+// by default.
+const maxEntriesPerPayload = 0
+
 // StatsWriterConfig contains the configuration to customize the behaviour of a TraceWriter.
 type StatsWriterConfig struct {
-	UpdateInfoPeriod time.Duration
-	SenderConfig     QueuablePayloadSenderConf
+	MaxEntriesPerPayload int
+	UpdateInfoPeriod     time.Duration
+	SenderConfig         QueuablePayloadSenderConf
 }
 
 // DefaultStatsWriterConfig creates a new instance of a StatsWriterConfig using default values.
 func DefaultStatsWriterConfig() StatsWriterConfig {
 	return StatsWriterConfig{
-		UpdateInfoPeriod: 1 * time.Minute,
-		SenderConfig:     DefaultQueuablePayloadSenderConf(),
+		MaxEntriesPerPayload: maxEntriesPerPayload,
+		UpdateInfoPeriod:     1 * time.Minute,
+		SenderConfig:         DefaultQueuablePayloadSenderConf(),
 	}
 }
--- a/writer/stats_writer.go
+++ b/writer/stats_writer.go
@@ -108,6 +108,9 @@ func (w *StatsWriter) handleStats(stats []model.StatsBucket) {
 		nbEntries, nbStatBuckets, len(payloads),
 	)
 
+	if len(payloads) > 1 {
+		atomic.AddInt64(&w.info.Splits, 1)
+	}
 	atomic.AddInt64(&w.info.StatsBuckets, int64(nbStatBuckets))
 
 	headers := map[string]string{
@@ -285,12 +288,14 @@ func (w *StatsWriter) monitor() {
 			swInfo.StatsBuckets = atomic.SwapInt64(&w.info.StatsBuckets, 0)
 			swInfo.Bytes = atomic.SwapInt64(&w.info.Bytes, 0)
 			swInfo.Retries = atomic.SwapInt64(&w.info.Retries, 0)
+			swInfo.Splits = atomic.SwapInt64(&w.info.Splits, 0)
 			swInfo.Errors = atomic.SwapInt64(&w.info.Errors, 0)
 
 			w.statsClient.Count("datadog.trace_agent.stats_writer.payloads", int64(swInfo.Payloads), nil, 1)
 			w.statsClient.Count("datadog.trace_agent.stats_writer.stats_buckets", int64(swInfo.StatsBuckets), nil, 1)
 			w.statsClient.Count("datadog.trace_agent.stats_writer.bytes", int64(swInfo.Bytes), nil, 1)
 			w.statsClient.Count("datadog.trace_agent.stats_writer.retries", int64(swInfo.Retries), nil, 1)
+			w.statsClient.Count("datadog.trace_agent.stats_writer.splits", int64(swInfo.Splits), nil, 1)
 			w.statsClient.Count("datadog.trace_agent.stats_writer.errors", int64(swInfo.Errors), nil, 1)
 
 			info.UpdateStatsWriterInfo(swInfo)
--- a/writer/stats_writer.go
+++ b/writer/stats_writer.go
@@ -99,25 +99,16 @@ func (w *StatsWriter) Stop() {
 }
 
 func (w *StatsWriter) handleStats(stats []model.StatsBucket) {
-	numStats := len(stats)
-	if numStats == 0 {
+	payloads, nbStatBuckets, nbEntries := w.buildPayloads(stats, w.conf.MaxEntriesPerPayload)
+	if len(payloads) == 0 {
 		return
 	}
 
-	log.Debugf("going to flush stats buckets, %d buckets", numStats)
-	atomic.AddInt64(&w.info.StatsBuckets, int64(numStats))
+	log.Debugf("going to flush %v entries in %v stat buckets in %v payloads",
+		nbEntries, nbStatBuckets, len(payloads),
+	)
 
-	statsPayload := &model.StatsPayload{
-		HostName: w.hostName,
-		Env:      w.env,
-		Stats:    stats,
-	}
-
-	data, err := model.EncodeStatsPayload(statsPayload)
-	if err != nil {
-		log.Errorf("encoding issue: %v", err)
-		return
-	}
+	atomic.AddInt64(&w.info.StatsBuckets, int64(nbStatBuckets))
 
 	headers := map[string]string{
 		languageHeaderKey:  strings.Join(info.Languages(), "|"),
@@ -125,11 +116,128 @@ func (w *StatsWriter) handleStats(stats []model.StatsBucket) {
 		"Content-Encoding": "gzip",
 	}
 
-	atomic.AddInt64(&w.info.Bytes, int64(len(data)))
+	for _, p := range payloads {
+		// synchronously send the payloads one after the other
+		data, err := model.EncodeStatsPayload(p)
+		if err != nil {
+			log.Errorf("encoding issue: %v", err)
+			return
+		}
 
-	payload := NewPayload(data, headers)
+		payload := NewPayload(data, headers)
+		w.payloadSender.Send(payload)
 
-	w.payloadSender.Send(payload)
+		atomic.AddInt64(&w.info.Bytes, int64(len(data)))
+	}
+}
+
+type timeWindow struct {
+	start, duration int64
+}
+
+// buildPayloads returns a set of payload to send out, each paylods guaranteed
+// to have the number of stats buckets under the given maximum.
+func (w *StatsWriter) buildPayloads(stats []model.StatsBucket, maxEntriesPerPayloads int) ([]*model.StatsPayload, int, int) {
+	if len(stats) == 0 {
+		return []*model.StatsPayload{}, 0, 0
+	}
+
+	// 1. Get an estimate of how many payloads we need, based on the total
+	//    number of map entries (i.e.: sum of number of items in the stats
+	//    bucket's count map).
+	//    NOTE: we use the number of items in the count map as the
+	//    reference, but in reality, what take place are the
+	//    distributions. We are guaranteed the number of entries in the
+	//    count map is > than the number of entries in the distributions
+	//    maps, so the algorithm is correct, but indeed this means we could
+	//    do better.
+	nbEntries := 0
+	for _, s := range stats {
+		nbEntries += len(s.Counts)
+	}
+
+	if maxEntriesPerPayloads <= 0 || nbEntries < maxEntriesPerPayloads {
+		// nothing to do, break early
+		return []*model.StatsPayload{&model.StatsPayload{
+			HostName: w.hostName,
+			Env:      w.env,
+			Stats:    stats,
+		}}, len(stats), nbEntries
+	}
+
+	nbPayloads := nbEntries / maxEntriesPerPayloads
+	if nbEntries%maxEntriesPerPayloads != 0 {
+		nbPayloads++
+	}
+
+	// 2. Create a slice of nbPayloads maps, mapping a time window (stat +
+	//    duration) to a stat bucket. We will build the payloads from these
+	//    maps. This allows is to have one stat bucket per time window.
+	pMaps := make([]map[timeWindow]model.StatsBucket, nbPayloads)
+	for i := 0; i < nbPayloads; i++ {
+		pMaps[i] = make(map[timeWindow]model.StatsBucket, nbPayloads)
+	}
+
+	// 3. Iterate over all entries of each stats. Add the entry to one of
+	//    the payload container mappings, in a round robin fashion. In some
+	//    edge cases, we can end up having the same entry in several
+	//    inputted stat buckets. We must check that we never overwrite an
+	//    entry in the new stats buckets but cleanly merge instead.
+	i := 0
+	for _, b := range stats {
+		tw := timeWindow{b.Start, b.Duration}
+
+		for ekey, e := range b.Counts {
+			pm := pMaps[i%nbPayloads]
+			newsb, ok := pm[tw]
+			if !ok {
+				newsb = model.NewStatsBucket(tw.start, tw.duration)
+			}
+			pm[tw] = newsb
+
+			if _, ok := newsb.Counts[ekey]; ok {
+				newsb.Counts[ekey].Merge(e)
+			} else {
+				newsb.Counts[ekey] = e
+			}
+
+			if _, ok := b.Distributions[ekey]; ok {
+				if _, ok := newsb.Distributions[ekey]; ok {
+					newsb.Distributions[ekey].Merge(b.Distributions[ekey])
+				} else {
+					newsb.Distributions[ekey] = b.Distributions[ekey]
+				}
+			}
+			if _, ok := b.ErrDistributions[ekey]; ok {
+				if _, ok := newsb.ErrDistributions[ekey]; ok {
+					newsb.ErrDistributions[ekey].Merge(b.ErrDistributions[ekey])
+				} else {
+					newsb.ErrDistributions[ekey] = b.ErrDistributions[ekey]
+				}
+			}
+			i++
+		}
+	}
+
+	// 4. Create the nbPayloads payloads from the maps.
+	nbStats := 0
+	nbEntries = 0
+	payloads := make([]*model.StatsPayload, 0, nbPayloads)
+	for _, pm := range pMaps {
+		pstats := make([]model.StatsBucket, 0, len(pm))
+		for _, sb := range pm {
+			pstats = append(pstats, sb)
+			nbEntries += len(sb.Counts)
+		}
+		payloads = append(payloads, &model.StatsPayload{
+			HostName: w.hostName,
+			Env:      w.env,
+			Stats:    pstats,
+		})
+
+		nbStats += len(pstats)
+	}
+	return payloads, nbStats, nbEntries
 }
 
 // monitor runs the event loop of the writer's monitoring
--- a/writer/stats_writer.go
+++ b/writer/stats_writer.go
@@ -14,87 +14,72 @@ import (
 	writerconfig "github.com/DataDog/datadog-trace-agent/writer/config"
 )
 
-// StatsWriter ingests stats buckets and flushes their aggregation to the API.
+// StatsWriter ingests stats buckets and flushes them to the API.
 type StatsWriter struct {
-	stats    info.StatsWriterInfo
-	hostName string
-	env      string
-	conf     writerconfig.StatsWriterConfig
-	InStats  <-chan []model.StatsBucket
-
 	BaseWriter
+
+	// InStats is the stream of stat buckets to send out.
+	InStats <-chan []model.StatsBucket
+
+	// info contains various statistics about the writer, which are
+	// occasionally sent as metrics to Datadog.
+	info info.StatsWriterInfo
+
+	// hostName specifies the resolved host name on which the agent is
+	// running, to be sent as part of a stats payload.
+	hostName string
+
+	// env is environment this agent is configured with, to be sent as part
+	// of the stats payload.
+	env string
+
+	conf writerconfig.StatsWriterConfig
 }
 
-// NewStatsWriter returns a new writer for services.
+// NewStatsWriter returns a new writer for stats.
 func NewStatsWriter(conf *config.AgentConfig, InStats <-chan []model.StatsBucket) *StatsWriter {
 	writerConf := conf.StatsWriterConfig
 	log.Infof("Stats writer initializing with config: %+v", writerConf)
 
+	bw := *NewBaseWriter(conf, "/api/v0.2/stats", func(endpoint Endpoint) PayloadSender {
+		return NewCustomQueuablePayloadSender(endpoint, writerConf.SenderConfig)
+	})
 	return &StatsWriter{
-		hostName: conf.HostName,
-		env:      conf.DefaultEnv,
-		conf:     writerConf,
-		InStats:  InStats,
-		BaseWriter: *NewBaseWriter(conf, "/api/v0.2/stats", func(endpoint Endpoint) PayloadSender {
-			return NewCustomQueuablePayloadSender(endpoint, writerConf.SenderConfig)
-		}),
+		BaseWriter: bw,
+		InStats:    InStats,
+		hostName:   conf.HostName,
+		env:        conf.DefaultEnv,
+		conf:       writerConf,
 	}
 }
 
-// Start starts the writer.
+// Start starts the writer, awaiting stat buckets and flushing them.
 func (w *StatsWriter) Start() {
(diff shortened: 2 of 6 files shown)


gbbr:
@vlad-mh can I review or is it still WIP?

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+
+	if maxEntriesPerPayloads <= 0 || nbEntries < maxEntriesPerPayloads { // Nothing to do, break early.
+		return []*model.StatsPayload{&model.StatsPayload{
+			HostName: w.hostName,
+			Env:      w.env,
+			Stats:    stats,
+		}}, len(stats), nbEntries
+	}
+
+	nbPayloads := nbEntries / maxEntriesPerPayloads
```
LotharSee: As this is integer division, shouldn't it be `nbEntries / maxEntriesPerPayloads + 1`?

vlad-mh:
@gbbr ready for review

writer/stats_writer.go
LotharSee: Oops saw next line ; forget it.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+	//    from these maps. This allows is to have one stat bucket
+	//    per time window.
+	pMaps := make([]map[timeWindow]model.StatsBucket, nbPayloads)
+	for i := 0; i < nbPayloads; i++ {
+		pMaps[i] = make(map[timeWindow]model.StatsBucket, nbPayloads)
+	}
+
+	// 3. Iterate over each entries of each stats. Add the entry
+	//    to one of the payload container mappings, in a round
+	//    robin fashion. In some edge cases, we can end up having
```
LotharSee: In which case can we have that? I don't see how that could happen as the concentrator is already aggregating?

writer/config/stats_writer.go
```diff
@@ -4,14 +4,23 @@ import "time"
...
-	UpdateInfoPeriod time.Duration
-	SenderConfig     QueuablePayloadSenderConf
+	MaxEntriesPerPayload int
+	UpdateInfoPeriod     time.Duration
+	SenderConfig         QueuablePayloadSenderConf
 }
 
-// DefaultStatsWriterConfig creates a new instance of a StatsWriterConfig using default values.
+// DefaultStatsWriterConfig creates a new instance of a
+// StatsWriterConfig using default values.
```
gbbr: It should be fine to have this comment one one line... Do you have a set limit you want us to enforce? I was thinking 120 should be ok.

writer/config/stats_writer.go
```diff
@@ -4,14 +4,23 @@ import "time"
...
+	// An entry has an average size of 125 bytes in a compressed
+	// payload. The current Datadog intake API limits a compressed
+	// payload to ~3MB (24,000 entries), but let's have the
+	// default ensure we don't have paylods > 1.5 MB (12,000
+	// entries).
+
 	return StatsWriterConfig{
-		UpdateInfoPeriod: 1 * time.Minute,
-		SenderConfig:     DefaultQueuablePayloadSenderConf(),
+		MaxEntriesPerPayload: 12000,
```
gbbr: How would you feel about moving this default value into a constant, and the comment above as documentation for that constant? I think that would go great together and make everything clearer. As the comment stands now, it's a bit unclear at first what it's referring too because it's not referring to the direct next line.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
 )
 
-// StatsWriter ingests stats buckets and flushes their aggregation to the API.
+// StatsWriter ingests stats buckets and flushes them to the API.
 type StatsWriter struct {
-	stats    info.StatsWriterInfo
+	BaseWriter
+
+	// InStats is the stream of stat buckets to send out.
+	InStats <-chan []model.StatsBucket
```
gbbr: Since this is the _only_ input channel, let's call it the usual `in`. It will pair nicely with the same naming of the trace writer's "in" channel, like we have now ([link here](https://github.com/DataDog/datadog-trace-agent/blob/416acd39e81c5d6a1c8b79fe30abee295d4325b0/writer/trace_writer.go#L37)). It needs not be an exported field as it is not used anywhere. A lot of the identifiers in the agent are exported but aren't used in foreign packages.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+	// eye on things.
+	info info.StatsWriterInfo
+
+	// hostName is the resolved host name on which the agent is
+	// running, to be sent as part of a stats payload.
 	hostName string
-	env      string
-	conf     writerconfig.StatsWriterConfig
-	InStats  <-chan []model.StatsBucket
+	// env is environment this agent is configured with, to be
```
gbbr:
```go
// env specifies the environment that this agent is configured to. It will be
// sent as part of the stats payload.
```
Please add a blank line above this comment, to keep it consistent.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+	BaseWriter
+
+	// InStats is the stream of stat buckets to send out.
+	InStats <-chan []model.StatsBucket
+
+	// info holds some stats about the writer itself, to keep an
+	// eye on things.
+	info info.StatsWriterInfo
+
+	// hostName is the resolved host name on which the agent is
```
gbbr: s/is/specifies

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
 
 	return &StatsWriter{
-		hostName: conf.HostName,
-		env:      conf.DefaultEnv,
-		conf:     writerConf,
-		InStats:  InStats,
 		BaseWriter: *NewBaseWriter(conf, "/api/v0.2/stats", func(endpoint Endpoint) PayloadSender {
 			return NewCustomQueuablePayloadSender(endpoint, writerConf.SenderConfig)
 		}),
+
```
gbbr: I would remove all the blank lines here. No reason to group things, there's only a couple of them. I've noticed in Go it's generally preferred to have little blank lines most of the time, unless comments intervene or if it absolutely makes sense in large snippets.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
 func NewStatsWriter(conf *config.AgentConfig, InStats <-chan []model.StatsBucket) *StatsWriter {
 	writerConf := conf.StatsWriterConfig
 	log.Infof("Stats writer initializing with config: %+v", writerConf)
 
 	return &StatsWriter{
-		hostName: conf.HostName,
-		env:      conf.DefaultEnv,
-		conf:     writerConf,
-		InStats:  InStats,
 		BaseWriter: *NewBaseWriter(conf, "/api/v0.2/stats", func(endpoint Endpoint) PayloadSender {
```
gbbr: This isn't your code but it looks really funky. I wonder if there's any reason to have a copy here. Probably not. Do you think we could remove the `*`? It might be nicer to initialize the value before the struct declaration and then just assign it.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
 func (w *StatsWriter) Start() {
 	w.BaseWriter.Start()
+
+	// main goroutine
 	go func() {
 		defer watchdog.LogOnPanic()
 		w.Run()
 	}()
+
+	// monitoring goroutine
```
gbbr: I don't think any of these comments add value. The code is simple enough as-is. Can we remove them?

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+
+		hostName: conf.HostName,
+		env:      conf.DefaultEnv,
+
+		conf: writerConf,
 	}
 }
 
-// Start starts the writer.
+// Start starts the different goroutines of the writer, including the
```
gbbr: Maybe this would suffice:
```go
// Start starts the writer, awaiting stat buckets and flushing them.
```
If anyone is reading the code, they can see that two methods are started on this struct, and those methods have their own documentation. I'd suggest to keep it simple.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
-// StatsWriter ingests stats buckets and flushes their aggregation to the API.
+// StatsWriter ingests stats buckets and flushes them to the API.
 type StatsWriter struct {
-	stats    info.StatsWriterInfo
+	BaseWriter
+
+	// InStats is the stream of stat buckets to send out.
+	InStats <-chan []model.StatsBucket
+
+	// info holds some stats about the writer itself, to keep an
```
gbbr: I'm afraid that "keep an eye on things" isn't quite specific enough to make this field understandable. Maybe something along the lines of:
```go
// info contains various statistics about the writer, which are occasionally sent as metrics to Datadog.
```
Would it be better?

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
 	hostName string
-	env      string
-	conf     writerconfig.StatsWriterConfig
-	InStats  <-chan []model.StatsBucket
+	// env is environment this agent is configured with, to be
+	// sent as part of the stats payload.
+	env string
 
-	BaseWriter
+	conf writerconfig.StatsWriterConfig
```
gbbr: I don't feel strongly about this but it might make everything more "symmetrical" to add a comment here too, since you've started this for all the other fields.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+	// monitoring goroutine
+	go func() {
+		defer watchdog.LogOnPanic()
+		w.monitor()
+	}()
 }
 
-// Run runs the main loop of the writer goroutine. If flushes
-// stats buckets once received from the concentrator.
+// Run runs the event loop of the writer' main goroutine. It reads
```
gbbr: s/writer'/writer's

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+// stat buckets from InStats, builds stat payloads and sends them out
+// using the base writer's payloadSender.
```
gbbr: IMHO it should suffice to say:
```go
// using the base writer.
```

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
-			go w.updateInfo()
 		case <-w.exit:
 			log.Info("exiting stats writer")
 			return
 		}
 	}
 }
 
-// Stop stops the main Run loop.
+// Stop stops the writer's goroutines.
```
gbbr:
```go
// Stop stops the writer.
```

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
 func (w *StatsWriter) Stop() {
 	close(w.exit)
 	w.exitWG.Wait()
+
+	// Closing the base writer, among other things, will close the
```
gbbr: This comment seems like something that would belong inside the `BaseWriter`'s documentation, not here. Can we please remove it?

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
 
 	headers := map[string]string{
 		languageHeaderKey:  strings.Join(info.Languages(), "|"),
 		"Content-Type":     "application/json",
 		"Content-Encoding": "gzip",
 	}
 
-	atomic.AddInt64(&w.stats.Bytes, int64(len(data)))
+	for _, p := range payloads {
+		// Synchronously send the paylods one after the other
```
gbbr: s/paylods/payloads

Please either lowercase the first letter of the comment or add a period at the end, whichever you prefer.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+
+		atomic.AddInt64(&w.info.Bytes, int64(len(data)))
+	}
+}
+
+type timeWindow struct {
+	start, duration int64
+}
+
+// buildPayloads returns a set of payloads to send out, each paylods
```
gbbr: s/paylods/payload

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+// guaranteed to have the number of stats buckets under the given
+// maximum.
+func (w *StatsWriter) buildPayloads(stats []model.StatsBucket, maxEntriesPerPayloads int) ([]*model.StatsPayload, int, int) {
+	if len(stats) == 0 { // It could happen apparently
```
gbbr: Can we remove this comment? I don't think it adds any value.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+		return []*model.StatsPayload{}, 0, 0
+	}
+
+	// 1. Get an estimate of how many payloads we need, based on
+	//    the total number of map entries (i.e: sum of number of
```
gbbr: s/i.e/i.e.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+	//    items in the stats bucket's count map). Called N for the
```
gbbr: Why call it `N` for the rest of the description. It's already called `nbEntries`. It will make more sense to call it that in the other comments too. What do you think?

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+	//    the number of entries in the count map is > than the
+	//    number of entries in the distributions maps, so the
+	//    algorithm is correct, but indeed this means we could do
+	//    better.
+	nbEntries := 0
+	for _, s := range stats {
+		nbEntries += len(s.Counts)
+	}
+
+	if maxEntriesPerPayloads <= 0 || nbEntries < maxEntriesPerPayloads { // Nothing to do, break early.
```
gbbr: Can we please put this comment on the next line?

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+	// 2. Create a slice of N maps, mapping a time window (stat +
+	//    duration) to a stat bucket. We will build the payloads
+	//    from these maps. This allows is to have one stat bucket
+	//    per time window.
+	pMaps := make([]map[timeWindow]model.StatsBucket, nbPayloads)
+	for i := 0; i < nbPayloads; i++ {
+		pMaps[i] = make(map[timeWindow]model.StatsBucket, nbPayloads)
+	}
+
+	// 3. Iterate over each entries of each stats. Add the entry
```
gbbr: Iterate over all entries of each stat...

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+	//    to one of the payload container mappings, in a round
+	//    robin fashion. In some edge cases, we can end up having
+	//    a same entry in several inputted stat buckets. We must
```
gbbr: s/a/the

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+	//    check we never overwrite a entry in the new stats
```
gbbr: check **that** we never over write **an** entry.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+	//    buckets but cleanly merge instead.
+	i := 0
+	for _, sb := range stats {
```
gbbr: I really love short variable names, but in this case it might make the below code easier to reason about if we were to call it `bucket`, or `buck` if you will. Then we could also have `newbuck` or `newbucket`. Or even `b` if you like that more. It will pair nicer with the doc.

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+		tw := timeWindow{sb.Start, sb.Duration}
+
+		for ekey, e := range sb.Counts {
```
gbbr: Why not the typical `k, v` - or if you want `k, c` (for count)

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+	nbPayloads := nbEntries / maxEntriesPerPayloads
+	if nbEntries%maxEntriesPerPayloads != 0 {
+		nbPayloads++
+	}
+
+	// 2. Create a slice of N maps, mapping a time window (stat +
+	//    duration) to a stat bucket. We will build the payloads
+	//    from these maps. This allows is to have one stat bucket
+	//    per time window.
+	pMaps := make([]map[timeWindow]model.StatsBucket, nbPayloads)
```
gbbr: What's the reasoning behind `pMaps`? Wouldn't `buckets` be simpler and better?

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+
+	// 3. Iterate over each entries of each stats. Add the entry
+	//    to one of the payload container mappings, in a round
+	//    robin fashion. In some edge cases, we can end up having
+	//    a same entry in several inputted stat buckets. We must
+	//    check we never overwrite a entry in the new stats
+	//    buckets but cleanly merge instead.
+	i := 0
+	for _, sb := range stats {
+		tw := timeWindow{sb.Start, sb.Duration}
```
gbbr: s/tw/window

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+					newsb.ErrDistributions[ekey] = sb.ErrDistributions[ekey]
+				}
+			}
+			i++
+		}
+	}
 
-	payload := NewPayload(data, headers)
+	// 4. Create the N payloads from the maps.
+	nbStats := 0
```
gbbr: I'd opt for:
```go
var nbStats, nbEntries int
```

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
+	nbEntries = 0
+	payloads := make([]*model.StatsPayload, 0, nbPayloads)
+	for _, pm := range pMaps {
```
gbbr: How about:
```go
for _, b := range buckets {
```

writer/stats_writer_test.go
```diff
@@ -169,6 +170,210 @@ func TestStatsWriter_UpdateInfoHandling(t *testing.T) {
 	assert.Equal(expectedNumErrors, errorsSummary.Sum)
 }
 
+func TestStatsWriter_BuildPayloads(t *testing.T) {
+	assert := assert.New(t)
+
+	// Common case, no duplicate entries
+	{
```
gbbr: Instead of creating these artificial closures, let's remove the comments and use [sub-tests](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks).

writer/stats_writer_test.go
```diff
@@ -169,6 +170,210 @@ func TestStatsWriter_UpdateInfoHandling(t *testing.T) {
...
+		assert.Equal(expectedNbPayloads, len(payloads))
+		assert.Equal(expectedNbPayloads, nbStatBuckets)
+		assert.Equal(expectedNbEntries, nbEntries)
+
+		for i := 0; i < expectedNbPayloads; i++ {
+			assert.Equal(1, len(payloads[i].Stats))
+			assert.Equal(expectedNbEntriesByPayload[i], len(payloads[i].Stats[0].Counts))
+		}
+
+		// Assert the counts match
```
gbbr: Let's not leave commented out code.

writer/stats_writer_test.go
```diff
@@ -169,6 +170,210 @@ func TestStatsWriter_UpdateInfoHandling(t *testing.T) {
...
+		// 		for ekey, e := range s.Counts {
+		// 			assert.Equal(expectedCounts[ekey], e.Value)
+		// 		}
+		// 	}
+		// }
+		// Assert the counts match
+		actualCounts := make(map[string]float64, expectedNbEntries)
+		for _, p := range payloads {
+			for _, s := range p.Stats {
+				for ekey, e := range s.Counts {
```
gbbr: Please use `k, v`, `k, c` or `key, count`, etc.

writer/stats_writer_test.go
```diff
@@ -169,6 +170,210 @@ func TestStatsWriter_UpdateInfoHandling(t *testing.T) {
...
+		// time window (start: 0, duration 1e9).
+		stats := []model.StatsBucket{
+			fixtures.RandomStatsBucket(5),
+			fixtures.RandomStatsBucket(5),
+			fixtures.RandomStatsBucket(5),
+		}
+
+		// Remove duplicates so that we have a predictable
+		// state. In another case we'll test with duplicates.
+		expectedNbEntries := 0
```
gbbr: This is duplicated in two tests. IMHO we should try and remove duplication. This test is too long and complex. It should be short and easy to reason about. If possible, let's create some helper methods.

If we are using a fixture method and then iterating over it to change it, that means that we're not using the right method. Let's instead create helper methods for this. Maybe you can add something new to `fixtures` or just keep it in here.

writer/stats_writer.go
ufoot: I tend to prefer `nbPayloads := ((nbEntries - 1) / maxEntriesPerPayloads) + 1` which avoid a test, but your version might be more readable. Just a side note.

writer/config/stats_writer.go
vlad-mh: Yeah my editor is set to 80.. I'll change to 120

writer/config/stats_writer.go
vlad-mh: makes sense, will do

writer/stats_writer.go
vlad-mh: Had the same train of thought but didn't want to extend the scope of the PR to other files, as this is a publicly exported field and the PR is already big. If you insist, I can?

writer/stats_writer.go
vlad-mh: works for me

writer/stats_writer.go
vlad-mh: 👍

writer/stats_writer.go
vlad-mh: 👍

writer/stats_writer.go
vlad-mh: figured it was self explanatory

writer/stats_writer.go
vlad-mh: will do

writer/stats_writer.go
vlad-mh: as you wish

writer/stats_writer.go
vlad-mh: ok

writer/stats_writer.go
vlad-mh: 👍

writer/stats_writer.go
```diff
@@ -14,151 +14,294 @@ import (
...
 	w.BaseWriter.Stop()
 }
 
 func (w *StatsWriter) handleStats(stats []model.StatsBucket) {
-	numStats := len(stats)
-
-	// If no stats, we can't construct anything
-	if numStats == 0 {
+	payloads, nbStatBuckets, nbEntries := w.buildPayloads(stats, w.conf.MaxEntriesPerPayload)
+	if len(payloads) == 0 {
```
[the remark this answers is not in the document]
vlad-mh: we test for len of stats in buildPayloads right away, so same result.

writer/stats_writer.go
vlad-mh: ok

writer/stats_writer.go
vlad-mh: yup

writer/stats_writer.go
vlad-mh: pMaps => payloadMapping
each payload is mapped to a new bucket.
agreed something nicer could be found, but i don't like buckets, and payloads only make sense later in the func.

writer/stats_writer.go
vlad-mh: Indeed, it seemed like this couldn't happen with out current implementation. But it did show up right away in unit tests due to how the fixtures are done, and I don't think we should rely too much on the concentrator's logic. I figured a bit of pain now would go a long way in terms of reusability and potential sneaky bug.

writer/stats_writer_test.go
vlad-mh: was looking for them and didn't find in the doc, as I was looking for "case" and not sub-test :D perfect, thanks!

writer/stats_writer_test.go
vlad-mh: agreed

writer/stats_writer_test.go
vlad-mh: oups

writer/stats_writer.go
gbbr: I see. You are mapping buckets to time windows. What about `distribution`? Would that be any better?

writer/stats_writer.go
gbbr: Ok, I see how this works. `Monitor` is not the best method name for an exit channel 😄but that's another story. Let's keep it. Thanks for explaining.

writer/stats_writer.go
gbbr: Oh, it totally is. I just meant that all the other fields have comments and it will look better if this has one too. But it's a purely esthetical thing. You can leave it as it is, I absolutely don't mind at all.

writer/stats_writer.go
gbbr: Ah, my bad. I thought you added this field. You can leave it as is. If you were to change it, it would be super awesome and nice IMO but up to you. You don't have to!

writer/stats_writer.go
vlad-mh: would've loved to but nbEntries is not a redeclaration, just a value reset.

vlad-mh:
Hopefully this version of the PR fits a bit more.

To be clear on the goal: this PR doesn't aim to fix the overall architecture of how we create and flush stats bucket. There might be better ways to architecture it as a whole, and this requires more thought. This PR does, however, try to ensure we don't go above a certain payload size in worse case situations (very high volume services, or high resource cardinality), as the buckets end up being refused by the API, and as a result we have wrong generated metrics.

@ufoot and I chatted IRL about this. My understanding is that the collector relies on time window and having a full view of data during that time window, so we can't trigger a flush earlier than the ticker, based on number of entries. I think overall this logic still holds?

@gbbr I refactored the common code in the tests and used the sub test feature. I went through most of the nits as well.

vlad-mh:
Hopefully the last round of logic changes.
- the default value for the max number of entries is set to 0, thus disabling the feature. This will allow us to opt-in when payloads are being rejected by the API and see if it works well.
- the last commit adds the "splits" field to the writer's info, and report the number of split the agent is doing as a DD metric.

writer/stats_writer.go
```diff
@@ -14,151 +14,291 @@ import (
...
 
-	w.statsClient.Count("datadog.trace_agent.stats_writer.payloads", int64(swInfo.Payloads), nil, 1)
-	w.statsClient.Count("datadog.trace_agent.stats_writer.stats_buckets", int64(swInfo.StatsBuckets), nil, 1)
-	w.statsClient.Count("datadog.trace_agent.stats_writer.bytes", int64(swInfo.Bytes), nil, 1)
-	w.statsClient.Count("datadog.trace_agent.stats_writer.retries", int64(swInfo.Retries), nil, 1)
-	w.statsClient.Count("datadog.trace_agent.stats_writer.errors", int64(swInfo.Errors), nil, 1)
+	// 1. Get an estimate of how many payloads we need, based on the total
+	//    number of map entries (i.e.: sum of number of items in the stats
+	//    bucket's count map).
+	//    NOTE: we use the number of items in the count map as the
```
LotharSee: As curiosity, how is that possible to have a different count of Counts and Distribution?

vlad-mh: we will always have more counts than distributions: https://github.com/DataDog/datadog-trace-agent/blob/master/model/statsraw.go#L89

PR #433, opened 2018-06-06
config: make analyzed_spans configurable with an environment variable

bmermet:
This PR makes it possible to configure spans to be analyzed as transactions with a environment variable. The format for this environment variable is:

```
DD_APM_ANALYZED_SPANS="service_name|operation_name=1,other_service|other_operation=0.5"
```

### DIFF
--- a/config/agent_test.go
+++ b/config/agent_test.go
@@ -13,14 +13,14 @@ func TestDefaultConfig(t *testing.T) {
 	agentConfig := NewDefaultAgentConfig()
 
 	// assert that some sane defaults are set
-	assert.Equal(agentConfig.ReceiverHost, "localhost")
-	assert.Equal(agentConfig.ReceiverPort, 8126)
+	assert.Equal("localhost", agentConfig.ReceiverHost)
+	assert.Equal(8126, agentConfig.ReceiverPort)
 
-	assert.Equal(agentConfig.StatsdHost, "localhost")
-	assert.Equal(agentConfig.StatsdPort, 8125)
+	assert.Equal("localhost", agentConfig.StatsdHost)
+	assert.Equal(8125, agentConfig.StatsdPort)
 
-	assert.Equal(agentConfig.LogLevel, "INFO")
-	assert.Equal(agentConfig.Enabled, true)
+	assert.Equal("INFO", agentConfig.LogLevel)
+	assert.Equal(true, agentConfig.Enabled)
 
 }
 
@@ -155,9 +155,9 @@ func TestUndocumentedYamlConfig(t *testing.T) {
 	assert.Len(agentConfig.AnalyzedSpansByService, 2)
 	assert.Len(agentConfig.AnalyzedSpansByService["web"], 2)
 	assert.Len(agentConfig.AnalyzedSpansByService["db"], 1)
-	assert.Equal(agentConfig.AnalyzedSpansByService["web"]["request"], 0.8)
-	assert.Equal(agentConfig.AnalyzedSpansByService["web"]["django.request"], 0.9)
-	assert.Equal(agentConfig.AnalyzedSpansByService["db"]["intake"], 0.05)
+	assert.Equal(0.8, agentConfig.AnalyzedSpansByService["web"]["request"])
+	assert.Equal(0.9, agentConfig.AnalyzedSpansByService["web"]["django.request"])
+	assert.Equal(0.05, agentConfig.AnalyzedSpansByService["db"]["intake"])
 }
 
 func TestConfigNewIfExists(t *testing.T) {
@@ -197,15 +197,15 @@ func TestUndocumentedIni(t *testing.T) {
 	assert.NoError(err)
 
 	// analysis legacy
-	assert.Equal(c.AnalyzedRateByServiceLegacy["web"], 0.8)
-	assert.Equal(c.AnalyzedRateByServiceLegacy["intake"], 0.05)
+	assert.Equal(0.8, c.AnalyzedRateByServiceLegacy["web"])
+	assert.Equal(0.05, c.AnalyzedRateByServiceLegacy["intake"])
 	// analysis
 	assert.Len(c.AnalyzedSpansByService, 2)
 	assert.Len(c.AnalyzedSpansByService["web"], 2)
 	assert.Len(c.AnalyzedSpansByService["db"], 1)
-	assert.Equal(c.AnalyzedSpansByService["web"]["request"], 0.8)
-	assert.Equal(c.AnalyzedSpansByService["web"]["django.request"], 0.9)
-	assert.Equal(c.AnalyzedSpansByService["db"]["intake"], 0.05)
+	assert.Equal(0.8, c.AnalyzedSpansByService["web"]["request"])
+	assert.Equal(0.9, c.AnalyzedSpansByService["web"]["django.request"])
+	assert.Equal(0.05, c.AnalyzedSpansByService["db"]["intake"])
 }
 
 func TestAnalyzedSpansEnvConfig(t *testing.T) {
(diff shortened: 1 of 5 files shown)


config/merge_env.go
```diff
@@ -68,4 +69,51 @@ func mergeEnv(c *AgentConfig) {
...
+	rate, err := strconv.ParseFloat(splits[1], 64)
+	if err != nil {
+		return "", 0, fmt.Errorf("Unabled to parse rate")
+	}
+	return splits[0], rate, nil
+}
+
+func readAnalyzedSpanEnvVariable(analyzedSpansEnvVariable string) (map[string]map[string]float64, error) {
+	// the format is: service_name|operation_name=rate,other_service|other_operation=rate
+	analyzedSpansByService := make(map[string]map[string]float64)
```
gbbr: I'd suggest shortening this name to `analyzedSpans` (perhaps using the return variable from the above suggestion). I think that given the context of this function and the below code, it should simplify things and still make sense.

config/merge_env.go
```diff
@@ -68,4 +69,51 @@ func mergeEnv(c *AgentConfig) {
...
+		return "", 0, fmt.Errorf("Bad format")
+	}
+	rate, err := strconv.ParseFloat(splits[1], 64)
+	if err != nil {
+		return "", 0, fmt.Errorf("Unabled to parse rate")
+	}
+	return splits[0], rate, nil
+}
+
+func readAnalyzedSpanEnvVariable(analyzedSpansEnvVariable string) (map[string]map[string]float64, error) {
```
gbbr: This is not an environment variable but a string. So maybe this would be a better, shorter signature:
```go
func parseAnalyzedSpans(env string) (analyzedSpans map[string]map[string]float64, err error) {
```

config/merge_env.go
```diff
@@ -68,4 +69,51 @@ func mergeEnv(c *AgentConfig) {
...
+	if analyzedSpansEnvVariable == "" {
+		return analyzedSpansByService, nil
+	}
+	tokens := strings.Split(analyzedSpansEnvVariable, ",")
+	for _, token := range tokens {
+		name, rate, err := parseNameAndRate(token)
+		if err != nil {
+			return nil, err
+		}
+		serviceName, operationName, err := parseAnalyzedSpanFormat(name)
```
gbbr: `parseAnalyzedSpanFormat` does not say much about this function. Can we call it `parseServiceAndOp`. It will be a nice name and a good parallel to `parseNameAndRate`. We could have:
```go
service, op, err := parseServiceAndOp(name)
```

config/merge_env.go
```diff
@@ -68,4 +69,51 @@ func mergeEnv(c *AgentConfig) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		service := analyzedSpansByService[serviceName]
```
gbbr: This is confusing. Can we please simplify it and replace it with (the potentially more tedious but more explicit):
```go
if _, ok := analyzedSpans[service]; !ok {
    analyzedSpans[service] = make(map[string]float64)
}
analyzedSpans[service][op] = rate
```
Much clearer IMHO.

config/merge_env_test.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAnalyzedSpansEnvConfigParsing(t *testing.T) {
+	assert := assert.New(t)
+
+	// Check valid cases
```
gbbr: Instead of these comments, would you mind having two [sub-tests](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks) instead?

config/agent_test.go
```diff
@@ -197,13 +197,29 @@ func TestUndocumentedIni(t *testing.T) {
...
+	c, _ := NewAgentConfig(nil, nil, nil)
+
+	assert.Len(c.AnalyzedSpansByService, 2)
+	assert.Len(c.AnalyzedSpansByService["service1"], 2)
+	assert.Len(c.AnalyzedSpansByService["service2"], 1)
+	assert.Equal(0.5, c.AnalyzedSpansByService["service1"]["operation1"], 0.5)
+	assert.Equal(float64(0), c.AnalyzedSpansByService["service1"]["operation3"])
+	assert.Equal(float64(1), c.AnalyzedSpansByService["service2"]["operation2"])
+
+	os.Setenv("DD_APM_ANALYZED_SPANS", "")
```
gbbr: You may use [`os.Unsetenv`](https://golang.org/pkg/os/#Unsetenv), and it'd be nice pair it with the `os.Setenv` above in a `defer` statement.

config/agent_test.go
```diff
@@ -155,9 +155,9 @@ func TestUndocumentedYamlConfig(t *testing.T) {
 	assert.Len(agentConfig.AnalyzedSpansByService, 2)
 	assert.Len(agentConfig.AnalyzedSpansByService["web"], 2)
 	assert.Len(agentConfig.AnalyzedSpansByService["db"], 1)
-	assert.Equal(agentConfig.AnalyzedSpansByService["web"]["request"], 0.8)
-	assert.Equal(agentConfig.AnalyzedSpansByService["web"]["django.request"], 0.9)
-	assert.Equal(agentConfig.AnalyzedSpansByService["db"]["intake"], 0.05)
+	assert.Equal(0.8, agentConfig.AnalyzedSpansByService["web"]["request"])
```
gbbr: Thanks for these 👍

config/merge_env.go
```diff
@@ -68,4 +69,51 @@ func mergeEnv(c *AgentConfig) {
 	if v := os.Getenv("DD_LOG_LEVEL"); v != "" {
 		c.LogLevel = v
 	}
+
+	if v := os.Getenv("DD_APM_ANALYZED_SPANS"); v != "" {
+		var err error
```
gbbr: Why is this `var` declaration here? `if` can already have its own variable scope.

config/merge_env.go
```diff
@@ -68,4 +69,51 @@ func mergeEnv(c *AgentConfig) {
...
+		c.AnalyzedSpansByService, err = readAnalyzedSpanEnvVariable(v)
+		if err != nil {
+			log.Errorf("Bad format for DD_APM_ANALYZED_SPANS it should be of the form \"service_name|operation_name=rate,other_service|other_operation=rate\", error: %v", err)
+		}
+	}
+}
+
+func parseNameAndRate(token string) (string, float64, error) {
+	splits := strings.Split(token, "=")
```
gbbr: Most Go projects I've seen use `parts` here. As does the Go language itself, you can find [many occurences](https://github.com/golang/go/blob/8f6e8f9efd433fc67ff1e6e1caa5bc8edcad38f6/src/cmd/asm/internal/asm/endtoend_test.go#L74). Would you mind changing to that? I'd love to have consistency not only project-wide but community-wide.

config/merge_env.go
```diff
@@ -68,4 +69,51 @@ func mergeEnv(c *AgentConfig) {
...
+	}
+	rate, err := strconv.ParseFloat(splits[1], 64)
+	if err != nil {
+		return "", 0, fmt.Errorf("Unabled to parse rate")
+	}
+	return splits[0], rate, nil
+}
+
+func readAnalyzedSpanEnvVariable(analyzedSpansEnvVariable string) (map[string]map[string]float64, error) {
+	// the format is: service_name|operation_name=rate,other_service|other_operation=rate
```
gbbr: I would move this comment above the function and use it to document:
```go
// parseAnalyzedSpans parses the env string to extract a map of spans to be analyzed by service and operation.
// The format is: ...
func parseAnalyzedSpans(env string) (analyzedSpans map[string]map[string]float64, err error) {
```

config/agent_test.go
gbbr: Actually, thinking about it more - isn't there a risk that this removes the environment that might already be set by the user? I would tend to think that the other test should suffice and that we can remove this. If you want to keep it for coverage, which is fine, maybe it's best to see `os.Getenv` first and restore (or unset) to that value in a `defer`, after the test, as a teardown.

bmermet: I don't think there is such a risk, the environment variables are scoped to the current process: https://groups.google.com/forum/#!topic/golang-nuts/1jkFENw0gH8.

config/merge_env.go
bmermet: I did it that way because when I did:
```go
c.AnalyzedSpansByService, err := readAnalyzedSpanEnvVariable(v)
```
I got a compiler error:
```
merge_env.go:74:4: non-name c.AnalyzedSpansByService on left side of :=
```

PR #435, opened 2018-06-18
cmd/trace-agent: default to YAML config

gbbr:
* Small refactor of configuration code
* Adds packages `flags` and `osutil`
* Always default to Agent YAML config _unless_ it's not there and we have a legacy config available.


cmd/trace-agent/main_nix.go
```diff
@@ -4,28 +4,14 @@ package main
...
 	_ "net/http/pprof"
 
 	"github.com/DataDog/datadog-trace-agent/watchdog"
 )
 
-func init() {
-	// command-line arguments
-	// TODO: load from the .yaml automatically if there
-	flag.StringVar(&opts.configFile, "config", "/etc/dd-agent/datadog.conf", "Datadog Agent config file location")
-	flag.StringVar(&opts.legacyConfigFile, "ddconfig", "/etc/dd-agent/trace-agent.ini", "Deprecated extra configuration option.")
```
LotharSee: for context, the `ddconfig` config is from before the packaging  with the Agent 5 (when the trace-agent was its own .deb) ; so it should be fair to remove it.

However, we might want:
- Also remove the whole logic which goes with `opts.legacyConfigFile`.
- Do it in another PR as it is another concern

cmd/trace-agent/agent.go
```diff
@@ -113,7 +113,6 @@ func NewAgent(ctx context.Context, conf *config.AgentConfig) *Agent {
 		conf:               conf,
 		dynConf:            dynConf,
 		ctx:                ctx,
-		die:                die,
```
LotharSee: I might have missed something, but how could `a.die` https://github.com/DataDog/datadog-trace-agent/blob/d087519e0e04b33c5eabb2268f339d106adddca1/cmd/trace-agent/agent.go#L315 work without this?

config/agent.go
```diff
@@ -165,50 +169,65 @@ func getHostname(ddAgentBin string) (string, error) {
 
 }
 
-// NewAgentConfig creates the AgentConfig from the standard config
-func NewAgentConfig(conf *File, legacyConf *File, agentYaml *YamlAgentConfig) (*AgentConfig, error) {
-	c := NewDefaultAgentConfig()
-	var err error
+const legacyConfigFile = "/etc/dd-agent/datadog.conf"
```
LotharSee: - Let's call it agent5 something of legacy, as it'd help understand it is still legit/relevant.
- Windows need a different value. We could however avoid implementing this specific case by just checking how we package the Agent5 on Windows today and if it comes with the `-config` option.

config/agent.go
```diff
@@ -165,50 +169,65 @@ func getHostname(ddAgentBin string) (string, error) {
...
 
-	if conf != nil {
-		// Agent 5
-		err = mergeIniConfig(c, conf)
+// validate validates the given configuration, filling in any missing defaults
+// and merging any environment variables into it.
+func validate(c *AgentConfig) (*AgentConfig, error) {
```
LotharSee: Nit: I'd expect validate to only deal with errors/defaults and not the mergeEnv, as this is a real feature more than a simple validation.

gbbr:
Thanks! All your comments are valid and I will address them. Note that I’ve marked the PR as WIP. In case it’s spamming you, you might want to unsubscribe until it’s ready for review. Otherwise, all comments are always welcome!

cmd/trace-agent/agent.go
gbbr: I don’t know! This is weird!

config/agent.go
gbbr: Completely agree, this is confusing. It's just a temporary mid-refactor stage. Wasn't sure what to do with these yet.

gbbr:
@truthbk added you to check the following:

* `(config/agent.go).Load` to confirm that the new settings are backwards compatible. They should be.
* `(config/agent.go).(*AgentConfig).acquireHostname` for removing the Python shell code.

cmd/trace-agent/agent.go
gbbr: Resolved.

config/agent.go
```diff
@@ -6,23 +6,50 @@ import (
...
+
+var (
+	// ErrMissingAPIKey is returned when the config could not be validated due to missing API key.
+	ErrMissingAPIKey = errors.New("you must specify an API Key, either via a configuration file or the DD_API_KEY env var")
+
+	// ErrMissingHostname is returned when the config could not be validated due to missing hostname.
+	ErrMissingHostname = errors.New("failed to automatically set the hostname, you must specify it via configuration for or the DD_HOSTNAME env var")
+)
+
+const (
```
LotharSee: Should we move these to `merge_env.go` to make it more readable?

gbbr: Yes. Done.

config/agent.go
```diff
@@ -126,89 +138,89 @@ func NewDefaultAgentConfig() *AgentConfig {
...
+		if err := cfg.LoadIni(cfgPath); err != nil {
+			return cfg, err
+		}
+	case ".yaml":
+		if err := cfg.LoadYaml(cfgPath); err != nil {
+			return cfg, err
 		}
-		c.HostName = hostname
+	default:
+		return cfg, errors.New("unrecognised file extension (need .yaml)")
```
truthbk: I understand you're trying to deprecate `.ini`, `.conf` extensions, but this might be a little confusing since `ini` and `conf` extensions are actually supported.

gbbr: I agree. Done.

PR #438, opened 2018-06-19
quantizer: reduce spaces in quantized queries.

gbbr:
This is a breaking change. It removes spaces:
* Before any of the characters: ` ,=).`
* After any of the characters: `=(.`

This was causing a lot of unnecessary extra bytes previously.

### DIFF
--- a/cmd/trace-agent/agent_test.go
+++ b/cmd/trace-agent/agent_test.go
@@ -93,12 +93,12 @@ func TestFormatTrace(t *testing.T) {
 	assert.Equal(5000, len(result.Resource))
 	assert.NotEqual("Non-parsable SQL query", result.Resource)
 	assert.NotContains(result.Resource, "42")
-	assert.Contains(result.Resource, "SELECT name FROM people WHERE age = ?")
+	assert.Contains(result.Resource, "SELECT name FROM people WHERE age=?")
 
 	assert.Equal(5003, len(result.Meta["sql.query"])) // Ellipsis added in quantizer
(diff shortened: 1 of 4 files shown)


gbbr:
A discussion happened in the past suggesting that saving those extra bytes would serve as a valuable improvement. Based on what you're saying it sounds like it might not be so.

The other thought was to try and ensure that the outputted SQL is still valid to parse elsewhere in the pipeline. To clarify: `SELECT user.email, user.id, ...` currently quantizes into `SELECT user . email, user . id, ...` (notice the space) which would look both weird (and as a bug in the UI) and also render the SQL invalid.

If it still doesn't make sense to have it here and is not worth the breaking change, maybe we can do it in JavaScript in the app. It should be a really small change.

LotharSee:
Right it could be done in the UI.
Concerning the valid SQL output, that's something we might want to do in the future.
In that case, let's ideally "batch" all the breaking changes together.

gbbr:
@LotharSee @palazzem this query is a good example of too many spaces:
```sql
SELECT daily_values.*, LEAST ( ( ? - @runtot ), value ), ( @runtot : = @runtot + daily_values.value ) FROM ( SELECT @runtot : = ? ), daily_values WHERE daily_values . subject_id = ? AND daily_values . subject_type = ? AND ( daily_values.date BETWEEN ? AND ? ) HAVING value >= ? ORDER BY date
```
IMO we should consider this change. But possibly on the next major release when it'll be more obvious that there are breaking thing.

PR #442, opened 2018-06-22
Makefile: add 'binaries' command to help release pre-compiled binaries.

gbbr:
Allows running:
```bash
make binaries V=6.2.1
```
Which would result in checking out and compiling the binaries for revision `6.2.1` into the `./bin` folder. They are named `trace-agent-{$GOOS}-{$GOARCH}-6.2.1`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,16 @@ install:
 	go generate ./info
 	go install ./cmd/trace-agent
 
+binaries:
+	test -n "$(V)" # $$V must be set to the release version, e.g. "make binaries V=1.2.3"
+
+	# cross-compiling binaries using suffix $(V)
+	mkdir -p ./bin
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -20,6 +20,15 @@ install:
 	go generate ./info
 	go install ./cmd/trace-agent
 
+binaries:
+	test -n "$(V)" # $$V must be set to the release version, e.g. "make binaries V=1.2.3"
```
LotharSee: If that's expected to be the binary version, you should also set ` TRACE_AGENT_VERSION` as that's the version going to be set in the binary itself (returned by `-info`, tagging the hearthbeat metric, ...)

gbbr: Good catch! Thanks! I've added the `go generate` command before creating the binaries, based on `V`.

PR #443, opened 2018-06-25
config: add "connection_limit" option to yaml config

gbbr:
This adds the `connection_limit` configuration option to the yaml
configuration `apm_config` section to match the `ini` settings.


gbbr:
Sorry @LotharSee only read the second part of your comment now. Let's see what @palazzem has to say (since he requested it). Based on that answer I will make a follow-up PR.

gbbr:
One thing to note: "watchdog" kills the agent if the connection limit is exceeded, is this really what we want? The "old" ConnectionLimit simply errors. Seems less brutal.

palazzem:
@gbbr for now let's keep that change because we can deal with that later (in a minor release since we're deprecating a behavior). For the WatchDog question, we may revisit the approach.

PR #448, opened 2018-06-29
doc: add example yaml file

gbbr:
Adds an `example.yaml` configuration to serve as guidance for an overview of all agent configuration options.

### DIFF
--- a/datadog.example.yaml
+++ b/datadog.example.yaml
@@ -1,0 +1,91 @@
+# This file describes the trace agent specific configuration and its defaults.
+# It does not cover the entirety of configuration available to the Datadog
+# Agent 6. To get an overview of all the options, see:
+# https://github.com/DataDog/datadog-agent/blob/master/pkg/config/config_template.yaml
+#
+# The minimum necessary configuration to run the trace agent is uncommented.
+
+# Your Datadog API key. It can be found here:
+# https://app.datadoghq.com/account/settings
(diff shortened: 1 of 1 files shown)


example.yaml
```diff
@@ -0,0 +1,85 @@
+# This file describes the trace agent configuration and its defaults.
+# The minimum necessary configuration is uncommented.
+
+# Datadog API Key
+api_key: 1234
```
palazzem: can we change it as we do in Agent 6?
```
# The Datadog api key to associate your Agent's data with your organization.
# Can be found here:
# https://app.datadoghq.com/account/settings
api_key:
```

example.yaml
```diff
@@ -0,0 +1,85 @@
...
+ 
+# # Datadog Metrics Intake
+# dd_url: https://app.datadog.com
```
palazzem:
```
# The host of the Datadog intake server to send Agent data to
dd_url: https://app.datadoghq.com
```

example.yaml
```diff
@@ -0,0 +1,85 @@
+# This file describes the trace agent configuration and its defaults.
+# The minimum necessary configuration is uncommented.
```
palazzem: I would write something like: "for all available options please check the Agent 6 configuration" and maybe add a link to something we have for A6.

Makes sense?

PR #452, opened 2018-07-11
cmd/trace-agent: added zipkin v2 receiver

gbbr:
Adds support for Zipkin HTTP reporters at the endpoint: `http://localhost:8126/zipkin/v2/spans`

### Limitations

- Whenever possible, client applications should be configured to not share the same span ID between spans. In Go the setting is [`WithSharedSpans(false)`](https://godoc.org/github.com/openzipkin/zipkin-go#WithSharedSpans) and in Java it's [`supportsJoin(false)`](https://static.javadoc.io/io.zipkin.brave/brave/5.0.0/brave/Tracing.Builder.html#supportsJoin-boolean-). If this setting is not applied or not supported, the implementation will cover for it but it is error prone in scenarios where the span with the duplicate ID comes in in a separate payload. In that case one of the spans might be dropped somewhere else in the pipeline.

- Sampling priority is set to "_always keep_", meaning that the agent will do no sampling (but the backend will), leaving sampling decisions up to the client.


adriancole:
I think the "dropping somewhere else in the pipeline" approach is reasonable, but those testing it out can tell you how strongly or not they feel about it. Since B3 sampling status only makes sense when you have a single repo, ideally any instrumentation lacking a flag could be corrected before it becomes a problem..

gbbr:
@adriancole thanks a lot for your input. We are closing this for now (it was a hackathon project) as we don't yet know if we'll be able to commit to it 100%. In the meantime I will keep it in a private fork. We will definitely consider it however as I find it works quite well.

PR #457, opened 2018-07-18
obfuscate: add cache obfuscators

gbbr:
**Redis**
Adds an opt-in obfuscator for spans of type `redis`, targeting the `redis.raw_command` tag. As a rule of thumb, all the tokens marked `value` or `member` in the [Redis command specification](https://redis.io/commands) will be obfuscated.

**Memcached**
Adds an opt-in obfuscator for spans of type `memcached`, targeting the `memcached.command` tag. Based on the [protocol specification](https://github.com/memcached/memcached/blob/master/doc/protocol.txt) everything after the first new line is truncated.


tylerbenson:
What should client libraries do to take advantage of this?  Is there a way for client libraries to know if obfuscation is implemented in the agent before deciding to send unobfuscated data?  Perhaps some sort of agent version check?

obfuscate/obfuscate.go
```diff
@@ -38,6 +40,9 @@ func (o *Obfuscator) Obfuscate(span *model.Span) {
 		o.obfuscateSQL(span)
 	case "redis":
 		o.quantizeRedis(span)
```
palazzem: 👍 `quantizeRedis` must not be changed.

obfuscate/obfuscate_test.go
```diff
@@ -0,0 +1,166 @@
...
+		"redis.raw_command",
+		"SET key val",
+		"SET key val",
+		&config.ObfuscationConfig{},
+	))
+
+	t.Run("http/enabled", testConfig(
+		"http",
+		"http.url",
+		"http://foo.com/1/2?q=asd",
```
palazzem: nit: can you convert `http://foo.com/1/2?q=asd` into `http://mysite.mydomain/1/2?groupby=availability-zone`. It doesn't change anything, it's just that I would like to avoid having `foo`, `bar` or others in our tests/examples. I guess you can apply it globally in the PR.

obfuscate/obfuscate_test.go
```diff
@@ -0,0 +1,166 @@
...
+		"http://foo.com/1/2?q=asd",
+		&config.ObfuscationConfig{},
+	))
+
+	t.Run("json/enabled", testConfig(
+		"elasticsearch",
+		"elasticsearch.body",
+		`{"foo": "bar"}`,
```
palazzem: `{"foo": "bar"}` -> `{"role": "database"}`

obfuscate/redis.go
```diff
@@ -18,6 +18,9 @@ var redisCompoundCommandSet = map[string]bool{
 	"CLIENT": true, "CLUSTER": true, "COMMAND": true, "CONFIG": true, "DEBUG": true, "SCRIPT": true}
 
 // quantizeRedis generates resource for Redis spans
+// TODO(gbbr): Refactor this method to use the tokenizer and
+// remove "compactWhitespaces". This method is buggy when commands
```
palazzem: Can you define _buggy_? What is the behavior?

obfuscate/obfuscate_test.go
gbbr: Done.

obfuscate/obfuscate_test.go
gbbr: Done.

PR #468, opened 2018-09-05
cmd/trace-agent: enable stats calculation for non-top-level spans

p-lambert:
This PR makes it possible to generate APM stats from from non-top-level spans.
In order to enable stats computation for those, users will be required to flag their spans of interest with `datadog.trace_metrics: true` meta data.


LotharSee:
Here is what we discussed IRL @palazzem about allowing to force stats for a given span.
This PR is on top of some existing code that was meant to allow it.


We might want to spec with you exactly which tag / values to use for such a flag?

gbbr:
FYI we've started using [`golang/go` style](https://golang.org/doc/contribute.html#commit_messages) commit messages. I've changed the PR title to match that, can you please amend your commit message to be the same when merging? I was out and missed your [other PR](https://github.com/DataDog/datadog-trace-agent/pull/467).

Also please prefer "Squash & Merge" where possible.

p-lambert:
@gbbr: changed the commit message.
@LotharSee @palazzem: have we agreed on the key value pair?

LotharSee:
Not yet wanting on Emanuele to come back from PTO. 
Do you need it to be merged quickly for testing or are you fine to wait for him?

p-lambert:
> Not yet wanting on Emanuele to come back from PTO.
> Do you need it to be merged quickly for testing or are you fine to wait for him?

@LotharSee totally fine to wait for him to come back as I can test on staging with the current key/value

gbbr:
I guess this was already tested tons by now. Should it go into 6.7.0 next week?

stephenh:
I don't see any "yes please" comments yet, so I'll indulge myself with commenting that we'd really like to use this. Thanks!

PR #471, opened 2018-09-17
obfuscate: use strings.Builder from go1.10

gbbr:
This change updates JSON and Redis obfuscator to use `strings.Builder`
instead of `bytes.Buffer` for a minor speed increase. It also updates CI
to use go1.10, where this new structure was added.

Additionally, it also fixes a minor bug with an error message in the `config`
package.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ jobs:
           resource_class: large
 
           docker:
-               - image: circleci/golang:1.9.4
+               - image: circleci/golang:1.10
 
           steps:
                - checkout
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ it manually.
 
 ## Development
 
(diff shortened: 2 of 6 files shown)


gbbr:
```
RedisObfuscator/445b-4      8.42µs ± 0%    6.48µs ± 0%  -23.06%
RedisQuantizer-4            1.27µs ± 0%    1.23µs ± 0%   -2.92%

RedisObfuscator/445b-4      9.68kB ± 0%    3.74kB ± 0%  -61.32%
RedisQuantizer-4            1.48kB ± 0%    1.37kB ± 0%   -7.32%

RedisObfuscator/445b-4         133 ± 0%        94 ± 0%  -29.32%
RedisQuantizer-4              12.0 ± 0%      11.0 ± 0%   -8.33%
```

gbbr:
@rochdev 👏 excellent! Done.

PR #472, opened 2018-09-18
config: added tests for various hostname acquisition strategies

gbbr:
This change ensures that configuration can only be loaded if it is valid. It adds tests for various hostname acquisition strategies to ensure hostname is always set or otherwise trace agent should fail to start.

It also renames folders `test_cases` to `testdata`, which seems more common for Go.


cmd/trace-agent/main.go
```diff
@@ -91,17 +91,8 @@ func runAgent(ctx context.Context) {
 
 	cfg, err := config.Load(flags.ConfigPath)
 	if err != nil {
-		if !os.IsNotExist(err) {
```
gbbr: Config must be validated when loaded, not separately using another call. This way, the config is loaded, validated, and overridden from environment in one call.

PR #477, opened 2018-09-26
filters: apply replace rules only after obfuscation

### LINKED ISSUE #462: cmd/trace-agent: resource replace rules get applied to "sql.query" tag too
Moved from https://github.com/DataDog/dd-trace-go/issues/274.

I work on a service that has to use complicated and varying WHERE expressions in its queries. There is no way to avoid this because, among other reasons, we have to use IN expressions with a varying number of arguments, and we can not convert the IN expressions into JOINs for performance reasons. This makes the Datadog APM UI harder to use because it results in a very high cardinality of resource names. The docs also indicate there is a hard limit to the allowed cardinality of a resource, and we don't want to lose data.

There are some things you could do automatically to try and group queries. But I think it might be better if you just let users customize it as an option, since there isn't really a way to cover every use case for every database. So, can you provide a hook so we can set our own function that normalizes resource names in queries? (And maybe you could provide a few examples that people can use if it fits their use case.)

Eg. I'd like to be able to write a function that automatically turns a query like this:

```
SELECT col1, col2, col3
FROM tablea JOIN tableb on tablea.col52 = tableb.col3
WHERE (tablea.col5 in (?, ?, ?, ?) AND tableb.col4 in (?, ?)) AND tablea.col76 = 5 AND tableb.col33 in (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
```
into a query like `SELECT col1, col2, col3 FROM tablea JOIN tableb ON tablea.col52 = tableb.col3` using a regular expression. Then the original query could be added as a tag so it can still be viewed for debugging.

gbbr:
Fixes #462 

Problem:

* When using replace rules to modify the resource of an SQL span, the resulting `sql.query` tag created by the agent is also affected by the replace rules because it is a copy of the resulting resource. This change moves the replacer _after_ obfuscation, when the `sql.query` tag is already created, leaving the original (obfuscated) query visible to the user. This also makes more sense from a UX point of view because our users will want to provide regular expressions for replacements based on what they see in the UI, which is the post-obfuscation entry.

Changes to `filter` package:

* Removes `Filter` interface
* Exports `resourceFilter` as `Blacklister`
* Exports `tagReplacer` as `Replacer`

Changes to processor:

* Instead of running a list of filters (which weren't all filters to begin with), keep an instance of the `Replacer` and `Blacklister` and run them in the correct order.
* Run `Blacklister` early and `Replacer` only after obfuscation.


cmd/trace-agent/agent.go
```diff
@@ -50,7 +48,8 @@ func (pt *processedTrace) getSamplingPriority() (int, bool) {
 type Agent struct {
 	Receiver           *HTTPReceiver
 	Concentrator       *Concentrator
-	Filters            []filters.Filter
+	Blacklister        *filters.Blacklister
```
tylerbenson: Can I suggest Denylister instead?
https://github.com/rails/rails/issues/33677

cmd/trace-agent/agent_test.go
```diff
@@ -107,6 +109,71 @@ func TestFormatTrace(t *testing.T) {
...
+		// • obfuscator runs before replacer
+		// • obfuscator obfuscates both resource and "sql.query" tag
+		// • resulting resource is obfuscated with replacements applied
+		// • resulting "sql.query" tag is obfuscated with no replacements applied
+		cfg := config.New()
+		cfg.APIKey = "test"
+		cfg.ReplaceTags = []*config.ReplaceRule{{
+			Name: "resource.name",
+			Re:   regexp.MustCompile("AND.*"),
+			Repl: "...",
```
tylerbenson: This replacement is something that would normally break the parser for normalization, right?

gbbr: Possibly, although that wasn't the case being tested here. But you're right, this is yet another bug that is resolved by this change.

cmd/trace-agent/agent.go
gbbr: I'm more along the lines of the second comment in that thread.

tylerbenson: Still worth considering though.

PR #480, opened 2018-10-04
cmd/trace-agent: factor out receiver into "api" package

gbbr:
The goal is to separate all the stuff from `cmd/trace-agent` into smaller packages and ultimately end up with a clear separation between the three main (high level) functionalities of the agent:

* API (aka receiver)
* Processing pipeline (sampling, collecting stats, etc. - currently cmd/trace-agent)
* Writers

Clearer separation will eventually facilitate better testing. There is no change to functionality.


gbbr:
Thanks for helping.

PR #485, opened 2018-10-09
config: add 'site' option

gbbr:
Adds support for a new `site` top level configuration option, such that:
* When `site` is set, `https://trace.agent.<site>` is used.
* When both `site` and `apm_dd_url` are set, the latter takes precedence and user is notified.
* When only `apm_dd_url` is set, that value will be used.

### DIFF
--- a/config/agent_test.go
+++ b/config/agent_test.go
@@ -63,6 +63,24 @@ func TestConfigHostname(t *testing.T) {
 	})
 }
 
+func TestSite(t *testing.T) {
+	for name, tt := range map[string]struct {
+		file string
+		url  string
+	}{
+		"default":  {"./testdata/site_default.yaml", "https://trace.agent.datadoghq.com"},
+		"eu":       {"./testdata/site_eu.yaml", "https://trace.agent.datadoghq.eu"},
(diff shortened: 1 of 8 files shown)


config/merge_yaml.go
```diff
@@ -197,6 +201,15 @@ func (c *AgentConfig) loadYamlConfig(yc *YamlAgentConfig) {
 	if yc.StatsdPort > 0 {
 		c.StatsdPort = yc.StatsdPort
 	}
+	if yc.Site != "" {
+		c.APIEndpoint = apiEndpointPrefix + yc.Site
+	}
+	if yc.TraceAgent.Endpoint != "" {
+		c.APIEndpoint = yc.TraceAgent.Endpoint
```
palazzem: 👍

config/agent_test.go
```diff
@@ -63,6 +63,24 @@ func TestConfigHostname(t *testing.T) {
 	})
 }
 
+func TestSite(t *testing.T) {
```
palazzem: @olivielpeau can you double check if it respects entirely our specification?

gbbr:
@palazzem PTAL

PR #487, opened 2018-10-10
obfuscate: clean up SQL obfuscator

gbbr:
* Renaming and simplifying comments
* Removed `agent.parse.error` tag as it brought no real details about parsing errors.
* Reset obfuscator only once, before it runs.

No apparent impact on performance:
```
Tokenizer/Escaping/512-8    7.27µs ± 0%    7.30µs ± 0%  +0.32%
Tokenizer/Grouping/199-8    4.16µs ± 0%    4.17µs ± 0%  +0.19%

Tokenizer/Escaping/512-8    4.40kB ± 0%    4.40kB ± 0%   0.00%
Tokenizer/Grouping/199-8    2.72kB ± 0%    2.72kB ± 0%   0.00%

Tokenizer/Escaping/512-8      87.0 ± 0%      87.0 ± 0%   0.00%
Tokenizer/Grouping/199-8      68.0 ± 0%      68.0 ± 0%   0.00%
```

### DIFF
--- a/obfuscate/sql.go
+++ b/obfuscate/sql.go
@@ -8,12 +8,9 @@ import (
 	log "github.com/cihub/seelog"
 )
 
-const (
-	sqlQueryTag      = "sql.query"
-	sqlQuantizeError = "agent.parse.error"
-)
+const sqlQueryTag = "sql.query"
 
-// tokenFilter is a generic interface that a tokenConsumer expects. It defines
+// tokenFilter is a generic interface that a sqlObfuscator expects. It defines
 // the Filter() function used to filter or replace given tokens.
 // A filter can be stateful and keep an internal state to apply the filter later;
 // this can be useful to prevent backtracking in some cases.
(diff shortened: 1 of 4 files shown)


obfuscate/sql.go
```diff
@@ -151,32 +148,20 @@ type tokenConsumer struct {
 // Process the given SQL or No-SQL string so that the resulting one is properly altered. This
 // function is generic and the behavior changes according to chosen tokenFilter implementations.
 // The process calls all filters inside the []tokenFilter.
-func (t *tokenConsumer) Process(in string) (string, error) {
-	out := &bytes.Buffer{}
-	t.tokenizer.InStream.Reset(in)
-
+func (t *sqlObfuscator) obfuscate(in string) (string, error) {
+	var out bytes.Buffer
+	t.reset(in)
```
palazzem: Is it enough to reset it once? I guess it was redundant, but can you check the benchmarks (I think we should have something) to detect how many allocations / performance we have? Just for posterity, since this is a critical component that is called so many times.

gbbr: I have updated the PR description. There is a small difference in time, but I think that's negligible and most likely unrelated to the change. Happy to address performance improvements in another PR. How's it look?

gbbr:
@palazzem PTAL

PR #488, opened 2018-10-10
obfuscate: filter out contents of IN function

gbbr:
This change enhances the discarding filter to remove contents of `IN` functions.

Additionally, it moves, documents and renames token types that were meant to be used only by filters out of the `tokenizer.go` file to underline their special meaning.

### DIFF
--- a/obfuscate/sql_test.go
+++ b/obfuscate/sql_test.go
@@ -346,8 +346,24 @@ FROM [Blogs] AS [b
 ORDER BY [b].[Name]`,
 			`Non-parsable SQL query`,
 		},
+		{
+			"SELECT * FROM `host` WHERE `id` IN (\"jim@james.com\")",
+			"SELECT * FROM host WHERE id IN ( ? )",
+		},
+		{
+			`
+SELECT p.FirstName, p.LastName  
+FROM Person.Person AS p  
+    JOIN Sales.SalesPerson AS sp  
+    ON p.BusinessEntityID = sp.BusinessEntityID  
+WHERE p.BusinessEntityID IN  
+   (SELECT BusinessEntityID  
+   FROM Sales.SalesPerson IN (SELECT * FROM entity WHERE user = "charlie")
+   WHERE SalesQuota > 250000)
+`,
+			`SELECT p.FirstName, p.LastName FROM Person.Person JOIN Sales.SalesPerson ON p.BusinessEntityID = sp.BusinessEntityID WHERE p.BusinessEntityID IN ( ? )`,
+		},
 	}
-
 	for _, c := range cases {
 		s := SQLSpan(c.query)
 		NewObfuscator(nil).Obfuscate(s)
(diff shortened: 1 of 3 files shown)


obfuscate/sql_test.go
```diff
@@ -346,8 +346,24 @@ FROM [Blogs] AS [b
...
+			"SELECT * FROM host WHERE id IN ( ? )",
+		},
+		{
+			`
+SELECT p.FirstName, p.LastName  
+FROM Person.Person AS p  
+    JOIN Sales.SalesPerson AS sp  
+    ON p.BusinessEntityID = sp.BusinessEntityID  
+WHERE p.BusinessEntityID IN  
+   (SELECT BusinessEntityID
```
palazzem: I'm wondering if there is a way to not trigger the obfuscation if inside `IN ()` we have a `SELECT` token. How is the obfuscated query before this change? Can you post it here?

gbbr: > I'm wondering if there is a way to not trigger the obfuscation if inside IN () we have a `SELECT` token.

There is a solution to anything, but that would be super hacky and error prone. AFAIK you can have _any_ query inside the `IN` function, not just `SELECT`.

> How is the obfuscated query before this change? Can you post it here?

This is the result by running the `master` branch:
```
SELECT p.FirstName, p.LastName FROM Person.Person JOIN Sales.SalesPerson ON p.BusinessEntityID = sp.BusinessEntityID WHERE p.BusinessEntityID IN ( SELECT BusinessEntityID FROM Sales.SalesPerson IN ( SELECT * FROM entity WHERE user = charlie ) WHERE SalesQuota > ? )
```

As you can see, this clearly reflects that the current solution does not solve the main problem. The word `charlie` is not obfuscated and could be sensitive data.

---

Once more, I'd like to clarify that the issue that we are solving here is not that we aren't obfuscating the `IN` function, but that we aren't obfuscating double-quoted strings. Even with this change, sensitive data will still leak when using double quotes, which is a common scenario. See:

* [MySQL String Literals](https://dev.mysql.com/doc/refman/8.0/en/string-literals.html)
* [MS-SQL quoting](https://social.msdn.microsoft.com/Forums/sqlserver/en-US/b2443854-5f02-4ebe-ad8d-61725a95b2bf/forum-faq-how-do-i-use-double-quotes-and-single-quotes-when-building-a-query-string?forum=sqldataaccess)

As initial suggested, my proposal is to obfuscate double-quoted strings too, but then again you can also have queries that look like:

```
SELECT "name" FROM users
```
resulting in:
```
SELECT ? FROM users
```
or
```
SELECT name FROM users WHERE "email"="a@b.com"
```
resulting in:
```
SELECT name FROM users WHERE ?=?
```

gbbr: > There is a solution to anything, but that would be super hacky and error prone

I also want to clarify about my statement above: it applies if we implement it within the current solution with filters. If we were to interpret the SQL query (which we are already starting to), then it'd most likely be a lot cleaner and more accurate, but would require a larger amount of work (some refactoring) and a lot of thorough testing.

PR #490, opened 2018-10-10
cmd/trace-agent: improve Agent.Process readability and add Span.SetMetric

LotharSee:
- Reorder the content of `Agent.Process` by more meaningful units and add extra comments.
- Add a `Span.SetMetric` function and use it in relevant places.

This prepare future changes involving that function and the usage of many more metrics.

It should have been two PRs (they are 2 commits) but as they are built on top of each other, I went for the easy way.

### DIFF
--- a/cmd/trace-agent/agent.go
+++ b/cmd/trace-agent/agent.go
@@ -183,27 +183,19 @@ func (a *Agent) Process(t model.Trace) {
 		return
 	}
 
-	// All traces should go through either through the normal score sampler or
-	// the one dedicated to errors
-	samplers := make([]*Sampler, 0, 2)
-	if traceContainsError(t) {
-		samplers = append(samplers, a.ErrorsScoreSampler)
-	} else {
-		samplers = append(samplers, a.ScoreSampler)
-	}
-
-	// We get the address of the struct holding the stats associated to no tags
-	// TODO: get the real tagStats related to this trace payload.
-	ts := a.Receiver.Stats.GetTagStats(info.Tags{})
-	stat := &ts.TracesPriorityNone
+	// Root span is used to carry some trace-level metadata, such as sampling rate and priority.
 	root := t.GetRoot()
-	priority, hasPriority := root.Metrics[sampler.SamplingPriorityKey]
-	if hasPriority {
-		// If Priority is defined, send to priority sampling, regardless of priority value.
-		// The sampler will keep or discard the trace, but we send everything so that it
-		// gets the big picture and can set the sampling rates accordingly.
-		samplers = append(samplers, a.PrioritySampler)
 
+	// We get the address of the struct holding the stats associated to no tags.
+	// TODO: get the real tagStats related to this trace payload (per lang/version).
+	ts := a.Receiver.Stats.GetTagStats(info.Tags{})
+
+	// Extract priority early, as later goroutines might manipulate the Metrics map in parallel which isn't safe.
+	priority, hasPriority := root.Metrics[sampler.SamplingPriorityKey]
+
+	// Depending on the sampling priority, count that trace differently.
+	stat := &ts.TracesPriorityNone
+	if hasPriority {
 		if priority < 0 {
 			stat = &ts.TracesPriorityNeg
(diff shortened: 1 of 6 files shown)


cmd/trace-agent/agent.go
```diff
@@ -274,37 +270,49 @@ func (a *Agent) Process(t model.Trace) {
...
-		defer watchdog.LogOnPanic()
-
-		// Trace sampling
-		sampled := false
-		for _, s := range samplers {
-			// Consider trace as sampled if at least one of the samplers kept it
-			sampled = s.Add(pt) || sampled
-		}
+	// Don't go through sampling for < 0 priority traces
+	if !hasPriority || priority < 0 {
```
gbbr: I'm confused here. If the trace doesn't have a user set sampling priority, wouldn't that mean it goes through the normal process of sampling? Shouldn't this simply be:
```go
if priority < 0 {
    // negative priority means reject
    return
}
```

cmd/trace-agent/agent.go
```diff
@@ -183,27 +183,19 @@ func (a *Agent) Process(t model.Trace) {
...
-	} else {
-		samplers = append(samplers, a.ScoreSampler)
-	}
+	// Root span is used to carry some trace-level metadata, such as sampling rate and priority.
+	root := t.GetRoot()
 
-	// We get the address of the struct holding the stats associated to no tags
-	// TODO: get the real tagStats related to this trace payload.
+	// Extract priority early, as later goroutines might manipulate the Metrics map in parallel which isn't safe.
+	priority, hasPriority := root.Metrics[sampler.SamplingPriorityKey]
```
gbbr: I think it'd be nicer to move this above line 198 (`if hasPriority {`). It's not used until then. Should be fine I think.

cmd/trace-agent/agent.go
```diff
@@ -274,37 +270,49 @@ func (a *Agent) Process(t model.Trace) {
 		defer watchdog.LogOnPanic()
 		// Everything is sent to concentrator for stats, regardless of sampling.
 		a.Concentrator.Add(pt)
```
gbbr: I'm wondering if this goroutine could be merged with the other one (service extractor). I have the feeling (but not the benchmarks to prove it) that it will be a fairly good perf. increase. If you're also unsure, just leave it, I'll write some benchmarks and try it out later.

gbbr:
Rebased for you and re-worded your commits a bit.

cmd/trace-agent/agent.go
LotharSee: I actually had the same thoughts working on that function. The service extractor is barely doing anything, so I'd guess that a goroutine is bit overdoing it. Although, I don't have any benchmark to power such argument.

Let's maybe investigate and do a PR independently?

cmd/trace-agent/agent.go
LotharSee: Right, will adjust!

cmd/trace-agent/agent.go
```diff
@@ -231,12 +223,22 @@ func (a *Agent) Process(t model.Trace) {
...
+		span.Truncate()
+	}
+	a.Replacer.Replace(&t)
 
-	// Need to do this computation before entering the concentrator
-	// as they access the Metrics map, which is not thread safe.
+	// Extract the client sampling rate.
+	clientSampleRate := sampler.GetTraceAppliedSampleRate(root)
+	// Combine it with the pre-sampling rate.
+	preSamplerRate := a.Receiver.PreSampler.Rate()
```
gbbr: Is this rate ever going to be different than 1? We’ve stopped making it configurable. I’d like to put it into a constant in a subsequent PR and simply use that. Would that be ok?

LotharSee: So, we stopped making its initial value being configurable. But this value gets below 1 whenever the Agent uses too many resources -> it will drop a proportion of incoming payloads.

Note that I did split it into two variables (while we still only use its product, previously `rate`) to make it more readable / because in a next PR I'll do something out of each individual value.

PR #494, opened 2018-10-15
cmd/trace-agent: combine sampler rates

furmmon:
Traces are sampled by various independent samplers. If one decides to sample it, the trace is sampled. The usual samplers are: a scoreSampler and a prioritySampler.

In sampled traces we add the applied `sampling_rate`. This value is currently wrong as it does not take in account the sampling rate applied by the prioritySampler.

This PR aims at fixing the sample_rate values

1. Fetching Priority sampler sample rate
Retrieving client applied rate with the key `_sampling_priority_rate_v1`. If it does not exist, applying the next rate that will be sent to clients.

2. Merging signature|error sampling rates with priority.
Signature/Error samplers are run in parallel to the Priority sampler. Both are independent.
Sampling rate is, signatureSamplingRate + prioritySamplingRate - signatureSamplingRate*prioritySamplingRate 

3. Add sampling rate to the pre-sampler and agent sampler rates that are already set in the trace

4. Isolate sampling decision in a function and have specific tests for it


furmmon:
Not sure about the tests coverage required. Should I simulate with fixed traces rate with different priority the complete logic?
ex: 
- isolate the sampling decisions in a function
- send precise traces with priorities at specific rate
- assert sampling rate set on those traces

cmd/trace-agent/agent.go
```diff
@@ -272,40 +272,40 @@ func (a *Agent) Process(t model.Trace) {
...
+	sampled = sampled || scoreSampled
+	sampleRate = sampler.MergeParallelSamplingRates(scoreSampleRate, sampleRate)
+	sampler.AddSampleRate(pt.Root, sampleRate)
+
+	if sampled {
+		sampledTrace.Trace = &pt.Trace
+	}
+
+	sampledTrace.Transactions = a.TransactionSampler.Extract(pt)
+	// TODO: attach to these transactions the client, pre-sampler and transaction sample rates.
```
furmmon: Not sure of what was the goal of Transactions.
Should I attach specific sample rates there somehow?

sampler/prioritysampler.go
```diff
@@ -24,7 +24,9 @@ import (
 const (
 	// SamplingPriorityKey is the key of the sampling priority value in the metrics map of the root span
 	SamplingPriorityKey = "_sampling_priority_v1"
-	syncPeriod          = 3 * time.Second
+	// SamplingPriorityRateKey is the key to get the sample rate applied to a trace
+	SamplingPriorityRateKey = "_sampling_priority_rate_v1"
```
furmmon: Need to agree on the key format and it should be added to tracers payloads

sampler/scoresampler_test.go
```diff
@@ -98,41 +98,6 @@ func TestMaxTPS(t *testing.T) {
 		0.01+defaultDecayFactor-1)
 }
 
-func TestSamplerChainedSampling(t *testing.T) {
```
gbbr: Why are these tests removed?

sampler/signature.go
```diff
@@ -20,10 +20,10 @@ func (p spanHashSlice) Less(i, j int) bool { return p[i] < p[j] }
 func (p spanHashSlice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
 func sortHashes(hashes []spanHash)         { sort.Sort(spanHashSlice(hashes)) }
 
-// computeSignatureWithRootAndEnv generates the signature of a trace knowing its root
+// ComputeSignatureWithRootAndEnv generates the signature of a trace knowing its root
 // Signature based on the hash of (env, service, name, resource, is_error) for the root, plus the set of
 // (env, service, name, is_error) of each span.
-func computeSignatureWithRootAndEnv(trace model.Trace, root *model.Span, env string) Signature {
+func ComputeSignatureWithRootAndEnv(trace model.Trace, root *model.Span, env string) Signature {
```
gbbr: Please revert the changes in this file, we won't need to use these. We shouldn't export something (technically making it public API) for the sake of having it usable in tests.

cmd/trace-agent/agent_test.go
```diff
@@ -210,6 +210,97 @@ func TestProcess(t *testing.T) {
...
+			serviceSignature := sampler.ComputeServiceSignature(span, "")
+			prioritySampler := agent.PrioritySampler.engine.(*sampler.PriorityEngine)
+			priorityRate := prioritySampler.Sampler.GetSignatureSampleRate(serviceSignature)
+			expectedRate := sampler.MergeParallelSamplingRates(priorityRate, scoreRate)
+			assert.Equal(t, expectedRate, sampleRate, "score sampler rate in parallel to next priority rate")
+		}
+	}
+}
+
+func TestSamplingRace(t *testing.T) {
```
gbbr: Please remove this test, we don't need it. Tests already run in parallel and any races will be picked up by the race detector (if it is enabled).

cmd/trace-agent/agent_test.go
```diff
@@ -210,6 +210,97 @@ func TestProcess(t *testing.T) {
 	})
 }
 
+func TestSampling(t *testing.T) {
```
gbbr: I realise most of the trace agent has tests like this, and that has become technical debt. Let's start and promote better habits. With the new form of the `sample` method, it is now easy to create table-driven tests.

We now have the method:
```go
func (a *Agent) sample(pt processedTrace) (rate float64, ok bool) {
```
Its input is a `processedTrace`, for which we should expect a certain `rate` and `ok` value in return.

The tricky part here is that we can not control the output of each individual sampler, in the current context. It is also not our responsibility to test each called sampler here, as those should be tested on their own, separately. For this reason, we can mock out the samplers. Luckily, a sampler looks like this:

```go
type Sampler struct {
         // ... some fields
	engine sampler.Engine
}
```
Where `sampler.Engine` is an interface that we can mock.

So, we can create a `mockSampler` which always returns what we want. The goal would be to create an `Agent` structure like this:
```go
a := &Agent{
    ScoreSampler: newMockSampler(0.3, true),
    ErrorsSampler: newMockSampler(0.5, true),
    PrioritySampler: newMockSampler(0, false),
}
```
`newMockSampler` would always return a fake sampler implementation which always returns the `rate` and `ok` values we give it. Then we can test that `a.sample` behaves as it should, based on different expectations.

```go
func newMockSampler(wantRate float64, wantOk bool) *Sampler {
    return &Sampler{engine: &sampler.MockEngine{wantRate, wantOk}}
}
```
`MockEngine` should be fairly easy to do (it should be in the `sampler` package):
```go
var _ Engine = (*mockEngine)(nil)

type MockEngine struct {
    wantRate float64
    wantOk bool
}

// ... implement all sampler.Engine methods (they don't have to do anything)

func (e *MockEngine) Add(_ processedTrace) (float64, bool) {
    return e.wantRate, e.wantOk
}
```

And now, finally, we can write a simple to read test, like this:
```go
for name, tt := range []struct{
    hasErrors, hasPriority bool
    rateScore, ratePriority float64
    okScore, okPriority bool
    wantRate float64
    wantOk bool
}{
    "rate-only": {
        hasErrors: true,
        rateScore: 0.4,
        okScore: true,
        wantRate: 0.4,
        expectOk: true,
    },
    // continue with other tests, to cover all cases and branch in the method...
} {
    t.Run(tt.name, func(t *testing.T) {
        a := &Agent{
            ScoreSampler: newMockSampler(tt.rateScore, tt.okScore),
            ErrorsSampler: newMockSampler(tt.rateScore, tt.okScore),
            PrioritySampler: newMockSampler(tt.ratePriority, tt.okPriority),
        }
        var pt processedTrace
        if tt.hasErrors {
            // make sure 'pt' has an error
        }
        if tt.hasPriority {
            pt.Root.Metrics[sampler.SamplingPriorityKey] = 1
        }
        gotRate, gotOk := a.sample(pt)
        assert.EqualValues(t, tt.wantRate, gotRate)
    })
}
```

Even though it wasn't necessarily fast to reach this form, we can now very easily see all the test cases in a fairly short test that is much easier to understand and follow.

What do you think?

cmd/trace-agent/agent.go
```diff
@@ -272,40 +272,40 @@ func (a *Agent) Process(t model.Trace) {
...
 
-		sampled := false
-		for _, s := range samplers {
-			// Consider trace as sampled if at least one of the samplers kept it.
-			sampled = s.Add(pt) || sampled
-		}
-		if sampled {
-			sampledTrace.Trace = &pt.Trace
-		}
+	if hasPriority {
```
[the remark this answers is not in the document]
furmmon: this is unreadable
I don't understand  as well _, ok := pt.Root.Metrics[sampler.SamplingPriorityKey]; ok` than hasPriority
A compromised could be a small `func hasPriority(trace) bool` that can also be used in the process call above

cmd/trace-agent/agent_test.go
furmmon: Sorry I might have missed it.
What I'm testing here is parallel calls of a same trace signature on a unique sampler. This would trigger a race condition with the race test.

cmd/trace-agent/agent.go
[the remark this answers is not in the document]
gbbr: `ok` is the return value of `a.sample` and is a combination of `ok1` and `ok2`. If you don't like that use `sampled1` and `sampled2` or `okScore` and `okPriority`.

As for checking if map values exist, `ok` is the standard variable name to use. Let's not use `hasPriority`. No need for a func either.

cmd/trace-agent/agent_test.go
furmmon: I think this is nice, but I'll need to upgrade the test coverage of both samplers.
The test below is a complete integration test of the samplers, all cases pass but I guess it's hard to read.

cmd/trace-agent/agent.go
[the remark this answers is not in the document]
gbbr: Or, if you want, we can add another method in `model/span.go`, like this:
```go
func (s *Span) HasPriority() bool {
    if s.Metrics == nil {
        return false
    }
    _, ok := s.Metrics[sampler.SamplingPriorityKey]
    return ok
}
```
Then we can just use `pt.Root.HasPriority`. If you prefer this.

cmd/trace-agent/sampler.go
```diff
@@ -63,15 +63,16 @@ func (s *Sampler) Run() {
 }
 
 // Add samples a trace and returns true if trace was sampled (should be kept), false otherwise
-func (s *Sampler) Add(t processedTrace) bool {
+func (s *Sampler) Add(t processedTrace) (sampled bool, sampleRate float64) {
```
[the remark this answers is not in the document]
furmmon: I prefer sampled over ok, I think it's more explicit.
If ok is false, what does it mean? The `rate` value is still right

cmd/trace-agent/agent_test.go
gbbr: I understand, but here we should write unit tests, not integration tests. Integration tests are a completely different story. And yes, the samplers have to be tested separately.

cmd/trace-agent/sampler.go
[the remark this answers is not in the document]
gbbr: Use `ok`, it is the standard name for a second argument signifying if an operation finished OK or not, in Go.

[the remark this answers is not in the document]
gbbr: Ok, use `sampled` then.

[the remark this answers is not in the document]
gbbr: Ok, let's use `sample` then, it's fine.

sampler/coresampler_test.go
```diff
@@ -39,3 +40,33 @@ func TestSamplerLoop(t *testing.T) {
...
+	assert.Equal(t, 0.44, MergeParallelSamplingRates(rate1, rate2))
+	assert.Equal(t, 0.44, MergeParallelSamplingRates(rate2, rate1))
+
+	rate1 = 0.0
+	rate2 = 0.5
+	assert.Equal(t, 0.5, MergeParallelSamplingRates(rate1, rate2))
+	assert.Equal(t, 0.5, MergeParallelSamplingRates(rate2, rate1))
+}
+
+func TestAddSampleRate(t *testing.T) {
```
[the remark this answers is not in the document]
gbbr: Actually I think this one is fine. No need for a table here. Sorry. My bad.

sampler/coresampler.go
```diff
@@ -176,3 +176,20 @@ func GetTraceAppliedSampleRate(root *model.Span) float64 {
...
 	root.SetMetric(model.SpanSampleRateMetricKey, sampleRate)
 }
+
+// MergeParallelSamplingRates merges two rates from Sampler1, Sampler2. Both samplers law are independant,
+// and {sampled} = {sampled by Sampler1} or {sampled by Sampler2}
+func MergeParallelSamplingRates(rate1 float64, rate2 float64) float64 {
+	if rate1 >= 1 || rate2 >= 1 {
+		return 1
+	}
+	return rate1 + rate2 - rate1*rate2
```
gbbr: One last thing I forgot to ask you is if you could explain this formula. I'm not very strong at maths, maybe this is something simple. If it's not, maybe we should add a comment saying where this formula comes from?

gbbr: This is beyond me, but luckily we have people like you to think about this 😅I'll leave it up to you to decide if we need to put this in a comment. Maybe it will be helpful to people looking at the code in the future?

sampler/scoresampler_test.go
furmmon: Score sampler does not add anymore the applied sampling rate to a trace. Behavior tested here is removed

cmd/trace-agent/agent.go
```diff
@@ -272,40 +272,35 @@ func (a *Agent) Process(t model.Trace) {
...
+		a.sampledTraceChan <- &writer.SampledTrace{
+			Trace:        &pt.Trace,
+			Transactions: a.TransactionSampler.Extract(pt),
 		}
+	}()
+}
 
-		// Trace sampling.
-		var sampledTrace writer.SampledTrace
+func (a *Agent) sample(pt processedTrace) (sampled bool) {
```
gbbr: Sorry for going back on my original decision, but the word "sample" just repeats itself too much, and "sampled"doesn't sound right for the return value.

This method should have the same signature as the other sampling methods, as it is pretty much doing the same thing:
```go
func (a *Agent) sample(pt processedTrace) (rate float64, ok bool)
```
Please use "rate" instead of "sample rate" and "ok" instead of sampled. You don't have to use these variables inside the function body, they are here only for documentation.

This function returns the _rate_ of the sampler and whether it's _ok_ to sample or not. The word "sampled" is not correct here, it is the past tense, specifying that something happened. After this function was called nothing was sampled, it simply returns a decision (ok to sample or not), the actual sampling happens in the Process function.

Not only that but the word "ok" is standard in Go. You can search also the standard library and see for yourself, some examples:
```
/usr/local/go/src/cmd/asm/internal/asm/parse.go|488 col 78| func (p *Parser) register(name string, prefix rune) (r1, r2 int16, scale int8, ok bool) {
/usr/local/go/src/cmd/asm/internal/arch/arm.go|148 col 98| func ARMMRCOffset(op obj.As, cond string, x0, x1, x2, x3, x4, x5 int64) (offset int64, op0 obj.As, ok bool) {
/usr/local/go/src/cmd/go/internal/load/pkg.go|736 col 42| func findInternal(path string) (index int, ok bool) {
/usr/local/go/src/cmd/go/internal/load/pkg.go|841 col 40| func FindVendor(path string) (index int, ok bool) {
/usr/local/go/src/cmd/go/internal/load/path.go|16 col 45| func hasSubdir(root, dir string) (rel string, ok bool) {
/usr/local/go/src/cmd/compile/internal/gc/main.go|826 col 39| func findpkg(name string) (file string, ok bool) {
/usr/local/go/src/cmd/cgo/util.go|19 col 61| func run(stdin []byte, argv []string) (stdout, stderr []byte, ok bool) {
/usr/local/go/src/cmd/fix/main_test.go|37 col 112| func parseFixPrint(t *testing.T, fn func(*ast.File) bool, desc, in string, mustBeGofmt bool) (out string, fixed, ok bool) {
...
```
Please make this change and don't forget to put `ok` last.

cmd/trace-agent/agent.go
```diff
@@ -272,40 +272,35 @@ func (a *Agent) Process(t model.Trace) {
...
 
-		sampled := false
-		for _, s := range samplers {
-			// Consider trace as sampled if at least one of the samplers kept it.
-			sampled = s.Add(pt) || sampled
-		}
-		if sampled {
-			sampledTrace.Trace = &pt.Trace
-		}
+	_, hasPriority := pt.Root.Metrics[sampler.SamplingPriorityKey]
```
gbbr:
```go
if _, ok := pt.Root.Metrics[sampler.SamplingPriorityKey]; ok {
```

cmd/trace-agent/agent.go
```diff
@@ -272,40 +272,40 @@ func (a *Agent) Process(t model.Trace) {
...
+		scoreSampled, scoreSampleRate = a.ScoreSampler.Add(pt)
+	}
 
-		if !sampledTrace.Empty() {
-			a.sampledTraceChan <- &sampledTrace
-		}
-	}()
+	sampled = sampled || scoreSampled
+	sampleRate = sampler.MergeParallelSamplingRates(scoreSampleRate, sampleRate)
+	sampler.AddSampleRate(pt.Root, sampleRate)
```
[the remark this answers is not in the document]
gbbr: You didn't follow this comment.

cmd/trace-agent/agent_test.go
```diff
@@ -210,6 +210,66 @@ func TestProcess(t *testing.T) {
...
+			priorityRate: 0.1,
+			wantRate:     0.5,
+		},
+		"error and priority": {
+			hasErrors:      true,
+			hasPriority:    true,
+			scoreErrorRate: 0.8,
+			priorityRate:   0.2,
+			wantRate:       sampler.CombineRates(0.8, 0.2),
+		},
```
gbbr: You didn't completely follow this comment either, we need to all the scenarios here:
* errors (true) + priority (true)
* errors (true) + priority (false)
* errors (false) + priority (true)
* errors only (true)
* errors only (false)
* score (true) + priority (true)
* score (true) + priority (false)
* score (false) + priority (true)
* score only (true)
* score only (false)

It is important to guarantee that the sampler works correctly in all expected scenarios.

cmd/trace-agent/agent_test.go
```diff
@@ -210,6 +210,66 @@ func TestProcess(t *testing.T) {
...
+			a := &Agent{
+				ScoreSampler:       newMockSampler(false, tt.scoreRate),
+				ErrorsScoreSampler: newMockSampler(false, tt.scoreErrorRate),
+				PrioritySampler:    newMockSampler(false, tt.priorityRate),
+			}
+			root := &model.Span{
+				Service:  "serv1",
+				Start:    time.Now().UnixNano(),
+				Duration: (100 * time.Millisecond).Nanoseconds(),
+				Metrics:  map[string]float64{}}
```
gbbr: Please put all closing curly braces on the next line.

cmd/trace-agent/agent_test.go
```diff
@@ -210,6 +210,66 @@ func TestProcess(t *testing.T) {
...
+				root.Error = 1
+			}
+			pt := processedTrace{
+				Trace: model.Trace{root},
+				Root:  root}
+			if tt.hasPriority {
+				pt.Root.Metrics[sampler.SamplingPriorityKey] = 1
+			}
+
+			a.sample(pt)
```
gbbr: Here you shouldn't ignore the return values. That's the output of the test. You will obtain `rate` and `ok` and see if they match the tests expectations.

cmd/trace-agent/agent_test.go
```diff
@@ -210,6 +210,66 @@ func TestProcess(t *testing.T) {
...
+				Start:    time.Now().UnixNano(),
+				Duration: (100 * time.Millisecond).Nanoseconds(),
+				Metrics:  map[string]float64{}}
+
+			if tt.hasErrors {
+				root.Error = 1
+			}
+			pt := processedTrace{
+				Trace: model.Trace{root},
+				Root:  root}
```
gbbr: Here also, curly on the next line, or everything in one line. As you wish.

cmd/trace-agent/sampler.go
```diff
@@ -62,15 +62,16 @@ func (s *Sampler) Run() {
...
```
gbbr: As explained above, please use the suggested signature, "sampled" is not the right word here, nothing got sampled, it is simply an indication whether the called should or shouldn't sample. You also didn't reverse the order:
```go
func (s *Sampler) Add(t processedTrace) (rate float64, ok bool)
```

cmd/trace-agent/sampler.go
```diff
@@ -63,15 +63,16 @@ func (s *Sampler) Run() {
...
 	atomic.AddUint64(&s.totalTraceCount, 1)
 
-	if s.engine.Sample(t.Trace, t.Root, t.Env) {
+	sampled, sampleRate = s.engine.Sample(t.Trace, t.Root, t.Env)
```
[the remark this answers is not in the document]
gbbr: Please follow this comment. The word "sample" is repeated too much. Makes it hard to follow reading the code.

sampler/coresampler.go
```diff
@@ -52,7 +52,7 @@ type Engine interface {
 	// Stop the sampler.
 	Stop()
 	// Sample a trace.
-	Sample(trace model.Trace, root *model.Span, env string) bool
+	Sample(trace model.Trace, root *model.Span, env string) (sampled bool, samplingRate float64)
```
gbbr: Same here:
```go
Sample(trace model.Trace, root *model.Span, env string) (rate float64, ok bool)
```
No need to say "samplingRate", "rate" should do.

sampler/coresampler.go
```diff
@@ -179,3 +179,56 @@ func GetTraceAppliedSampleRate(root *model.Span) float64 {
...
+
+// AddSampleRate adds a new sampling rate to the trace sampling rate. Previous and new sampling rate must be independant
+// and the sampling decisions sequential.
+func AddSampleRate(root *model.Span, sampleRate float64) {
+	initialRate := GetTraceAppliedSampleRate(root)
+	newRate := initialRate * sampleRate
+	SetTraceAppliedSampleRate(root, newRate)
+}
+
+// MockEngine mocks a sampler engine
```
gbbr: This is a test component, so it must go in a test file, maybe `coresampler_test.go`?

sampler/prioritysampler.go
gbbr: Did you agree? Are we keeping this?

cmd/trace-agent/sampler.go
```diff
@@ -118,3 +119,12 @@ func (s *Sampler) logStats() {
 		}
 	}
 }
+
+type mockSamplerEngine struct {
```
gbbr: This must also go in a test file.

cmd/trace-agent/sampler.go
```diff
@@ -118,3 +119,12 @@ func (s *Sampler) logStats() {
...
+	engine sampler.Engine
+}
+
+func newMockSampler(wantSampled bool, wantRate float64) *Sampler {
+	return &Sampler{
+		engine: sampler.NewMockEngine(wantSampled, wantRate)}
```
gbbr: Curly on next line here too please 🙏

cmd/trace-agent/agent.go
gbbr: As discussed offline, we can keep `sampled` because `ok` could hint at the fact that the `rate` is not reliable, which is false. Resolving comment.

sampler/prioritysampler.go
```diff
@@ -86,41 +88,53 @@ func (s *PriorityEngine) Stop() {
 }
 
 // Sample counts an incoming trace and tells if it is a sample which has to be kept
-func (s *PriorityEngine) Sample(trace model.Trace, root *model.Span, env string) bool {
+func (s *PriorityEngine) Sample(trace model.Trace, root *model.Span, env string) (sampled bool, sampleRate float64) {
```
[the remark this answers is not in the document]
gbbr: As discussed, let's go with `(sampled bool, rate float64)` here.

PR #495, opened 2018-10-16
{cmd/trace-agent,writer}: fix race conditions

gbbr:
Fixes some (non-fatal) race conditions detected by the `-race` flag.

### DIFF
--- a/cmd/trace-agent/sampler.go
+++ b/cmd/trace-agent/sampler.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"reflect"
+	"sync/atomic"
 	"time"
 
 	log "github.com/cihub/seelog"
@@ -16,9 +17,10 @@ import (
 // Sampler chooses wich spans to write to the API
 type Sampler struct {
 	// For stats
-	keptTraceCount  int
-	totalTraceCount int
-	lastFlush       time.Time
+	keptTraceCount  uint64
+	totalTraceCount uint64
+
+	lastFlush time.Time
 
 	// actual implementation of the sampling logic
 	engine sampler.Engine
@@ -62,10 +64,10 @@ func (s *Sampler) Run() {
 
 // Add samples a trace and returns true if trace was sampled (should be kept), false otherwise
 func (s *Sampler) Add(t processedTrace) bool {
-	s.totalTraceCount++
+	atomic.AddUint64(&s.totalTraceCount, 1)
 
 	if s.engine.Sample(t.Trace, t.Root, t.Env) {
-		s.keptTraceCount++
+		atomic.AddUint64(&s.keptTraceCount, 1)
 		return true
 	}
 
@@ -79,12 +81,9 @@ func (s *Sampler) Stop() {
 
 // logStats reports statistics and update the info exposed.
 func (s *Sampler) logStats() {
-
 	for now := range time.Tick(10 * time.Second) {
-		keptTraceCount := s.keptTraceCount
-		totalTraceCount := s.totalTraceCount
-		s.keptTraceCount = 0
-		s.totalTraceCount = 0
+		keptTraceCount := atomic.SwapUint64(&s.keptTraceCount, 0)
+		totalTraceCount := atomic.SwapUint64(&s.totalTraceCount, 0)
 
 		duration := now.Sub(s.lastFlush)
 		s.lastFlush = now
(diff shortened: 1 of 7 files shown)


cmd/trace-agent/sampler.go
```diff
@@ -16,9 +18,11 @@ import (
...
 type Sampler struct {
 	// For stats
-	keptTraceCount  int
-	totalTraceCount int
-	lastFlush       time.Time
+	keptTraceCount  uint64
+	totalTraceCount uint64
+
+	mu        sync.Mutex // guards lastFlush
+	lastFlush time.Time
```
gbbr: Are the `sampler.Engine` implementations thread-safe?

cmd/trace-agent/sampler.go
```diff
@@ -79,15 +83,14 @@ func (s *Sampler) Stop() {
...
-
 	for now := range time.Tick(10 * time.Second) {
-		keptTraceCount := s.keptTraceCount
-		totalTraceCount := s.totalTraceCount
-		s.keptTraceCount = 0
-		s.totalTraceCount = 0
+		keptTraceCount := atomic.SwapUint64(&s.keptTraceCount, 0)
+		totalTraceCount := atomic.SwapUint64(&s.totalTraceCount, 0)
 
+		s.mu.Lock()
```
AlexJF: Why the need for the lock? No one else touches `lastFlush` except the logStats routine?

writer/fixtures_test.go
```diff
@@ -99,8 +99,7 @@ func (c *testPayloadSender) Start() {
 
 // Run executes the core loop of this sender.
 func (c *testPayloadSender) Run() {
-	c.exitWG.Add(1)
-	defer c.exitWG.Done()
+	defer close(c.exit)
```
AlexJF: This will lead to double close on exit. We probably just want to move the exitWG interactions to `Start`?

writer/fixtures_test.go
```diff
@@ -187,8 +184,8 @@ func (m *testPayloadSenderMonitor) Run() {
 
 // Stop stops this payload monitor and waits for it to stop.
 func (m *testPayloadSenderMonitor) Stop() {
-	close(m.exit)
-	m.exitWG.Wait()
+	m.exit <- struct{}{}
```
bmermet: This works fine, but FYI for the general case of handling goroutine terminations the tomb package provides a nice abstraction: https://godoc.org/gopkg.in/tomb.v2

writer/fixtures_test.go
```diff
@@ -187,8 +184,8 @@ func (m *testPayloadSenderMonitor) Run() {
...
+	<-m.exit
```
AlexJF: This is a racy termination. `Stop` may read the added struct before the routine executing `Run` does so in which case `Stop` will essentially do a no-op.

AlexJF: Oh wait, no it's not racy because exit is not buffered right? So it will only move to the next line if Run has already received the packet. Well, I guess it works but kinda feels less intuitive, at least to me.

gbbr: > This is a racy termination. Stop may read the added struct before the routine executing Run does so in which case Stop will essentially do a no-op.

I don't get this point. The send will block until someone reads from the channel, which will be the loop. Then the loop exits and the channel closes. There is no race here. Try to reproduce it in a test.

gbbr:
> Here's how I think a Start/Stop should look for correctness and no races:

Actually that's the original race right there. Calling `Wait` while `Add`-ing will cause a race condition. Check the documentation for [`sync.WaitGroup.Add`](https://golang.org/pkg/sync/#WaitGroup.Add)

cmd/trace-agent/sampler.go
gbbr: DOH, true.

writer/fixtures_test.go
gbbr: I'd rather not add a new dependency now. If it makes sense, we'll add it as part of a bigger refactor and use it throughout. That package could be used in many components including this one.

gbbr:
@AlexJF @bmermet PTAL

AlexJF:
Right. Although having moved all the waitgroup interactions to just `Start` and `Stop`, triggering the race condition would require a concurrent `Start` and `Stop` which is by definition racy.

PR #496, opened 2018-10-17
writer: allow writing to multiple endpoints.

gbbr:
### Description

This change adds a new set of options to the `apm_config` sub-section of the config file, enabling the possibility of forwarding data (traces, stats & services) to additional endpoints besides the main one.

Example configuration:
```yaml
apm_config:
  ...
  # The trace agent will also forward traces, stats and services to these endpoints,
  # besides the main endpoint.
  additional_endpoints:
    # Simply specify the list of URLs to forward to along with one or more API keys
    # to use for that URL.
    https://trace.agent.datadoghq.eu:
    - 1234abcd
    - 4567efghi
    https://trace.agent.other-endpoint.com:
    - 890jkl
```

### References
- Logs: https://github.com/DataDog/datadog-agent/pull/2381
- Infrastructure: https://github.com/DataDog/datadog-agent/pull/352
- Process: https://github.com/DataDog/datadog-process-agent/pull/147


gbbr:
@olivielpeau can you please check the `yaml` snippet in the description of this PR? No need to review the code. I'm curious if you are ok with it. It differs a bit from process & infra (but I think it's more similar to logs). Should we all use the same? If we use the map structure (like process & infra) we won't be able to add more options such as maybe using SSL validation or not on a per-endpoint basis, or whatever might come up later.

olivielpeau:
@gbbr Thanks for the ping. I feel Logs has pretty specific use cases (doesn't use HTTP, and they expect every endpoint to require more parameters than simply a URL and an api key), which AFAIK don't apply to Infra/Trace/Process. So I would prefer trace to remain consistent with infra and process here, unless there's already a use case today that's specific to trace and justifies using a format that's closer to Logs'.

To me this makes even more sense as we try to make infra/process/trace's configuration use the same library, this logic could be shared. If we need to make the format evolve in the future we could then easily do it consistently across infra/trace/process, which would make the change more easy to understand on the customers' side as well IMO.

gbbr:
@olivielpeau many thanks for your input, I've changed the config to match the rest and updated the PR description.

gbbr:
@AlexJF @palazzem ready for review

config/agent.go
```diff
@@ -38,9 +45,10 @@ type AgentConfig struct {
 	ConfigPath string // the source of this config, if any
 
 	// API
-	APIEndpoint string
-	APIKey      string `json:"-"` // never publish this
-	APIEnabled  bool
+	APIEndpoint         string
```
gbbr: Maybe I should merge these two into one:
```go
    Endpoints []*Endpoint
```

cmd/trace-agent/main.go
```diff
@@ -138,15 +138,13 @@ func runAgent(ctx context.Context) {
 	}
 
 	// Initialize dogstatsd client
-	err = statsd.Configure(cfg)
+	err = statsd.Configure(cfg, []string{"version:" + info.Version})
```
AlexJF: Uhh, nice!

config/agent.go
AlexJF: Agree, don't see the point of treating main vs additional differently and if we need to we can just ensure the main is the first in this endpoints list.

config/merge_yaml.go
```diff
@@ -210,13 +211,23 @@ func (c *AgentConfig) loadYamlConfig(yc *YamlAgentConfig) {
...
+				Host:    url,
+				APIKey:  key,
+				NoProxy: noProxy[url],
+			})
 		}
 	}
+	if host, ok := noProxy[c.APIEndpoint]; ok {
+		log.Infof("Trace Agent main endpoint matches `proxy.no_proxy` list item %q: ignoring proxy", host)
+		c.NoProxy = true
+	}
```
AlexJF: Merging all endpoints would also simplify this part

writer/datadog_endpoint.go
```diff
@@ -22,38 +27,67 @@ var userAgent = fmt.Sprintf(
...
+	for _, e := range conf.AdditionalEndpoints {
+		c := client
+		if e.NoProxy != conf.NoProxy {
+			// this endpoint needs a different client.
+			c = newClient(conf, e.NoProxy)
+		}
+		apiKey := e.APIKey
+		if apiKey == "" {
+			// if this endpoint doesn't have its own API key, try the main one.
+			apiKey = conf.APIKey
```
AlexJF: I wouldn't even allow this to occur. Looking at the infra PR, they dedupe endpoints missing apikey already at the config: https://github.com/DataDog/datadog-agent/pull/352/files#diff-091ace02b49a46f55f2586cfec89a160R95

writer/datadog_endpoint.go
```diff
@@ -88,5 +122,20 @@ func (e *DatadogEndpoint) Write(payload *Payload) error {
...
+	return fmt.Sprintf("DataDogEndpoint(%q)", e.Host+e.path)
+}
+
+// timeout is the HTTP timeout for POST requests to the Datadog backend
+const timeout = 10 * time.Second
+
+// newClient returns a http.Client configured with the Agent options.
+func newClient(conf *config.AgentConfig, noProxy bool) *http.Client {
+	transport := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: conf.SkipSSLValidation},
```
AlexJF: We've had some issues with some other code that was using custom transports. We should probably include all settings set in `DefaultTransport` as well otherwise most timeouts will be zeroed out and we'll wait for much longer than we should when network issues happen.

https://golang.org/src/net/http/transport.go (line 42). I'm fine with this going to another PR but we should track this.

writer/multi_writer.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// newMultiSender returns a new factory to generate PayloadSenders capable of forwarding
+// payloads to multiple endpoints.
+func newMultiSender(cfg config.QueuablePayloadSenderConf) func([]Endpoint) PayloadSender {
+	return func(endpoints []Endpoint) PayloadSender {
+		if len(endpoints) == 1 {
+			return NewCustomQueuablePayloadSender(endpoints[0], cfg)
+		}
+		var senders []PayloadSender
```
AlexJF: Minor nit, we can already allocate enough space for `len(endpoints)`.

writer/multi_writer.go
```diff
@@ -0,0 +1,79 @@
...
+	mwg     sync.WaitGroup   // monitor funnel waitgroup
+	mch     chan interface{} // monitor funneling channel
+}
+
+// newMultiSender returns a new factory to generate PayloadSenders capable of forwarding
+// payloads to multiple endpoints.
+func newMultiSender(cfg config.QueuablePayloadSenderConf) func([]Endpoint) PayloadSender {
+	return func(endpoints []Endpoint) PayloadSender {
+		if len(endpoints) == 1 {
+			return NewCustomQueuablePayloadSender(endpoints[0], cfg)
```
AlexJF: Nice shortcut here, saving a bunch of overhead if not needed :+1:

writer/multi_writer.go
```diff
@@ -0,0 +1,79 @@
...
+// events.
+type multiSender struct {
+	senders []PayloadSender
+	mwg     sync.WaitGroup   // monitor funnel waitgroup
+	mch     chan interface{} // monitor funneling channel
+}
+
+// newMultiSender returns a new factory to generate PayloadSenders capable of forwarding
+// payloads to multiple endpoints.
+func newMultiSender(cfg config.QueuablePayloadSenderConf) func([]Endpoint) PayloadSender {
```
AlexJF: The name here threw me off initially and breaks expectations imo.

Can we have a `newMultiSender(cfg, []Endpoint)` and turn this into 

```
newMultiSenderFactory(cfg) func([]Endpoint) {
   return func (endpoints) PayloadSender {
      return newMultiSender(cfg, endpoints);
  }
}
```

writer/service_writer_test.go
```diff
@@ -202,7 +206,10 @@ func testServiceWriter() (*ServiceWriter, chan model.ServicesMetadata, *testEndp
...
 	serviceWriter.BaseWriter.payloadSender.setEndpoint(testEndpoint)
 	testStatsClient := &testutil.TestStatsClient{}
-	serviceWriter.statsClient = testStatsClient
+	originalClient := statsd.Client
+	statsd.Client = testStatsClient
 
-	return serviceWriter, serviceChannel, testEndpoint, testStatsClient
+	return serviceWriter, serviceChannel, testEndpoint, testStatsClient, func() {
+		statsd.Client = originalClient
+	}
```
AlexJF: Yeah, this makes it a bit less of a hassle :joy:

writer/trace_writer.go
```diff
@@ -269,6 +268,7 @@ func (w *TraceWriter) resetBuffer() {
 }
 
 func (w *TraceWriter) updateInfo() {
+	// TODO(gbbr): Scope these stats per endpoint (see (config.AgentConfig).AdditionalEndpoints))
```
AlexJF: Lets add this TODO to the other writers as well?

writer/datadog_endpoint.go
gbbr: You're very right here, this was indeed an oversight. Great catch! I've copied the `http.DefaultTransport` except the `Proxy: http.ProxyFromEnvironment` part which I assume was likely also an oversight on our behalf. We can ponder whether to introduce that back in a separate topic. I'm afraid not to cause any backwards breaking changes because of introducing it.

config/agent.go
```diff
@@ -24,6 +24,13 @@ var (
 	ErrMissingHostname = errors.New("failed to automatically set the hostname, you must specify it via configuration for or the DD_HOSTNAME env var")
 )
 
+// Endpoint specifies an endpoint that the trace agent will write data (traces, stats & services) to.
+type Endpoint struct {
+	APIKey  string
```
AlexJF: Also, we might want to add `json:"-" // Never publish this` here like we have for the main `APIKey`?

writer/trace_writer.go
AlexJF: Sounds good!

config/merge_yaml.go
gbbr: A little :)

config/merge_env.go
```diff
@@ -54,23 +54,23 @@ func (c *AgentConfig) loadEnv() {
 		c.Hostname = v
 	}
 
+	if len(c.Endpoints) == 0 {
+		c.Endpoints = make([]*Endpoint, 1)
+	}
 	site := os.Getenv(envSite)
 	if site != "" {
-		c.APIEndpoint = apiEndpointPrefix + site
+		c.Endpoints[0].Host = apiEndpointPrefix + site
```
AlexJF: This works? Won't `c.Endpoints[0]` be a `nil` here?

Edit: Oh I think this works because the DefaultConfig already includes a default Endpoint[0] so we'd never actually execute the `if len(c.Endpoints) == 0` part.

gbbr: Ah, how could I miss this. Good catch. You are right about the defaults, but I'd like this to work even when `New` is not used. I've changed the code from `make` to:
```go
c.Endpoints = []*Endpoints{{}}
```
That should do it.

gbbr:
@AlexJF good catch, changed in next commit.

gbbr:
Seems Github is buggy. Kept getting errors from using the "Squash & Merge" button and PR is still open even though it was merged twice (see [master commits](https://github.com/DataDog/datadog-trace-agent/commits/master))

gbbr:
Seems Github is buggy. Kept getting errors from using the "Squash & Merge" button and PR is still open even though it was merged twice (see [master commits](https://github.com/DataDog/datadog-trace-agent/commits/master))

gbbr:
Seems Github is buggy. Kept getting errors from using the "Squash & Merge" button and PR is still open even though it was merged twice (see [master commits](https://github.com/DataDog/datadog-trace-agent/commits/master))

PR #503, opened 2018-10-24
writer: simplify monitor

gbbr:
In this change:
* Move `backoff` package into `writer` folder.
* Use only one struct. for events with a type field to avoid reflection.
* Slightly minimise memory copying.


writer/multi_writer.go
```diff
@@ -65,7 +65,7 @@ func (w *multiSender) Send(p *Payload) {
 	}
 }
 
-func (w *multiSender) Monitor() <-chan interface{} { return w.mch }
+func (w *multiSender) monitor() <-chan monitorEvent { return w.mch }
```
AlexJF: Why make `Monitor` private? This way, no one outside of this package will be able to use a sender?

writer/payload.go
```diff
@@ -29,32 +28,36 @@ func NewPayload(bytes []byte, headers map[string]string) *Payload {
...
-// SendStats represents basic stats related to the sending of a payload.
-type SendStats struct {
-	SendTime time.Duration
-	Host     string
-}
+// eventType is a type of event sent down the monitor channel.
+type eventType int
+
+const (
+	eventTypeSuccess eventType = iota
```
AlexJF: Again, if these are private no one outside of this package will be able to use a sender. If this is indeed your intention we should set the whole PayloadSender and everything directly related to it as private and only leave actual writers public.

gbbr: That's the plan, yeah. Nobody's using them outside of `writer`, so there is no point to export them or make them visible in the public documentation. Most of these packages should be internal anyway.

Would you like me to do that within the scope of this PR?

gbbr:
Thank you Alex.

PR #504, opened 2018-10-24
writer: make internal identifiers private

gbbr:
* Makes internal `writer` identifiers private
* Absorbs `BasePayloadSender` into `QueuablePayloadSender`
* Renames `QueuablePayloadSender` to `queuableSender`.

### DIFF
--- a/writer/endpoint.go
+++ b/writer/endpoint.go
@@ -2,49 +2,41 @@ package writer
 
 import (
 	"fmt"
-	"net/http"
 
 	log "github.com/cihub/seelog"
 )
 
 const languageHeaderKey = "X-Datadog-Reported-Languages"
 
-// Endpoint is an interface where we send the data from the Agent.
-type Endpoint interface {
+// endpoint is an interface where we send the data from the Agent.
+type endpoint interface {
 	// Write writes the payload to the endpoint.
-	Write(payload *Payload) error
+	write(payload *payload) error
 
-	// BaseURL returns the base URL for this endpoint. e.g. For the URL "https://trace.agent.datadoghq.eu/api/v0.2/traces"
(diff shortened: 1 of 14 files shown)


writer/multi_writer_test.go
```diff
@@ -163,7 +163,7 @@ func (m *mockPayloadSender) Reset() {
 	m.mu.Unlock()
 }
 
-func (m *mockPayloadSender) Start() {
+func (m *mockPayloadSender) start() {
```
AlexJF: Does it make sense to turn every method private? That way when using this class inside the writer package, it's not evident what constitutes the "public" interface of a sender vs what are internal helper methods. Perhaps it's sufficient to set the structs/interfaces themselves as private?

gbbr:
Ah, ok. I see what you mean now. I was looking at it only through the interface's methods, not through the entire structure.

writer/endpoint.go
```diff
@@ -2,49 +2,41 @@ package writer
...
 
 const languageHeaderKey = "X-Datadog-Reported-Languages"
 
-// Endpoint is an interface where we send the data from the Agent.
-type Endpoint interface {
+// endpoint is an interface where we send the data from the Agent.
+type endpoint interface {
 	// Write writes the payload to the endpoint.
-	Write(payload *Payload) error
+	write(payload *payload) error
```
AlexJF: As a matter of consistency we should strive for this uppercase public interface methods everywhere, what do you think?

gbbr:
Done 👍

writer/endpoint.go
gbbr: I wouldn't want to enforce that. I get the logic in the sender because of `setEndpoint` which is specifically there for tests, but I wouldn't want to make this a rule. If we capitalise these, it means that technically they can be accessed from outside the package too if this interface would be returned anywhere.

gbbr:
Thanks for helping out. Indeed `internal` is what's needed here.

PR #506, opened 2018-10-25
cmd/trace-agent: cleanup code related to APM events.

AlexJF:
Move most of the code dealing with event extraction into its own package and rename samplers to extractors.

Also, change most references of transactions into events following the documented nomenclature.

This sets the ground for https://github.com/DataDog/datadog-trace-agent/pull/508


constants/sampling.go
```diff
@@ -0,0 +1,4 @@
+package constants
+
+// SamplingPriorityKey is the key of the sampling priority value in the metrics map of the root span
+const SamplingPriorityKey = "_sampling_priority_v1"
```
AlexJF: ^ Regarding this file, I added this to fix some cyclic imports. Happy to move this somewhere else that also works 😄

gbbr: Let's move it into `model` where it originally was before I plucked it out, as discussed.

event/extractor/analyzed.go
```diff
@@ -0,0 +1,52 @@
+package eventextractor
```
gbbr: Let's just have one main package `event`, as discussed.

event/extractor/analyzed.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/DataDog/datadog-trace-agent/model"
+	"github.com/DataDog/datadog-trace-agent/sampler"
+)
+
+// AnalyzedExtractor is an event extractor that extracts APM events from traces based on
+// `(service name, operation name) => sampling ratio` mappings.
+type AnalyzedExtractor struct {
```
gbbr: I'd like to suggest calling it like this:
```go
// ratioExtractor extracts APM events from traces based on `(service name, operation name) => sampling ratio`
// mappings.
type ratioExtractor struct {
```
What do you think? The word analysed was always a bit confusing to me. But maybe I'm misunderstanding something.

event/extractor/analyzed.go
```diff
@@ -0,0 +1,52 @@
...
+	analyzedSpansByService map[string]map[string]float64
+}
+
+// NewAnalyzed returns an APM event extractor that extracts APM events from a trace following the provided
+// extraction rates for any spans matching a (service name, operation name) pair.
+func NewAnalyzed(analyzedSpansByService map[string]map[string]float64) *AnalyzedExtractor {
```
gbbr: If this is going back into a main `event` package, it will probably sound nice like this:
```go
event.NewRatioExtractor
```
Unless you think "analyzed" is better for some reason, which is fine by me. I just don't understand why this word is used, given that it doesn't appear in the documentation. What was analyzed (past tense) ?

event/extractor/analyzed_legacy.go
```diff
@@ -0,0 +1,51 @@
+package eventextractor
+
+import (
+	"github.com/DataDog/datadog-trace-agent/model"
+	"github.com/DataDog/datadog-trace-agent/sampler"
+)
+
+// LegacyAnalyzedExtractor is an event extractor that extracts APM events from traces based on `serviceName => sampling
+// ratio` mappings.
+type LegacyAnalyzedExtractor struct {
```
gbbr: Maybe this could also just be `legacyExtractor`?

event/extractor/analyzed_legacy.go
```diff
@@ -0,0 +1,51 @@
...
+	analyzedRateByService map[string]float64
```
gbbr: Do we need the word "analyzed"? Could it maybe simply be "rateByService"?

event/extractor/analyzed_legacy.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+}
+
+// Extract extracts analyzed spans from the trace and returns them as a slice
+func (s *LegacyAnalyzedExtractor) Extract(t model.ProcessedTrace, sampledTrace bool) []*model.APMEvent {
+	var events []*model.APMEvent
+
+	// inspect the WeightedTrace so that we can identify top-level spans
+	for _, span := range t.WeightedTrace {
+		if s.shouldAnalyze(span) {
```
gbbr: From here I take it that the word analyze is meant for the future, as in: should the backend analyze this? Right? In that case we can remove the word "analyzed" from all the structs, etc. Here it makes sense, on this method.

event/extractor/disabled.go
```diff
@@ -0,0 +1,15 @@
+package eventextractor
+
+import "github.com/DataDog/datadog-trace-agent/model"
+
+// DisabledExtractor is a no-op APM event extractor used when APM event extraction is disabled.
+type DisabledExtractor struct{}
+
+// NewDisabled returns a new APM event extractor that does not extract any events.
+func NewDisabled() *DisabledExtractor {
```
gbbr: This could just be `NoopExtractor` if you like, or `NewDisabledExtractor` (the longer version)

event/extractor/extractor.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/DataDog/datadog-trace-agent/config"
+	"github.com/DataDog/datadog-trace-agent/model"
+)
+
+// Extractor extracts APM event spans from a trace.
+type Extractor interface {
+	// Extract extracts APM event spans from the given weighted trace information and returns them.
+	Extract(t model.ProcessedTrace, sampledTrace bool) []*model.APMEvent
```
gbbr: I really like this! A lot! I think this is a great idea. For this reason, I'd like to make this the main interface to interact with this package. What do you think? This means:
* Make all `Extractor` implementations (`noopExtractor`, `ratioExtractor` and `legacyExtractor`) unexported.
* Keep all constructors exported (`NewRatioExtractor`, `NewLegacyExtractor` and `NoopExtractor`) and have them all return an `Extractor` focusing the documentation about how they work on top of the constructor.

I think this will give for a great package. Having this interface will make a lot of sense then. What do you think?

event/extractor/extractor.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+// FromConf creates a new APM event extractor based on the provided agent configuration.
+func FromConf(conf *config.AgentConfig) Extractor {
```
gbbr: I'm wondering if this is the place to have this? I'm a bit skeptical about passing around the `AgentConfig`. It always seemed to me that this config was meant for the agent itself only. Maybe this logic belongs in there? (currently `cmd/trace-agent`)?

event/extractor/extractor.go
gbbr: One last thing here: is there any way we can get rid of the second parameter? Is it needed? Wasn't this trace already sampled at this point? Or, would it make sense to include it into the `ProcessedTrace` struct instead? I think in your follow-up PR there can be a middle value (no decision) where the trace has neither been considered nor disconsidered yet.

cmd/trace-agent/agent.go
```diff
@@ -191,7 +169,7 @@ func (a *Agent) Process(t model.Trace) {
 	ts := a.Receiver.Stats.GetTagStats(info.Tags{})
 
 	// Extract priority early, as later goroutines might manipulate the Metrics map in parallel which isn't safe.
-	priority, hasPriority := root.Metrics[sampler.SamplingPriorityKey]
+	priority, hasPriority := root.GetSamplingPriority()
```
gbbr: 👍

cmd/trace-agent/agent.go
```diff
@@ -65,7 +43,7 @@ type Agent struct {
 	// tags based on their type.
 	obfuscator *obfuscate.Obfuscator
 
-	sampledTraceChan chan *writer.SampledTrace
+	tracePkgChan chan *writer.TracePackage
```
gbbr: ❤️

cmd/trace-agent/agent.go
```diff
@@ -258,38 +236,44 @@ func (a *Agent) Process(t model.Trace) {
 		a.ServiceExtractor.Process(pt.WeightedTrace)
 	}()
 
-	go func() {
+	go func(pt model.ProcessedTrace) {
 		defer watchdog.LogOnPanic()
 		// Everything is sent to concentrator for stats, regardless of sampling.
 		a.Concentrator.Add(pt)
-	}()
+	}(pt)
```
gbbr: 💯

cmd/trace-agent/agent.go
```diff
@@ -258,38 +236,44 @@ func (a *Agent) Process(t model.Trace) {
...
 		defer watchdog.LogOnPanic()
 
+		tracePkg := writer.TracePackage{}
+
 		sampled, rate := a.sample(pt)
-		if !sampled {
-			return
+
+		if sampled {
+			pt.Sampled = sampled
```
gbbr: This can be both inside the `if` and outside. Doesn't really matter, right? 😄

event/extractor.go
```diff
@@ -0,0 +1,11 @@
+package event
```
gbbr: Would be nice to document this package. It'd provide a nice opportunity to explain what events are in APM.

event/extractor_fixed_rate.go
```diff
@@ -0,0 +1,52 @@
...
+				Span:         span.Span,
+				TraceSampled: t.Sampled,
+			})
+		}
+	}
+
+	return events
+}
+
+func (s *fixedRateExtractor) shouldAnalyze(span *model.WeightedSpan, hasPriority bool, priority int) bool {
```
gbbr: In the other extractor it's called `shouldExtractEvent`. It'd be nice to chose one and use the same in both.

writer/trace_writer.go
```diff
@@ -19,15 +19,18 @@ import (
 
 const pathTraces = "/api/v0.2/traces"
 
-// SampledTrace represents the result of a trace sample operation.
-type SampledTrace struct {
-	Trace        *model.Trace
-	Transactions []*model.Span
+// TracePackage represents the result of a trace sampling operation.
+//
+// If a trace was sampled, then Trace will be set to that trace. Otherwise, it will be nil.
```
gbbr: These two lines seem like they would be destined above each of the fields in the struct below 😊

writer/trace_writer.go
```diff
@@ -19,15 +19,18 @@ import (
 
 const pathTraces = "/api/v0.2/traces"
 
-// SampledTrace represents the result of a trace sample operation.
-type SampledTrace struct {
-	Trace        *model.Trace
-	Transactions []*model.Span
+// TracePackage represents the result of a trace sampling operation.
```
gbbr: I'd add here (if you think it helps) that "A package can still be valid when Trace is `nil` if it contains Events. This happens when a trace wasn't sampled, but events were."

writer/trace_writer.go
```diff
@@ -19,15 +19,18 @@ import (
...
+//
+// If a trace was sampled, then Trace will be set to that trace. Otherwise, it will be nil.
+// If events were extracted from a trace, then Events will be populated from these events. Otherwise, it will be empty.
+type TracePackage struct {
+	Trace  *model.Trace
```
gbbr: It feels so unusual for this to be a pointer (to a slice, which is already technically a pointer).

writer/trace_writer.go
```diff
@@ -19,15 +19,18 @@ import (
...
+	Events []*model.APMEvent
 }
 
-// Empty returns true if this SampledTrace has no data.
-func (s *SampledTrace) Empty() bool {
-	return s.Trace == nil && len(s.Transactions) == 0
+// Empty returns true if this TracePackage has no data.
+func (s *TracePackage) Empty() bool {
+	return s.Trace == nil && len(s.Events) == 0
```
gbbr: I wonder why this wasn't `len(s.Trace) == 0` but it's fine.

writer/trace_writer_test.go
```diff
@@ -352,15 +355,21 @@ func testTraceWriter() (*TraceWriter, chan *SampledTrace, *testEndpoint, *testut
 	return traceWriter, payloadChannel, testEndpoint, testStatsClient
 }
 
-func randomSampledTrace(numSpans, numTransactions int) *SampledTrace {
-	if numSpans < numTransactions {
-		panic("can't have more transactions than spans in a RandomSampledTrace")
+func randomSampledTrace(numSpans, numEvents int) *TracePackage {
```
gbbr: s/randomSampledTrace/randomTracePackage/g

PR #508, opened 2018-10-25
event: introduce sampling with max events per second logic

AlexJF:
This PR introduces:
* A new `MaxEPS` (*M*ax *E*vents *P*er *S*econd) config option defaulting to 100.
* A MaxEPS event sampler.


cmd/trace-agent/agent.go
```diff
@@ -339,3 +343,21 @@ func eventExtractorFromConf(conf *config.AgentConfig) event.Extractor {
...
+	rateCounter.Start()
+
+	return event.NewBatchSampler(
+		event.NewSamplerChain(
+			[]event.Sampler{
+				// Sample all events for which their respective traces were sampled
+				event.NewSampledTraceSampler(),
+				// For those events that did not have its trace sampled, sample as many as possible respecting MaxEPS.
+				event.NewMaxEPSSampler(conf.MaxEPS, event.NewReadOnlyRateCounter(rateCounter)),
+			}, func(decision event.SamplingDecision) {
```
gbbr: Instead of having this callback here, can we not just pass the "rate counter" directly? Just like the "maxEPSSampler". Then you could add a `Stop` method to the batch sampler which could manage exiting the rate counter cleanly if needed.

event/sampler.go
```diff
@@ -0,0 +1,45 @@
+package event
+
+import (
+	"github.com/DataDog/datadog-trace-agent/model"
+)
+
+type SamplingDecision int8
+
+const (
+	NoDecision SamplingDecision = iota
```
gbbr: Let's use a shortened version of [`http` package style](https://golang.org/pkg/net/http/#pkg-constants) names here:
```go
const (
    DecisionNone SamplingDecision = iota
    DecisionSample
    DecisionDontSample
)
```
`iota` only once is enough.

event/sampler.go
```diff
@@ -0,0 +1,45 @@
+package event
+
+import (
+	"github.com/DataDog/datadog-trace-agent/model"
+)
+
+type SamplingDecision int8
```
gbbr: Let's make sure to document all exported identifiers, even if redundant. I think the linter should complains here anyway (or not?).

event/sampler_chain.go
```diff
@@ -0,0 +1,37 @@
...
+
+// NewSamplerChain creates a new sampler chain from the provided samplers and with the specified decision callback.
+func NewSamplerChain(samplers []Sampler, callback func(decision SamplingDecision)) Sampler {
+	return &samplerChain{
+		samplers: samplers,
+		callback: callback,
+	}
+}
+
+// Sample returns the first !NO_DECISION sampling decision from calls to the underlying samplers, in order.
```
gbbr:
```suggestion
// Sample returns the first DecisionDontSample sampling decision from calls to the underlying samplers, in order.
```
What does it mean? Does it always return this value? If none of the samplers return this value what happens then? We should try to clarify a bit.

event/sampler_chain.go
```diff
@@ -0,0 +1,37 @@
+package event
+
+import "github.com/DataDog/datadog-trace-agent/model"
+
+// samplerChain is an APMEvent sampler that submits an event to a sequence of samplers, returning as soon as one of
+// the samplers makes a decision.
+type samplerChain struct {
+	samplers []Sampler
+	callback func(decision SamplingDecision)
```
gbbr: Since there is only one use for this callback in the current PR, why not just use the `sampler.Backend` interface directly? This will make everything more obvious, cleaner, and easier to manage in the sense that:
* You could have the sampler chain `Start` the backend.
* You could have the sampler chain `Stop` the backend.
* Because of the two above points, in `cmd/trace-agent/agent.go` you would be able to cleanly exit the rate counter which you are starting (but never stopping). It has an internal routine, right? We should aim to exit cleanly even if at this apparent level it doesn't cause any fatal bugs.

event/sampler_chain.go
```diff
@@ -0,0 +1,37 @@
...
+func NewSamplerChain(samplers []Sampler, callback func(decision SamplingDecision)) Sampler {
+	return &samplerChain{
+		samplers: samplers,
+		callback: callback,
+	}
+}
+
+// Sample returns the first !NO_DECISION sampling decision from calls to the underlying samplers, in order.
+func (sc *samplerChain) Sample(event *model.APMEvent) SamplingDecision {
+	var result = NoDecision
```
gbbr: Why not just
```go
result := DecisionNone
```
`var` is not needed.

event/sampler_max_eps.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if sampled {
+		return Sample
+	}
+
+	return DontSample
+}
+
+// RateCounter keeps track of different event rates.
+type RateCounter interface {
```
gbbr: I'm not sure if this makes things easier or harder. You could remove it and just have a "read only" backend wrapper on top of the memory backend? Would it be overkill?

event/sampler_max_eps_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+			assert.InDelta(testCase.expectedSamplingRate, float64(sampled)/float64(len(testEvents)), testCase.expectedSamplingRate*testCase.deltaPct)
+
+			counter.AssertExpectations(t)
+		})
+	}
+}
+
+type MockRateCounter struct {
+	mock.Mock
```
gbbr: I'd like to avoid adding dependencies when possible. This could be used in many places in the project. We already have mock implementations elsewhere. If we are to introduce this dependency I'd prefer that we refactor all the mocks to use it, otherwise it will be just a mixture and future devs will end up copying the nearest to them (unfortunate but true).

This can just be:
```go
type MockRateCounter struct {
    mu sync.RWMutex
    countCalls uint64
    getRateCalls []float64
}

func (mc *MockRateCounter) Count() {
    mc.mu.Lock()
    mc.countCalls++
    mc.mu.Unlock()
}

func (mc *MockRateCounter) CountCalls() int {
    mc.mu.RLock()
    defer mc.mu.RUnlock()
    return int(mc.countCalls)
}

func (mc *MockRateCounter) GetRate(rate float64) {
    mc.mu.Lock()
    mc.getRateCalls = append(mc.getRateCalls, rate)
    mc.mu.Unlock()
}

func (mc *MockRateCounter) GetRateCalls() []float64 {
    mc.mu.RLock()
    defer mc.mu.RUnlock()
    return int(mc.getRateCalls)
}
```
Not much saved IMHO.

event/sampler_chain_test.go
```diff
@@ -0,0 +1,78 @@
...
+func TestChainSampler_NilCallback(t *testing.T) {
+	chain := NewSamplerChain([]Sampler{&FixedDecisionSampler{Sample}}, nil)
+	testEvent := &model.APMEvent{Span: testutil.RandomSpan()}
+
+	assert.NotPanics(t, func() {
+		chain.Sample(testEvent)
+	})
+}
+
+type FixedDecisionSampler struct {
```
gbbr: Is this actually `MockDecisionSampler` ?

event/sampler_max_eps.go
```diff
@@ -0,0 +1,105 @@
...
+	Count()
+	GetRate() float64
+}
+
+// SamplerBackendRateCounter is a RateCounter backed by a sampler.Backend.
+type SamplerBackendRateCounter struct {
```
gbbr: It feels a bit like this should just be another implementation of `sampler.Backend` in the `sampler` package, not here.

cmd/trace-agent/agent.go
```diff
@@ -262,6 +267,9 @@ func (a *Agent) Process(t model.Trace) {
 
 		// NOTE: Events can be extracted from non-sampled traces.
 		tracePkg.Events = a.EventExtractor.Extract(pt)
+		statsd.Client.Count("datadog.trace_agent.events.extracted", int64(len(tracePkg.Events)), nil, 1)
+		tracePkg.Events = a.EventSampler.Sample(tracePkg.Events)
```
gbbr: I'm curious because it seems a bit memory exhaustive to extract everything and then process it again: can we not just have the extractor have an EPS setting and do the sampling too?

event/sampler_max_eps.go
```diff
@@ -0,0 +1,110 @@
...
+
+	if sampled {
+		return DecisionSample
+	}
+
+	return DecisionDontSample
+}
+
+// rateCounter keeps track of different event rates.
+type rateCounter interface {
```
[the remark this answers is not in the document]
AlexJF: I'd rather not because the interface you propose is misleading (and was another reason why I added this intermediate layer). With that interface it sounds like we should only count *sampled* events. But actually we want to count *all* events.

Note that this will all change soon as we'll actually need to do uniform sampling here so we'll definitely not use the current sampler code (or this one will change a lot).

event/sampler_max_eps.go
```diff
@@ -0,0 +1,110 @@
...
+
+// Stop stops the underlying rate counter.
+func (s *maxEPSSampler) Stop() {
+	s.rateCounter.Stop()
+}
+
+// Sample determines whether or not we should sample the provided event in order to ensure no more than maxEPS events
+// are sampled every second.
+func (s *maxEPSSampler) Sample(event *model.APMEvent) SamplingDecision {
+	if event == nil {
```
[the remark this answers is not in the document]
AlexJF: We don't *need* it but I'd rather be safe than sorry. Wouldn't be the first time a surprise nil sneaks in.

cmd/trace-agent/agent.go
AlexJF: Not really (at least in a simple and clean way) because we'll actually have several extractors working in tandem. And also, separation of concerns.

As for the memory impact, we're mostly copying pointers to spans, not the entire span data so it's not that heavy.

AlexJF: Going back to this, we could inject a sampler into the extractor and call `Sample` directly during extraction. Adds tight coupling between the 2 but would save the intermediate slices.

gbbr: It's fine. I wish we had a benchmark for the `Process` function.

event/sampler_max_eps_test.go
```diff
@@ -0,0 +1,103 @@
...
+	type testCase struct {
+		events             []*model.APMEvent
+		maxEPS             float64
+		pastEPS            float64
+		expectedSampledPct float64
+		deltaPct           float64
+	}
+
+	testEvents := make([]*model.APMEvent, 1000)
+	for i, _ := range testEvents {
```
gbbr: This is used in both tests, can we make a helper function to generate a series of N test events?

event/sampler_max_eps_test.go
```diff
@@ -0,0 +1,103 @@
...
+		testEvents[i] = &model.APMEvent{Span: testutil.RandomSpan()}
+	}
+
+	testEventsSampledTraces := make([]*model.APMEvent, 1000)
```
gbbr: We can do the same here, make a helper function. It can be part of this package or `testutil` (but probably doesn't make sense to share it yet since it's not used anywhere else).

event/sampler_max_eps_test.go
```diff
@@ -0,0 +1,103 @@
...
+		"low EPS":      {testEvents, 100, 50, 1., 0},
+		"limit EPS":    {testEvents, 100, 100, 1., 0},
+		"overload EPS": {testEvents, 100, 150, 100. / 150., 0.05},
+		// We should always keep events for sampled traces even if we are above maxEPS
+		"overload EPS - sampled": {testEventsSampledTraces, 100, 500, 1., 0},
+		// We should always keep events for sampled traces even if we are above maxEPS
+		"nil events": {make([]*model.APMEvent, 5), 100, 0, 0, 0},
+	}
+
+	for name, testCase := range testCases {
```
gbbr: Can we please inline everything? That's what I've been doing everywhere and is seen in most examples with table tests. It makes everything more compact and easier to figure out. I mean this:
```go
for name, tt := range map[string]struct{
		events             []*model.APMEvent
		maxEPS             float64
		pastEPS            float64
		expectedSampledPct float64
		deltaPct           float64
} {
		"low EPS":      {testEvents, 100, 50, 1., 0},
		"limit EPS":    {testEvents, 100, 100, 1., 0},
		"overload EPS": {testEvents, 100, 150, 100. / 150., 0.05},
		// We should always keep events for sampled traces even if we are above maxEPS
		"overload EPS - sampled": {testEventsSampledTraces, 100, 500, 1., 0},
		// We should always keep events for sampled traces even if we are above maxEPS
		"nil events": {make([]*model.APMEvent, 5), 100, 0, 0, 0},
} {
    // t.Run ....
}
```
We don't really need the type right? Doing this and adding the helpers will make the test functions even slimmer.

event/sampler_max_eps_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	testEvents := make([]*model.APMEvent, 1000)
+	for i, _ := range testEvents {
+		testEvents[i] = &model.APMEvent{Span: testutil.RandomSpan()}
+	}
+
+	testEventsSampledTraces := make([]*model.APMEvent, 1000)
+	for i, _ := range testEventsSampledTraces {
+		testEventsSampledTraces[i] = &model.APMEvent{Span: testutil.RandomSpan(), TraceSampled: true}
```
gbbr: I'd really prefer it if we'd use the `SamplingDecision` type instead of boolean (`TraceSampled`). Otherwise what's the point of having it, right?

Although this probably means moving it into `model` to avoid a circular dependency problem 😊

AlexJF: Completely removed `SamplingDecision` as we no longer need the neutral state. Indeed, if we add it back, we should add it to all sampling logic.

AlexJF:
Superseded by https://github.com/DataDog/datadog-trace-agent/pull/513

PR #511, opened 2018-10-31
Tag APM events with associated rate metrics.

AlexJF:
This PR adds rate metrics to sampled APM events, one for each step at which an event could have been dropped:

* Trace (and thus event) could have been dropped at the client/tracer because we are not sampling at full rate there.
* Trace (and thus event) could have been dropped by our performance-guided presampler.
* Event could have been dropped at event extraction time.
* Event could have been dropped by our event samplers.


AlexJF:
Superseeded by https://github.com/DataDog/datadog-trace-agent/pull/513

PR #512, opened 2018-11-01
{config,flags}: change Windows paths from hard-coded to detecting program files

derekwbrown:
and programdata

In the process of converting the entire (windows) agent from hard-coding to proper handling.

Dependent upon https://github.com/DataDog/datadog-agent/pull/2546

### DIFF
--- a/config/config_windows.go
+++ b/config/config_windows.go
@@ -1,6 +1,13 @@
 package config
 
-const (
+import (
+	"path/filepath"
+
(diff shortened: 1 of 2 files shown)


gbbr:
@derekwbrown I’d like to prepare this for next week’s freeze. Any chance we could get rid of cgo as discussed? If not, let me know and I am ok with merging like this, but would highly prefer no cgo if possible.

PR #513, opened 2018-11-02
event: allow out-of-the-box APM event extraction.

AlexJF:
This PR adds functionality to enable out-of-the-box APM event extraction without requiring explicit agent configuration.

To that end, it adds:
* A metric based APM-event extraction that extracts APM events from spans containing an extraction rate metric.
* APM Event sampling logic, in particular, a sampler that limits `events/sec/agent`.
* Setting different sampling rates on sampled events to enable some stats computations on the backend.


event/extractor_metric_test.go
```diff
@@ -0,0 +1,71 @@
...
+		{"Priority 1 - Extraction rate = 1", createTestSpansWithEventRate(1), 1, 1},
+		// Priority 2 should have extraction rate of 1 so long as any extraction rate is set and > 0
+		{"Priority 2 - Missing extraction rate", createTestSpansWithEventRate(-1), 2, UnknownRate},
+		{"Priority 2 - Extraction rate = 0", createTestSpansWithEventRate(0), 2, 0},
+		{"Priority 2 - Extraction rate = 0.5", createTestSpansWithEventRate(0.5), 2, 1},
+		{"Priority 2 - Extraction rate = 1", createTestSpansWithEventRate(1), 2, 1},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
```
gbbr: This test looks very familiar 😉 Is there an easy way to remove the duplication between these two tests? If no, then leave it.

event/extractor_noop.go
```diff
@@ -10,6 +10,6 @@ func NewNoopExtractor() Extractor {
 	return &noopExtractor{}
 }
 
-func (s *noopExtractor) Extract(t model.ProcessedTrace) []*model.APMEvent {
-	return nil
+func (e *noopExtractor) Extract(s *model.WeightedSpan, priority int) (extract bool, rate float64) {
```
gbbr: You can also do:
```suggestion
func (*noopExtractor) Extract(_ *model.WeightedSpan, _ int) (extract bool, rate float64) {
```
Just FYI.

event/processor.go
```diff
@@ -0,0 +1,116 @@
...
+// NewProcessor returns a new instance of Processor configured with the provided extractors and samplers.
+//
+// Extractors will look at each span in the trace and decide whether it should be converted to an APM event or not. They
+// will be tried in the provided order, with the first non-neutral decision being the one applied.
+//
+// Samplers are applied, in order, to all extracted APM events, no matter which extractor extracted them. An event is
+// sampled if all samplers decided to sample it. As soon as a sampler decided to drop an event, it is dropped.
+// Samplers are assumed to be independent of each other so that the probability that an event survives sampling is
+// given by P(sampledA ^ sampledB ^ ... ^ sampledZ) = P(sampledA) * P(sampledB) * ... * P(sampledZ).
+func NewProcessor(extractors []Extractor, samplers []Sampler) *Processor {
```
gbbr: I really like this implementation 👍And really appreciate the awesome documentation too! Thank you!

event/processor.go
```diff
@@ -0,0 +1,116 @@
...
+	return &Processor{
+		extractors: extractors,
+		samplers:   samplers,
+	}
+}
+
+// Start starts all underlying samplers.
```
gbbr: I was told that good documentation usually doesn't talk about implementation details. And I like that. On that note:
```go
// Start starts the processor.
```
Also, is there ever a case when we want a new processor but not to start it? If not, maybe this can be removed and made part of `NewProcessor`

event/processor.go
```diff
@@ -0,0 +1,116 @@
...
+func (p *Processor) Stop() {
+	for _, sampler := range p.samplers {
+		sampler.Stop()
+	}
+}
+
+// Process takes a processed trace, extracts events from it, submits them to the samplers and returns a collection of
+// sampled events along with the total count of extracted events.
+func (p *Processor) Process(t model.ProcessedTrace) (events []*model.APMEvent, numExtracted int) {
+	// Short-circuit if there are no extractors configured
```
gbbr: This comment is also not needed IMHO.

event/processor.go
```diff
@@ -0,0 +1,116 @@
...
+
+	for _, span := range t.WeightedTrace {
+		var extractedEvent *model.APMEvent
+
+		// Loop through extractors until we find one that attempted to extract an event.
+		for _, extractor := range p.extractors {
+			extract, rate := extractor.Extract(span, priority)
+
+			// If the extractor didn't know what to do with this span, try the next one
+			if rate == UnknownRate {
```
gbbr: Same here, can we put the comment inside?
```go
if rate == RateNone {
    // try the next
```
Although this logic is a bit fuzzy here. Maybe we can be a bit more specific and phrase it differently than "didn't know what to do".

event/processor.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// If priority is not set on a trace, assume priority 0
+	if !hasPriority {
+		priority = 0
+	}
+
+	for _, span := range t.WeightedTrace {
+		var extractedEvent *model.APMEvent
+
+		// Loop through extractors until we find one that attempted to extract an event.
```
gbbr: The inline comment above `break` already clarifies this. I think we can take this one out.

event/processor.go
```diff
@@ -0,0 +1,116 @@
...
+			// next ones.
+			break
+		}
+
+		// If we didn't find any event in this span, try the next span
+		if extractedEvent == nil {
+			continue
+		}
+
+		numExtracted++
```
gbbr: `statsd.Count` could be called here.

event/processor.go
```diff
@@ -0,0 +1,116 @@
...
+				sampleDecision = false
+				break
+			}
+
+			// Total sample rate is the multiplication of each sampler's rate as we assume samplers are independent
+			sampleRate *= rate
+		}
+
+		// If the event survived global sampling then add it to the results.
+		if sampleDecision {
```
gbbr: Can we find a better name for this variable? Its name doesn't reflect a positive outlook, it feels a bit like it would better be suited as:
```go
if sampleDecision == true {
```
Which would be redundant. Maybe `sampleOk` or `sampleYes` would be better?

event/processor.go
```diff
@@ -0,0 +1,116 @@
...
+
+	priority, hasPriority := t.GetSamplingPriority()
+
+	// If priority is not set on a trace, assume priority 0
+	if !hasPriority {
+		priority = 0
+	}
+
+	for _, span := range t.WeightedTrace {
+		var extractedEvent *model.APMEvent
```
gbbr: Would it make sense to simply call this `event`?

event/processor_test.go
```diff
@@ -0,0 +1,102 @@
...
+		extractorRates       []float64
+		samplerRates         []float64
+		expectedExtractedPct float64
+		expectedSampledPct   float64
+		deltaPct             float64
+	}{
+		{"No extractors", nil, nil, 0, 0, 0},
+
+		// Test Extractors
+		{"Extractor(0) - No samplers", []float64{0.}, nil, 0, 0, 0},
```
gbbr: I'm curious (never seen this before): what this the outcome of `0.`? Is it to avoid writing `float64(0)`? In which case that's quite cool, but I don't think we need that here since it would be automatically implied?

event/sampler_max_eps_test.go
```diff
@@ -0,0 +1,96 @@
...
+	}{
+		"low EPS":      {testEvents, 100, 50, 1., 0},
+		"limit EPS":    {testEvents, 100, 100, 1., 0},
+		"overload EPS": {testEvents, 100, 150, 100. / 150., 0.05},
+		// We should always keep events for sampled traces even if we are above maxEPS
+		"overload EPS - sampled": {testEventsSampledTraces, 100, 500, 1., 0},
+	} {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+
```
gbbr: Nit but I think the number of empty lines is a bit excessive. This is quite uncommon in Go if you look at most projects and the standard library. Especially variable declarations are always grouped together, as well as if conditions referring to return arguments that are just above them, like:

```go
sample, rate := sampler.Sample(event)
if sample {
```
It would be nice if we could make most of the code in the repository could look more or less the same.

model/event.go
```diff
@@ -1,7 +1,75 @@
 package model
 
+const (
+	KeySamplingRateClientTrace     = "_dd.v1.rate.iusr"
```
gbbr: Can we document what these mean? It would be super helpful for the future.

model/event.go
```diff
@@ -1,7 +1,75 @@
...
+	return e.Span.GetMetricDefault(KeySamplingRatePreSampler, 1.0)
+}
+
+// SetPreSamplerSampleRate sets the rate at which the trace from which we extracted this event was sampled by the
+// agent's presampler.
+func (e *APMEvent) SetPreSamplerSampleRate(rate float64) {
+	if rate < 1 {
+		e.Span.SetMetric(KeySamplingRatePreSampler, rate)
+	} else {
+		delete(e.Span.Metrics, KeySamplingRatePreSampler)
```
gbbr: Why do we delete it when above 1?

model/span.go
```diff
@@ -48,6 +48,26 @@ func (s *Span) Weight() float64 {
 	return 1.0 / sampleRate
 }
 
+// GetMetric gets a value in the span Metrics map.
+func (s *Span) GetMetric(k string) (float64, bool) {
+	if s == nil || s.Metrics == nil {
+		return 0, false
+	}
+
+	val, ok := s.Metrics[k]
```
gbbr: `return s.Metrics[k]`

model/event.go
```diff
@@ -1,7 +1,75 @@
...
+		e.Span.SetMetric(KeySamplingRateEventExtraction, rate)
+	} else {
+		delete(e.Span.Metrics, KeySamplingRateEventExtraction)
+	}
+}
+
+// GetPreSamplerSampleRate gets the rate at which the trace from which we extracted this event was sampled by the
+// agent's presampler.
+// NOTE: This defaults to 1 if no rate is stored.
+func (e *APMEvent) GetPreSamplerSampleRate() float64 {
```
gbbr: Can we minimize the stutter a bit here and in all the following ones by doing s/SamplerSample/Sampler. I mean:
* GetPreSamplerSampleRate -> GetPreSamplerRate
* SetPreSamplerSampleRate -> SetPreSamplerRate
* etc...

model/span.go
AlexJF: This doesn't work for me for some reason. Gives me an error. Limitation on the compiler's part?

model/event.go
AlexJF: Above 1 and 1 are functionally equivalent. We delete it when it's set to 1 because we assume no rate set to be == 1. This is to save bandwidth when nothing gets sampled (which will hopefully be the case for most of our users).

event/constants.go
```diff
@@ -0,0 +1,4 @@
+package event
```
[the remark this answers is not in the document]
AlexJF: I can definitely put it in `extractor.go` but in theory it can also be used by Samplers, thus why I put it here.

cmd/trace-agent/agent.go
```diff
@@ -328,14 +337,20 @@ func traceContainsError(trace model.Trace) bool {
 	return false
 }
 
-func eventExtractorFromConf(conf *config.AgentConfig) event.Extractor {
+func eventProcessorFromConf(conf *config.AgentConfig) *event.Processor {
+	extractors := []event.Extractor{
+		event.NewMetricBasedExtractor(),
+	}
 	if len(conf.AnalyzedSpansByService) > 0 {
```
LotharSee: This introduces a behavior change where both the old and new config would be applied together.
I'd suggest to add the legacy extractor only if `len(conf.AnalyzedRateByServiceLegacy) == 0` to keep the previous behavior.

cmd/trace-agent/agent.go
```diff
@@ -260,8 +262,15 @@ func (a *Agent) Process(t model.Trace) {
...
-		// NOTE: Events can be extracted from non-sampled traces.
-		tracePkg.Events = a.EventExtractor.Extract(pt)
+		// NOTE: Events can be processed on non-sampled traces.
+		numExtracted := 0
+		tracePkg.Events = a.EventProcessor.Process(pt, event.ProcessorParams{
+			ClientSampleRate: clientSampleRate,
+			PreSampleRate:    preSamplerRate,
+			ExtractionCb:     func(e *model.APMEvent) { numExtracted++ },
+		})
+		statsd.Client.Count("datadog.trace_agent.events.extracted", int64(numExtracted), nil, 1)
```
LotharSee: 2 statsd requests per trace can get expensive for high traffic agents. Thus the usage of the `stat` accumulator. We probably want to add these variables there.

event/extractor_fixed_rate_test.go
```diff
@@ -1,62 +1,50 @@
...
-			test.trace.Sampled = rand.Int() > 0
-			analyzedSpans := ae.Extract(test.trace)
-
-			if test.expectedSampling {
-				assert.Len(analyzedSpans, 1, fmt.Sprintf("Trace %v should have been sampled", test.trace))
-			} else {
-				assert.Len(analyzedSpans, 0, fmt.Sprintf("Trace %v should not have been sampled", test.trace))
-			}
-		})
+		testExtractor(t, NewFixedRateExtractor(config), test)
```
gbbr: 👌

model/span.go
```diff
@@ -14,6 +14,22 @@ const (
...
 )
 
+// SamplingPriority is the type encoding a priority sampling decision.
+type SamplingPriority int8
+
+const (
+	// UnknownPriority is the value for SamplingPriority when no priority sampling decision could be found.
+	UnknownPriority SamplingPriority = -99
+	// UserDropPriority is the value set by a user to explicitly drop a trace.
+	UserDropPriority SamplingPriority = -1
```
gbbr: Can we do like everywhere else with the prefix? Either `Priority*` or `SamplingPriority*`. As you wish.

model/span.go
gbbr: Ah, yes. I see what you mean. Odd one. 😄

model/span.go
```diff
@@ -14,6 +14,22 @@ const (
 	SamplingPriorityKey = "_sampling_priority_v1"
 )
 
+// SamplingPriority is the type encoding a priority sampling decision.
+type SamplingPriority int8
+
+const (
+	// UnknownPriority is the value for SamplingPriority when no priority sampling decision could be found.
```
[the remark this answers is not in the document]
AlexJF: We used it in some tests so when I converted this to an actual type, I needed it defined. Also, with this available, I was also planning on using it for the processor, when a priority isn't set. In retrospect, having:

* -2 == User drop
* -1 == Auto drop
* 0 == Unset/Unknown
* 1 == Auto keep
* 2 == User keep

Would probably have been simpler?

cmd/trace-agent/agent.go
```diff
@@ -260,8 +262,15 @@ func (a *Agent) Process(t model.Trace) {
...
 		}
 
-		// NOTE: Events can be extracted from non-sampled traces.
-		tracePkg.Events = a.EventExtractor.Extract(pt)
+		// NOTE: Events can be processed on non-sampled traces.
+		numExtracted := 0
+		tracePkg.Events = a.EventProcessor.Process(pt, event.ProcessorParams{
+			ClientSampleRate: clientSampleRate,
+			PreSampleRate:    preSamplerRate,
+			ExtractionCb:     func(e *model.APMEvent) { numExtracted++ },
```
gbbr: Actually when I enumerated option b) as a callback I was hoping we could squeeze in the setting of rates there as well, but only later realized that the rates are only set on sampled events and here we are counting extracted events.

With the risk of you hating me: this doesn't look like a good reason to have a callback and I'd rather have your original solution where the count is returned. I guess we could consider it as some sort of stats for the processor which might've been your original intention too.

It just feels so odd to have this mapping function here for no reason other than counting.

Would you mind changing it back please? Sorry! 🙇  😊

model/span.go
AlexJF: Argh my brain works backwards :joy:

model/span.go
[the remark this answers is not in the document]
gbbr: I guess most obvious name would be `PriorityNone`, right? Because there is no priority set.

model/span.go
```diff
@@ -14,6 +15,22 @@ const (
...
+	PriorityNone SamplingPriority = math.MinInt8
+	// UserDropPriority is the value set by a user to explicitly drop a trace.
```
gbbr: Note that you forgot to update the identifier names in the comments.

model/span.go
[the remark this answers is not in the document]
LotharSee: It'd be really confusing to have numbers not aligning with what is actually transferred over the wire.
Is the usage of an extra bool to cover the case of unset (which is what we currently do?) causing troubles?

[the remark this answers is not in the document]
AlexJF: I'm not suggesting we change the scheme. It's too late now 😄. I'm also not hitting any particular issue and could definitely add the boolean as an extra parameter and change the (existing) test that was relying on this "disabled" value if you're set against having a `PriorityNone` encoding that information.

I personally, think the enum with this `PriorityNone` makes it simpler and clearer than a `(bool, int)` but don't feel strongly about it.

model/span.go
AlexJF: Hmm looks like IntelliJ is not doing the autorename for consts, only for methods. Probably a misconfig 😅 Fixed

config/agent.go
```diff
@@ -60,6 +60,7 @@ type AgentConfig struct {
 	// Sampler configuration
 	ExtraSampleRate float64
 	MaxTPS          float64
+	MaxEPS          float64
```
LotharSee: Suggestion: let's anticipate the need to configure it as an env variable and add it to `merge_env.go`.

model/span.go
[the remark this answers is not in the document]
gbbr: The truth is that we _have_ been using the `(SamplingPriority, bool)` format so far anyway, and it's the "Go way" of checking if something is set, so I tend to agree with Benjamin here. It also makes it much clearer, although I agree it's more tedious.

model/event.go
gbbr: Perhaps this is a good reason to put a small comment inside the `else` clause, since it's not so obvious why this happens.

model/span.go
```diff
@@ -14,6 +15,22 @@ const (
...
 )
 
+// SamplingPriority is the type encoding a priority sampling decision.
+type SamplingPriority int8
+
+const (
+	// PriorityNone is the value for SamplingPriority when no priority sampling decision could be found.
+	PriorityNone SamplingPriority = math.MinInt8
+	// PriorityUserDrop is the value set by a user to explicitly drop a trace.
+	PriorityUserDrop SamplingPriority = -1
```
gbbr: You don't have to write the type on each one. Just the first is enough in a group.

event/extractor.go
```diff
@@ -10,5 +10,5 @@ type Extractor interface {
 	// true if an extraction should happen or false otherwise. It also returns a rate specifying the extraction rate
 	// taken into account for this decision. If the returned rate is RateNone, then this extractor
 	// didn't find anything to extract.
-	Extract(span *model.WeightedSpan, priority model.SamplingPriority) (extract bool, rate float64)
+	Extract(span *model.WeightedSpan, hasPriority bool, priority model.SamplingPriority) (extract bool, rate float64)
```
gbbr: Oh damn. This is what you mean.

gbbr: @LotharSee do we really prefer this more than the enums?

model/event.go
```diff
@@ -1,7 +1,83 @@
 package model
 
+const (
```
LotharSee: We aren't set on these yet. See internal specification work.

event/extractor.go
gbbr: I'd say we go back to where we were 😊

model/event.go
AlexJF: Updated with the values mentioned there.

event/processor.go
```diff
@@ -0,0 +1,118 @@
+package event
+
+import "github.com/DataDog/datadog-trace-agent/model"
+
+// Processor is responsible for all the logic surrounding extraction and sampling of APM events from processed traces.
+type Processor struct {
+	extractors []Extractor
+	samplers   []Sampler
```
LotharSee: Kind of relate to Gabriel's comment in https://github.com/DataDog/datadog-trace-agent/pull/515

I'm not sure of the need / benefit of having generic `samplers` here. In practice we only have a maxEPS sampler and no other planned. It might be a too-early abstraction?

It could simplify the code remove the indirection.

model/event.go
```diff
@@ -1,7 +1,69 @@
...
+	// KeySamplingRateEventExtraction is the key of the metric storing the event extraction rate on an APM event.
+	KeySamplingRateEventExtraction = "_dd1.sr.eausr"
+	// KeySamplingRateEventSampler is the key of the metric storing the event sampler rate on an APM event.
```
LotharSee: Nit: call it the "max EPS sampler" instead?

event/extractor.go
```diff
@@ -4,8 +4,11 @@ import (
...
 )
 
-// Extractor extracts APM event spans from a trace.
+// Extractor is a special kind of sampler that decides whether an APM event should be extracted from a span.
 type Extractor interface {
-	// Extract extracts APM event spans from the given weighted trace information and returns them.
-	Extract(t model.ProcessedTrace) []*model.APMEvent
+	// Extract decides whether to extract an APM event from the provided span with the specified priority, returning
+	// true if an extraction should happen or false otherwise. It also returns a rate specifying the extraction rate
+	// taken into account for this decision. If the returned rate is RateNone, then this extractor
```
LotharSee: What justifies the distinction between a `RateNone` and a rate of 0?

AlexJF: The fact that the first matching extractor has priority. Picture this:

* Metric extraction for (servA, opA) explicitly set to 0.
* Agent configured extraction for (servA, opA) explicitly set to 1.

Without the `RateNone` vs 0 distinction we'd end up extracting events for all spans even though the user explicitly set the rate to 0.

LotharSee: Oh, that's to indicate to the chained extractors below if they have to run or not. It wasn't obvious at the first read.

It feels a bit of an odd indirect way to deal with the fact that we want the MetricExtractor to take precedence on the agent-config one.

Maybe `Extract` could return `(skip, extracted bool, rate float64)` instead?

But I guess that's a nitpick, it shouldn't slow this PR down.

event/sampler_max_eps.go
```diff
@@ -0,0 +1,105 @@
...
+	s.rateCounter.Stop()
+}
+
+// Sample determines whether or not we should sample the provided event in order to ensure no more than maxEPS events
+// are sampled every second.
+func (s *maxEPSSampler) Sample(event *model.APMEvent) (sampled bool, rate float64) {
+	// Count that we saw a new event
+	s.rateCounter.Count()
+
+	// Events with sampled traces are always kept even if that means going a bit above max eps.
```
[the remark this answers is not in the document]
AlexJF: Indeed I rechecked the [document](https://docs.google.com/document/d/134KDISeTiOUyFX9DYBhDBtUvNFlGnHBVGo0CwP6jMJY/edit#heading=h.og8qgstiqk88) and saw

```
That way we ensure that a trace sampled by the priority sampler will have a sampled event. (we purposely don’t do it for signature sampling because it would skew the stats in Trace Search).
```

I could swear at some point this document mentioned always keeping events for sampled traces. Indeed I can't find it right now. It could lead to strange UX where users can't find APM events associated with traces that they can find in the UI but, as you say, this should be a rare occurrence.

Going by that text though, it does seem like we don't want to completely remove this condition but rather change it to `Events associated with traces with priority 2 are always kept`, right?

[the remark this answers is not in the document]
LotharSee: Yes it might have been removed as we iterated :)


And very good point on `priority:2`!

Actually, for it to work without skewing the stats, we should either:

- skip totally the maxEPS for p2
- have a maxEPS instance only for p2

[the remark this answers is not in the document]
AlexJF: Went with MaxEPS ignoring p2 since that's what we do with MaxTPS as well.

event/processor.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	clientSampleRate := t.Root.GetClientSampleRate()
+	preSampleRate := t.Root.GetPreSampleRate()
+
+	for _, span := range t.WeightedTrace {
+		var event *model.APMEvent
+
+		for _, extractor := range p.extractors {
+			extract, rate, decided := extractor.Extract(span, priority)
```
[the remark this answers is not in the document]
AlexJF: I had something similar in one of the initial versions of this PR. Some issues that bug me:

* Maintainability issue: Event construction which was done in the `Processor` and is arguably "standard" among extractors would now be spread and repeated in each extractor forcing any changes to the base APMEvent structure to now touch several files.
* Performance issue (?): In a scenario where we are extracting at a rate of 1% from an input rate of 100K events/sec, we'd be creating 100K APMEvent objects per second from which we'd be keeping 1K objects after going through the extraction logic. I'm not sure how smart Go is with inlining returned structs but if the APMEvent objects are instantiated in the heap this would add extra tax to the garbage collector.
* Conceptual/Modelling nitpick: Extraction decision is now not the sole responsibility of `Extractor` but shared between `Extractor` and `Processor`. True enough, all our current Extractors use the same sampling logic for their final decision but whatever logic is used could be argued to be an Extractor implementation detail.

Now I don't feel super strongly about any of these (the first one is what bugs me the most) but I also don't think the current version is so confusing as to warrant yet another rewrite :)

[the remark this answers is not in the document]
gbbr: My main point here is that this new version seems much more succinct and easier to understand. I also feel that this couples the same type of behavior nicely together: sampling all in once place and extraction all in one place.

> Maintainability issue: Event construction which was done in the Processor and is arguably "standard" among extractors would now be spread and repeated in each extractor forcing any changes to the base APMEvent structure to now touch several files.

I don't think this is a real concern. This is true for any structure anywhere and should not be a reason to block us from having a nice clean architecture. We can reconsider this later if it happens. Also, an event extractor makes more sense when it returns an actual event.

> Extraction decision is now not the sole responsibility of Extractor but shared between Extractor and Processor. True enough, all our current Extractors use the same sampling logic for their final decision but whatever logic is used could be argued to be an Extractor implementation detail.

I find this is much better because extractors now extract an event from that settings map (or elsewhere) and return it along with the set rate, then, the processor samples it using two samplers. This seems less repetitive than having the sampling in each extractor.

> whatever logic is used could be argued to be an Extractor implementation detail.

It could, but is not the case now.

> Performance issue (?): In a scenario where we are extracting at a rate of 1% from an input rate of 100K events/sec, we'd be creating 100K APMEvent objects per second from which we'd be keeping 1K objects after going through the extraction logic. I'm not sure how smart Go is with inlining returned structs but if the APMEvent objects are instantiated in the heap this would add extra tax to the garbage collector.

This is a valid point. We are only copying a bunch of pointers and bool here. Nevertheless, I think performance discussions should not happen without having benchmarks to use as a basis.

> I also don't think the current version is so confusing as to warrant yet another rewrite

This wouldn't be a rewrite, just a rearrangement. I feel this would provide a great improvement on simplicity and readability. Unfortunately (mostly for me), it was a learning process to understand what's happening here. Now that I did, I feel that this would be a very nice, clean solution.

If you feel my requirement is too lengthy (time-wise) then leave it, but I think a solution like this would be simpler. The fact that you've thought about it too is also a good indication.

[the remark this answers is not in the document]
gbbr: To confirm: the current version is definitely not so confusing to warrant a rewrite. I was just trying to find the best, simplest solution. It's up to you to decide if you have the time to spend on it.

model/event.go
```diff
@@ -1,7 +1,69 @@
...
 
+const (
+	// KeySamplingRateEventExtraction is the key of the metric storing the event extraction rate on an APM event.
+	KeySamplingRateEventExtraction = "_dd1.sr.eausr"
+	// KeySamplingRateMaxEPSSampler is the key of the metric storing the max eps sampler rate on an APM event.
+	KeySamplingRateMaxEPSSampler = "_dd1.sr.eamax"
+)
+
 // APMEvent is an event extracted from received traces and sent to Datadog's Trace Search functionality.
 type APMEvent struct {
```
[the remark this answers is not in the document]
AlexJF: `ExtractedEvent` might be too specific (specially because we actually have Extracted vs Sampled events)? Just went with `Event`.

event/sampler.go
```diff
@@ -0,0 +1,15 @@
+package event
```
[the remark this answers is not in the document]
AlexJF: I removed the file but kept the interface (package private though) otherwise I can't easily check that the processor correctly interacts with `MaxEPSSampler` (working with a mock 2-levels deep is not fun).

event/extractor_metric.go
```diff
@@ -0,0 +1,34 @@
...
+func NewMetricBasedExtractor() Extractor {
+	return &metricBasedExtractor{}
+}
+
+// Extract decides whether to extract APM events from a span based on the value of the event extraction rate metric set
+// on that span. If priority is 2 (manually sampled) and extraction rate is bigger than 0, then an event is always
+// extracted.
+func (e *metricBasedExtractor) Extract(s *model.WeightedSpan, priority model.SamplingPriority) (extract bool, rate float64, decided bool) {
+	extractionRate, ok := s.GetEventExtractionRate()
+
```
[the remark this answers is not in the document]
AlexJF: Can we add an automatic linter? 😝

event/extractor_metric_test.go
```diff
@@ -0,0 +1,41 @@
...
+		if eventRate >= 0 {
+			spans[i].SetMetric(model.KeySamplingRateEventExtraction, eventRate)
+		}
+	}
+	return spans
+}
+
+func TestMetricBasedExtractor(t *testing.T) {
+	tests := []extractorTestCase{
+		{"No priority - Missing extraction rate", createTestSpansWithEventRate(-1), 0, -1},
```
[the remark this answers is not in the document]
AlexJF: I seem to disagree with a lot of Go's style decisions 😄. Using cryptic test names boggles the mind as it makes test code less human-readable.

Went with your second suggestion along with a comment indicating the structure of the naming to at least give some quick hints to people reading test code.

event/processor.go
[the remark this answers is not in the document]
AlexJF: Done with slight variations.

event/extractor.go
```diff
@@ -4,8 +4,9 @@ import (
...
 )
 
-// Extractor extracts APM event spans from a trace.
+// Extractor extracts APM events from matching spans.
 type Extractor interface {
-	// Extract extracts APM event spans from the given weighted trace information and returns them.
-	Extract(t model.ProcessedTrace) []*model.APMEvent
+	// Extract decides whether to extract an APM event from the provided span with the specified priority and returns
+	// the extracted event along with a suggested extraction sample rate (or nil if no event was extracted).
+	Extract(span *model.WeightedSpan, priority model.SamplingPriority) (*model.Event, float64)
```
gbbr: Let’s not make use of `nil` please. I don’t think that’s a good practice. What you did is fine but let’s add a third “ok” bool value.

PR #515, opened 2018-11-06
event: add max eps reporting.

AlexJF:
This PR adds periodic max eps sampling reporting. This reporting is 2-fold:
* Statsd gauges for current, max, sample_rate and max_reached.
* Log lines prompting the user to tune the `MaxEPS` setting if the maximum event rate was reached.

Sample output:

```
=== RUN   TestEventProcessorFromConf/metric_-_above_max_eps
1541500869865907000 [Warn] Max events per second reached (current=193.66/s, max=100.00/s). Some events are now being dropped (sample rate=0.52). Consider increasing MaxEPS setting.
1541500899864986000 [Warn] Max events per second reached (current=224.48/s, max=100.00/s). Some events are now being dropped (sample rate=0.45). Consider increasing MaxEPS setting.
```

### DIFF
--- a/event/sampler_max_eps.go
+++ b/event/sampler_max_eps.go
@@ -3,9 +3,14 @@ package event
 import (
 	"time"
 
+	log "github.com/cihub/seelog"
+
 	"github.com/DataDog/datadog-trace-agent/model"
 	"github.com/DataDog/datadog-trace-agent/sampler"
+	"github.com/DataDog/datadog-trace-agent/statsd"
 )
+
+const maxEPSReportFrequency = 10 * time.Second
 
 // maxEPSSampler (Max Events Per Second Sampler) is an event maxEPSSampler that samples provided events so as to try to ensure
 // no more than a certain amount of events is sampled per second.
@@ -15,6 +20,9 @@ import (
 type maxEPSSampler struct {
 	maxEPS      float64
 	rateCounter rateCounter
+
+	reportFrequency time.Duration
+	reportDone      chan bool
 }
 
 // NewMaxEPSSampler creates a new instance of a maxEPSSampler with the provided maximum amount of events per second.
@@ -22,16 +30,36 @@ func newMaxEPSSampler(maxEPS float64) *maxEPSSampler {
 	return &maxEPSSampler{
 		maxEPS:      maxEPS,
 		rateCounter: newSamplerBackendRateCounter(),
+
+		reportDone: make(chan bool),
 	}
 }
 
 // Start starts the underlying rate counter.
 func (s *maxEPSSampler) Start() {
 	s.rateCounter.Start()
+
+	go func() {
+		ticker := time.NewTicker(maxEPSReportFrequency)
+		defer close(s.reportDone)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-s.reportDone:
+				return
+			case <-ticker.C:
+				s.report()
+			}
+		}
+	}()
 }
 
 // Stop stops the underlying rate counter.
 func (s *maxEPSSampler) Stop() {
+	s.reportDone <- true
+	<-s.reportDone
+
 	s.rateCounter.Stop()
 }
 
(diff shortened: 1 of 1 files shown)


gbbr:
Edit: the for loop for the ticker in my example needs to be extended a bit, more along the lines of https://golang.org/pkg/time/#NewTicker to exit cleanly.

event/sampler_max_eps.go
```diff
@@ -3,42 +3,70 @@ package event
...
+
+	reportFrequency time.Duration
+	reportDone      chan bool
 }
 
-// NewMaxEPSSampler creates a new instance of a maxEPSSampler with the provided maximum amount of events per second.
-func NewMaxEPSSampler(maxEPS float64) Sampler {
-	return newMaxEPSSampler(maxEPS, newSamplerBackendRateCounter())
+// NewMaxEPSSampler creates a new instance of a MaxEPSSampler with the provided maximum amount of events per second.
+func NewMaxEPSSampler(maxEPS float64, reportFrequency time.Duration) *MaxEPSSampler {
```
gbbr: I think we can remove the `reportFrequency` from the arguments list, since it's not likely to be needed there, right? Then, for testing, you can keep it in the unexported constructor. It could be just a const.

event/sampler_max_eps.go
```diff
@@ -47,19 +75,51 @@ func (s *maxEPSSampler) Sample(event *model.APMEvent) (sampled bool, rate float6
 		return true, 1
 	}
```
gbbr: With the risk of becoming annoying here: can we remove all the whitespace?

event/sampler_max_eps.go
```diff
@@ -3,42 +3,70 @@ package event
...
+	}()
 }
 
 // Stop stops the underlying rate counter.
-func (s *maxEPSSampler) Stop() {
+func (s *MaxEPSSampler) Stop() {
+	s.reportDone <- true
+	<-s.reportDone
+
 	s.rateCounter.Stop()
```
gbbr: Shouldn't this happen before sending down the `reportDone` channel?

event/sampler_max_eps.go
```diff
@@ -3,42 +3,70 @@ package event
...
 }
 
 // Sample determines whether or not we should sample the provided event in order to ensure no more than maxEPS events
 // are sampled every second.
-func (s *maxEPSSampler) Sample(event *model.APMEvent) (sampled bool, rate float64) {
+// This method is thread-safe.
+func (s *MaxEPSSampler) Sample(event *model.APMEvent) (sampled bool, rate float64) {
```
gbbr: All are these changes of exporting "MaxEPSSampler" still needed, now that we've reverted your initial implementation?

event/sampler_max_eps.go
AlexJF: Doesn't really make a significant difference either way. But I like to stop things in the inverse order that I've started them in (kinda like what defer does).

event/sampler_max_eps.go
AlexJF: Indeed, forgot to revert these

event/sampler_max_eps.go
gbbr: > Doesn't really make a significant difference

You're right. The send is non-blocking, so I guess it should be fine. I was thinking that it might make more sense to stop sending ticks before exiting the routine, but in this case I guess we're good.

event/sampler_max_eps.go
```diff
@@ -3,15 +3,23 @@ package event
...
 	"time"
 
+	log "github.com/cihub/seelog"
+
 	"github.com/DataDog/datadog-trace-agent/model"
 	"github.com/DataDog/datadog-trace-agent/sampler"
+	"github.com/DataDog/datadog-trace-agent/statsd"
 )
 
+const maxEPSReportFrequency = 30 * time.Second
```
LotharSee: Let's move to 10 seconds, that's the same interval as dogstats. It'd allow best resolution on the resulting timeseries.

event/sampler_max_eps.go
```diff
@@ -23,47 +31,92 @@ func newMaxEPSSampler(maxEPS float64, rateCounter rateCounter) Sampler {
...
 
-	return
+	sampleRate := s.getSampleRate()
+	statsd.Client.Gauge("datadog.trace_agent.events.max_eps.sample_rate", sampleRate, nil, 1)
+
+	reachedMaxGaugeV := 0.
+	if sampleRate < 1 {
+		reachedMaxGaugeV = 1.
+		log.Warnf("Max events per second reached (current=%.2f/s, max=%.2f/s). "+
+			"Some events are now being dropped (sample rate=%.2f). Consider increasing MaxEPS setting.",
```
LotharSee: Let's suggest instead to adjust sample rates so that one prefer to keep less instead of bumping the max.

event/sampler_max_eps.go
```diff
@@ -23,47 +31,92 @@ func newMaxEPSSampler(maxEPS float64, rateCounter rateCounter) Sampler {
...
 func (s *maxEPSSampler) Stop() {
+	s.reportDone <- true
+	<-s.reportDone
+
 	s.rateCounter.Stop()
 }
 
 // Sample determines whether or not we should sample the provided event in order to ensure no more than maxEPS events
 // are sampled every second.
+// This method is thread-safe.
```
gbbr: I'd expect by default stuff to be thread safe. I generally find that it's good to write only if it's _not_. Is this needed?

event/sampler_max_eps.go
```diff
@@ -23,47 +31,92 @@ func newMaxEPSSampler(maxEPS float64, rateCounter rateCounter) Sampler {
...
+		reachedMaxGaugeV = 1.
+		log.Warnf("Max events per second reached (current=%.2f/s, max=%.2f/s). "+
+			"Some events are now being dropped (sample rate=%.2f). Consider adjusting event sampling rates.",
+			currentRate, maxRate, sampleRate)
+	}
+	statsd.Client.Gauge("datadog.trace_agent.events.max_eps.reached_max", reachedMaxGaugeV, nil, 1)
 }
 
-// rateCounter keeps track of different event rates.
+// rateCounter keeps track of different event rates. Implementations of rateCounter must be thread-safe.
```
gbbr: I'm not sure if this comment adds anything. It's just an interface, the documentation should talk about any implementation, even if your intention here is just to "rename" the methods of Backend.

PR #521, opened 2018-11-07
watchdog: remove high-memory consumption test

gbbr:
This change removes the 1GB memory test in order to allow our tests to
run in the "medium" resource_class of CircleCI.

### DIFF
--- a/watchdog/info_test.go
+++ b/watchdog/info_test.go
@@ -133,7 +133,6 @@ func TestMemHigh(t *testing.T) {
 		return
 	}
 	doTestMemHigh(t, 1e7)
-	doTestMemHigh(t, 1e9)
 }
 
 type testNetHandler struct {


gbbr:
There is currently a bug (?) in CircleCI where our resource class is no longer considered:
screen shot 2018-11-07 at 12 29 35 pm
Exact same build a bit earlier:
screen shot 2018-11-07 at 12 29 28 pm

Let's make sure we can run with the defaults.

gbbr:
@ufoot I assume our tests should still be valid enough even with this third one removed?

watchdog/info_test.go
```diff
@@ -133,7 +133,6 @@ func TestMemHigh(t *testing.T) {
 		return
 	}
 	doTestMemHigh(t, 1e7)
-	doTestMemHigh(t, 1e9)
```
ufoot: Indeed, if `1e7` works, this 1Gb test is not strictly mandatory.

PR #522, opened 2018-11-07
cmd/trace-agent: reduce histogram calls in concentrator

gbbr:
Attemtp to reduce UDP hits and flooding.

We had reports of excessive UDP packets (>3000 per 10 seconds) for metric `datadog.trace_agent.err_distribution.len`, all with values of zero. Seems uncalled for. This change should hopefully alleviate.


gbbr:
I reduced the rate to 0.2, but not sure if this will alleviate the problem enough.

gbbr:
Great! That's probably the best thing to do. Note that I've added another commit with a small improvement. PTAL.

PR #526, opened 2018-11-14
config: use shared config utility from datadog-agent

gbbr:
Changes in this PR:
* Bump `datadog-agent` dependency to `6.6.0` (first commit)
* Use shared `pkg/config` package to load configuration (second commit)
* Add tests for _all_ yaml config properties.

Current repository size: 203 MB. Note that this will be dissipated once the agent merges into the main agent.

⚠️ Please review using the second commit.


config/merge_yaml.go
```diff
@@ -221,118 +154,131 @@ func (c *AgentConfig) loadYamlConfig(yc *YamlAgentConfig) {
...
+			if err != nil {
+				osutil.Exitf("replace_tags: %s", err)
+			}
+			c.ReplaceTags = rt
 		}
-		c.ReplaceTags = rt
 	}
 
-	if yc.TraceAgent.APMNonLocalTraffic != nil && *yc.TraceAgent.APMNonLocalTraffic {
+	if config.Datadog.GetBool("apm_config.apm_non_local_traffic") {
```
AlexJF: If `apm_non_local_traffic` is set to `true` by default, users will be unable to disable this behaviour.

config/merge_yaml.go
```diff
@@ -212,8 +212,10 @@ func (c *AgentConfig) loadYamlConfig(path string) error {
 		}
 	}
 
-	if config.Datadog.GetBool("apm_config.apm_non_local_traffic") {
-		c.ReceiverHost = "0.0.0.0"
+	if config.Datadog.IsSet("apm_config.apm_non_local_traffic") {
+		if config.Datadog.GetBool("apm_config.apm_non_local_traffic") {
```
AlexJF: This still doesn't solve my main problem which is not handling the `== false` case to set it back to `127.0.0.1`. This is mostly future-proofing if at some point we set `0.0.0.0` as the default and I just realized the old version also didn't handle it correctly so it's not really this PR's fault 😅

gbbr: We can get back to this if we ever change the default. This is just an initial piece of work anyway.

PR #543, opened 2018-11-30
internal/agent: copied documentation from wiki

gbbr:


gbbr:
@LotharSee please take a quick look just in case any of it is outdated.

PR #544, opened 2018-11-30
test: add integration package

gbbr:
Adds a package which allows running a fake backend and an agent with a given configuration. The runner exposes methods to send payloads to the agent and a channel to receive anything the agent outputs to the backend.

An example of how tests would be written, along with a first integration test can be seen in `test/suite/hostname_test.go`


test/agent.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+	}
+	proc.Wait()
+}
+
+func (s *Runner) runAgentConf(path string) <-chan error {
+	s.StopAgent()
+	cmd := exec.Command("trace-agent", "-config", path)
+	s.log.Reset()
+	cmd.Stdout = s.log
```
p-lambert: Nice! We can even inspect logs <3

test/server.go
```diff
@@ -0,0 +1,203 @@
...
+func (s *Runner) Shutdown(wait time.Duration) error {
+	defer close(s.out)
+	defer atomic.StoreUint64(&s.started, 0)
+
+	s.StopAgent()
+	ctx, _ := context.WithTimeout(context.Background(), wait)
+	return s.srv.Shutdown(ctx)
+}
+
+func (s *Runner) startBackend() error {
```
p-lambert: Nit: I would probably have a `Backend` abstraction encapsulating this and leave `Runner` just orchestrating (start/stop) both the agent and the `Backend`.

test/buffer.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (sb *safeBuffer) Write(p []byte) (int, error) {
+	sb.mu.Lock()
+	defer sb.mu.Unlock()
+	if len(p) >= len(sb.b) {
+		// p is bigger than the whole buffer; we store only
+		// the last len(sb.b) bytes.
+		n := copy(sb.b, p[len(p)-len(sb.b):])
+		sb.off = n
+		return n, nil
```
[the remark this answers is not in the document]
gbbr: You are absolutely correct. Thanks for this. I have both missed this one and also haven't fully respected the specification for [io.Writer](https://golang.org/pkg/io/#Writer). I have corrected it in the next commit.

test/server.go
gbbr: These aspects we call "nits" are actually the ones I am most interested in when getting reviews. Structuring the code in the simplest, most decoupled and easy to understand (and test) way is what is most important to me. Your suggestion has also crossed my mind. I will explore this. It would even make sense to have the logic that starts/stops the agent as a separate unit.

gbbr:
@p-lambert I've separated out both the backend and the agent runner into separate more isolated components. I'd be curious to hear what you think if you can spare the time to take another look.

internal/test/agent.go
```diff
@@ -0,0 +1,182 @@
...
+	s.port = 8126
+	if v.IsSet("apm_config.receiver_port") {
+		s.port = v.GetInt("apm_config.receiver_port")
+	}
+	v.Set("apm_config.apm_dd_url", "http://"+s.ddAddr)
+	if !v.IsSet("api_key") {
+		v.Set("api_key", "testing123")
+	}
+	if !v.IsSet("apm_config.trace_writer.flush_period_seconds") {
+		v.Set("apm_config.trace_writer.flush_period_seconds", 1)
```
LotharSee: You can make it even faster (like 100ms), same with `waitForTrace` ; as the more you will write tests, the more you will waste time on these internal waits.

internal/test/testsuite/hostname_test.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/DataDog/datadog-trace-agent/internal/agent"
+	"github.com/DataDog/datadog-trace-agent/internal/test"
+	"github.com/DataDog/datadog-trace-agent/internal/test/testutil"
+)
+
+func TestHostname(t *testing.T) {
```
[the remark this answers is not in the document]
LotharSee: Okay I didn't know if that was for example or actual testing. It is likely still worth having one (but it doesn't have to be exhaustive) test of env variables in here (in case it gets lost in a layer you wouldn't catch without integration testing).

And it can clearly wait for later PRs.

internal/test/agent.go
gbbr: That's what I've tried the first time too. But currently it's an `int` value which is supposed to be in seconds. A float won't work right now. I guess we can try and modify the `config` to allow floating point numbers here for the sake of testing. Indeed waiting 1 second for each test could end up being a waste of time. I will investigate this.

You might have miss-read `waitForTrace`. It doesn't wait for 3 seconds. It waits until the first trace comes but has a timeout of 3 seconds (fails after that time).

LotharSee: Right, making it allow floats can wait another PR.
Concerning `waitForTrace`, it just means that in practice it will see/receive traces faster is the TraceWriter flushes faster. Also allowing it a shorter timeout.

p-lambert:
> @p-lambert I've separated out both the backend and the agent runner into separate more isolated components. I'd be curious to hear what you think if you can spare the time to take another look.

That's exactly what I had in mind. I think now the abstractions have each a well defined set of responsibilities and the code is easier to follow (and hopefully to maintain!). Great job!

internal/test/testsuite/hostname_test.go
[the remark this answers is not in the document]
gbbr: Thanks Benjamin. I will try to add it.

internal/test/agent.go
gbbr: I've added it here, since it was a very simple change.

internal/test/testsuite/hostname_test.go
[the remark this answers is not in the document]
gbbr: Did this too.

Issue #502, opened 2018-10-23
proposal: allow setting APM env. via environment variable


dalen:
I'm trying to get APM working on ECS fargate, but I can't get the traces to get tagged with the environment properly. Adding host tags on Fargate doesn't seem to work either.

Is there any way to set the `apm_config.env` key using an environment variable?

gbbr:
Do you not have access or the possibility to change the config? You can also set the `env` tag on traces for the same effect. Most tracers have global tags which you can initialize them with. For Go it would be:
```go
tracer.Start(tracer.WithGlobalTag(ext.Environment, "name"))
```

dalen:
I run it as a container in the ECS task, so to change the config file I would need a custom container image for each environment instead of just being able to use the `datadog/agent:latest` image. That would be doable, but a bit more cumbersome than just setting an ENV variable depening on environment.

The project I'm setting up at the moment is a rails project, but I can check if I can set the environment in the tracer config there in the meantime.

But anyway, adding `DD_APM_ENV` to the list at https://github.com/DataDog/datadog-trace-agent/blob/master/config/merge_env.go#L13-L29 would be really helpful.

gbbr:
> The project I'm setting up at the moment is a rails project, but I can check if I can set the environment in the tracer config there in the meantime.

Should be easy based on the Ruby tracer documentation: https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#environment-and-tags

> But anyway, adding DD_APM_ENV to the list at https://github.com/DataDog/datadog-trace-agent/blob/master/config/merge_env.go#L13-L29 would be really helpful.

I agree with you. This isn't the only case where an environment variable is missing as 1:1 mapping on the config. We'll have to consider better solutions to this.

For the environment variable, you can now do it in code; but if there's every anything else you can always just create a new Docker image based on the Datadog one and copy your own config into it.

Please let me know if this solves your problem, in which case we can close the issue.

dalen:
Yeah, doing it in code solves the issue for me for now, thanks for the pointers.

You can close it or keep it as a feature request to actually add that environment variable :)

gbbr:
Great! Ok, let's keep it as a proposal. Slightly edited the title.

gbbr:
I think it might be best after all to close this now since the problem has a solution. 

In a next release (soon) most config options will be settable via environment variables too.

PR #548, opened 2018-12-17
internal/config: allow setting environment using environment variable

dalen:
This fixes https://github.com/DataDog/datadog-trace-agent/issues/502


gbbr:
I thought we agreed within #502 that the problem was resolved and this isn't needed.

PR #552, opened 2019-01-03
internal/config: add DD_APM_ prefixed env. vars duplicates

gbbr:
Adds new environment variables (correctly prefixed), marking others as deprecated:
* DD_APM_CONNECTION_LIMIT
* DD_APM_RECEIVER_PORT
* DD_APM_MAX_EPS
* DD_APM_MAX_TPS
* DD_APM_IGNORE_RESOURCES

### DIFF
--- a/internal/config/env.go
+++ b/internal/config/env.go
@@ -13,38 +13,38 @@ import (
 func applyEnv() {
 	// Warning: do not use BindEnv to bind config variables. They will be overriden
 	// when using the legacy config loader.
-	for envKey, cfgKey := range map[string]string{
+	for _, override := range []struct{ env, key string }{
 		// Core agent:
-		"DD_SITE":           "site",
-		"DD_API_KEY":        "api_key",
-		"DD_HOSTNAME":       "hostname",
-		"DD_BIND_HOST":      "bind_host",
-		"DD_DOGSTATSD_PORT": "dogstatsd_port",
-		"DD_LOG_LEVEL":      "log_level",
-		"HTTPS_PROXY":       "proxy.https", // deprecated
-		"DD_PROXY_HTTPS":    "proxy.https",
+		{"DD_SITE", "site"},
+		{"DD_API_KEY", "api_key"},
+		{"DD_HOSTNAME", "hostname"},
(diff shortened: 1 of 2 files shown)


internal/config/env.go
```diff
@@ -14,32 +14,44 @@ func applyEnv() {
...
 	// when using the legacy config loader.
 	for envKey, cfgKey := range map[string]string{
-		"DD_SITE":                  "site",
-		"DD_API_KEY":               "api_key",
-		"DD_HOSTNAME":              "hostname",
-		"DD_BIND_HOST":             "bind_host",
-		"DD_DOGSTATSD_PORT":        "dogstatsd_port",
-		"DD_LOG_LEVEL":             "log_level",
-		"DD_CONNECTION_LIMIT":      "apm_config.connection_limit",
+		// Core agent:
```
erichgess: It's possible that a customer will have both the deprecated and the current env var name defined, so we should have an explicit order of precedence for deprecated env var names and current env var names.  So if both the deprecated and the current name for a parameter is defined we always use the current one.  

However, ordering on a map is not guaranteed so this loop may prioritize deprecated names over current names.  

Can we change this to support ensuring that if both versions of a name are defined we'll always get the current one?

gbbr: Nice catch! You’re totally right about map ordering. Will change this.

erichgess: No problem.  Glad my Golang skills are getting good enough to be helpful!

gbbr:
@erichgess I've added specific tests for this and replaced the map with an ordered sequence (a slice). PTAL

internal/config/env.go
gbbr: Done!

internal/config/env_test.go
```diff
@@ -14,6 +15,35 @@ func TestMain(m *testing.M) {
 }
 
 func TestLoadEnv(t *testing.T) {
+	t.Run("overrides", func(t *testing.T) {
+		// tests that newer envs. override deprecated ones
```
erichgess: 👍

PR #553, opened 2019-01-03
internal/agent: slightly improve Normalize output

gbbr:
Slightly improves error messages, making them more consistent and in some cases more explicit.

### DIFF
--- a/internal/agent/normalizer.go
+++ b/internal/agent/normalizer.go
@@ -33,35 +33,35 @@ var (
 func (s *Span) Normalize() error {
 	// Service
 	if s.Service == "" {
-		return errors.New("span.normalize: empty `Service`")
+		return errors.New("empty `Service`")
 	}
 	if len(s.Service) > MaxServiceLen {
-		return fmt.Errorf("span.normalize: `Service` too long (max %d chars): %s", MaxServiceLen, s.Service)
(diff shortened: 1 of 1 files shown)


internal/agent/normalizer.go
```diff
@@ -78,24 +78,24 @@ func (s *Span) Normalize() error {
...
 	// (or it is "le bug de l'an 2000")
 	if s.Start < Year2000NanosecTS {
-		return fmt.Errorf("span.normalize: invalid `Start` (must be nanosecond epoch): %d", s.Start)
+		return fmt.Errorf("invalid `Start` (must be nanosecond epoch): %d", s.Start)
 	}
 
 	// If the end date is too far away in the future, it's probably a mistake.
 	if s.Start+s.Duration > time.Now().Add(MaxEndDateOffset).UnixNano() {
-		return fmt.Errorf("span.normalize: more than %v in the future", MaxEndDateOffset)
+		return fmt.Errorf("invalid `Start`+`Duration`: too far in the future")
```
erichgess: Could we include the value of the invalid end date in the message?  My experience is that makes debugging much easier, and I think that's the error message pattern you're using elsewhere :)

gbbr: Sorry, too late now, but will do as a follow up.

PR #554, opened 2019-01-04
internal/api: move "trace dropped" log to level DEBUG

gbbr:
This change moves the log entry that notifies the user that his trace
was dropped (due to the agent being too slow) to Debug, as it has turned
out to become quite spammy in high traffic environments. It is also not
of much value to the user. Dropping is already counted as a metric and
will be visible in the app.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -248,18 +248,15 @@ func (r *HTTPReceiver) handleTraces(v Version, w http.ResponseWriter, req *http.
 			}
 			log.Errorf(errorMsg)
 		} else {
-			atomic.AddInt64(&ts.SpansDropped, int64(spans-len(normTrace)))
-
 			select {
 			case r.Out <- normTrace:
-				// if our downstream consumer is slow, we drop the trace on the floor
-				// this is a safety net against us using too much memory
-				// when clients flood us
+				// ok
 			default:
+				// agent is too slow; we have to drop the trace
 				atomic.AddInt64(&ts.TracesDropped, 1)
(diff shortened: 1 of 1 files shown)


internal/api/api.go
```diff
@@ -248,18 +248,15 @@ func (r *HTTPReceiver) handleTraces(v Version, w http.ResponseWriter, req *http.
 			}
 			log.Errorf(errorMsg)
 		} else {
-			atomic.AddInt64(&ts.SpansDropped, int64(spans-len(normTrace)))
```
gbbr: We can see in [the code for `NormalizeTrace`](https://github.com/DataDog/datadog-trace-agent/blob/1723cd82c9f94ea7af2d0db454672d6e192dce4b/internal/agent/normalizer.go#L136-L164) that spans are never removed from the trace during the normalization process, so this addition serves no purpose.

gbbr:
@LotharSee an alternative solution is to enable log throttling. Should we recommend that in this situation? A part of me feels this should stay an error.

LotharSee:
IMHO I'd suggest:

- keep one DEBUG log line and add the traceID to it.
- a periodic reporting of trace dropped is enough. We might want to review what is in place though as:
  - we might not report a WARNING when that value is > 0
  - This `TracesDropped` counter seems to account for various things, we might want a breakdown by "drop reason".

gbbr:
Thanks @LotharSee. I think adding a tag is a great idea. Closing in favor of https://github.com/DataDog/datadog-trace-agent/pull/561.

PR #558, opened 2019-01-09
internal/agent: remove event structure and methods

gbbr:
This change removes the `agent.Event` structure which was not needed.


AlexJF (approved):
Ahah we came full circle with this. LGTM

gbbr:
Thanks @AlexJF. Indeed.

